PR #3, opened 2020-07-16
Add (better) support for French users

geoffreydhuyvetters:
### Added

- Added and configured the `nuxt-i18n` module
- Added missing translations

### Changed

- Changed the iframe URL to the legacy app to include a `lang` parameter (see https://github.com/cultuurnet/udb3-angular-app/pull/150)
- Translated the `x geleden` time info in the notifications
- Moved the prettier config to `package.json` (which has precedence according to https://prettier.io/docs/en/configuration.html)
- Renamed camelCased translation keys

### Removed

- Removed `vue-i18n` package
- Removed unused translation keys

---
Ticket: https://jira.uitdatabank.be/browse/III-3357
Note: This PR requires a review & merge of https://github.com/cultuurnet/udb3-angular-app/pull/150


pages/login.vue
```diff
@@ -129,7 +129,7 @@
 
         <h1 class="brand">UiTdatabank</h1>
         <p class="lead text-center">
-          <span>{{ $t('main.lead') }}</span>
+          <span>Organiseer je een activiteit?</span>
```
bertramakers, line 132:
```suggestion
          <span>{{ $i18n.t('main.lead') }}</span>
```

pages/login.vue
```diff
@@ -236,20 +236,11 @@
...
-              src="~/assets/vlaanderen-logo.png"
-              class="logo-flanders-full"
-            />
-            <img
-              v-show="$i18n.locale === 'fr'"
-              src="~/assets/flanders-logo.png"
-              class="logo-flanders-full"
-            />
+            <img v-show="isNL" src="~/assets/vlaanderen-logo.png" class="logo-flanders-full" />
+            <img v-show="isFR" src="~/assets/flanders-logo.png" class="logo-flanders-full" />
```
bertramakers, line 240: Would it make sense to also add a key for the image filename to the translation files, and do something like this?
```
<img src="~/assets/{{ $i18n.t('flanders-logo-image') }}" class="logo-flanders-full" />
```
Or is that not possible/frowned upon?

bertramakers: We probably also need a translatable `alt` for this in any case 🤔

geoffreydhuyvetters:
Feel free to merge this one when you're ok with my replies actions to the comments @bertramakers

bertramakers:
I'm okay with your comments @geoffreydhuyvetters, but I'm going to add the missing the translations in this PR before merging it instead of creating a new one.

bertramakers:
@simonpubliq @brampauwelyn this is finally ready for review (again)!
Should cover everything concerning translations.

PR #6, opened 2020-07-20
Redirect to login on invalid token

geoffreydhuyvetters:
### Added

- redirect to login on invalid token
- created a fetch wrapper that triggers logout when the fetch failed (40X)


api/api.js
```diff
@@ -1,14 +1,43 @@
...
+    return false;
+  }
+
+  const now = Math.round(Date.now() / 1000);
+  return decodedToken.exp - now > 0;
+};
+
+export const fetchWithLogoutWhenFailed = async (...args) => {
+  const response = await fetch(...args);
+  if (!response.ok) {
```
bertramakers, line 24: I'm going to change this to check if the status is `401` or `403`. That's also how it was in the AngularJS app https://github.com/cultuurnet/udb3-angular-app/blob/9ed7733986bf7ccedc79e4d712a42e83998df700/app/scripts/services/udbhttpinterceptor.js#L33-L41
Reason being that now you would also get logged out for e.g. a `500 Internal Server Error` or `404 Not Found`

bertramakers:
Looks good @geoffreydhuyvetters, I just did one small change. I'll merge this to master once #3 is merged. 👌

PR #7, opened 2020-07-23
Feature/iii 3329

simon-debruijn:
### Added
Added giftbox feature

Schermafbeelding 2020-07-23 om 09 15 52
Schermafbeelding 2020-07-23 om 09 16 08
Schermafbeelding 2020-07-23 om 09 16 26
Schermafbeelding 2020-07-23 om 09 16 45

Ticket: [https://jira.uitdatabank.be/browse/III-3329](https://jira.uitdatabank.be/browse/III-3329)


components/giftbox.vue
```diff
@@ -0,0 +1,280 @@
...
+              :src="selectedFeature.image"
+              :alt="selectedFeature.title"
+              class="feature-image"
+            />
+          </a>
+          <!-- eslint-disable-next-line vue/no-v-html -->
+          <div class="feature-content" v-html="selectedFeature.body" />
+        </div>
+        <div v-else class="features-detail">
+          <p>Geen nieuwe features te melden.</p>
```
brampauwelyn, line 64: Use i18n 🌍  for translations. UDB is availabale in both french and dutch

components/giftbox.vue
```diff
@@ -0,0 +1,280 @@
...
+          this.selectedFeature = this.features[0];
+          this.seenFeatures.add(this.selectedFeature.uid);
+        }
+      });  
+    },
+    methods: {
+      async fetchFeatures() {
+        this.loading = true;
+        const response = await fetch(
+          'https://www.publiq.be/uitdatabank/articles.json',
```
brampauwelyn, line 93: Load this url from .env file

components/giftbox.vue
```diff
@@ -0,0 +1,280 @@
...
+    right: 1.5rem;
+    bottom: 1.5rem;
+    text-align: center;
+
+    .giftbox-button {
+      width: 100%;
+      height: 100%;
+      line-height: 100%;
+      border-radius: 50%;
+      background-color: #c0120c;
```
brampauwelyn, line 134: Maybe use a .scss file (_variables.scss) to declare the" brand colors" so you can reuse it as a variable in the different components.
Something like `udb-primary-color: #c0120c`

components/giftbox.vue
```diff
@@ -0,0 +1,280 @@
...
+      text-align: center;
+      -webkit-box-shadow: 0px 0px 35px -8px rgba(0, 0, 0, 0.75);
+      -moz-box-shadow: 0px 0px 35px -8px rgba(0, 0, 0, 0.75);
+      box-shadow: 0px 0px 35px -8px rgba(0, 0, 0, 0.75);
+      transition: background-color 500ms, transform 500ms;
+
+      &:active {
+        background-color: lighten($color: #c0120c, $amount: 40%);
```
brampauwelyn, line 142: Use variables for brand colors see previous comment

components/giftbox.vue
```diff
@@ -0,0 +1,280 @@
...
+    }
+  }
+</style>
+
+<style lang="scss">
+  $grey: #efefef;
+
+  .modal-open .modal {
+    overflow-y: hidden;
+    height: 95vh;
```
brampauwelyn, line 162: Don't use a fixed height for the modal. 
When there's not much content it looks a bit weird.

components/giftbox.vue
```diff
@@ -0,0 +1,280 @@
...
+      width: 65%;
+      height: 65%;
+      position: relative;
+      top: 17.5%;
+    }
+  }
+</style>
+
+<style lang="scss">
+  $grey: #efefef;
```
brampauwelyn, line 158: Maybe define it in a global .scss file so you can reuse in the different components

components/giftbox.vue
```diff
@@ -0,0 +1,280 @@
...
+        vertical-align: middle;
+        display: flex;
+        white-space: nowrap;
+        overflow: hidden;
+        text-overflow: ellipsis;
+        transition: background-color 500ms;
+
+        &:hover {
+          cursor: pointer;
+          background-color: darken(#ffffff, 10%);
```
brampauwelyn, line 201: Use variables for frequently used colors
```suggestion
          background-color: darken($white, 10%);
```

components/giftbox.vue
brampauwelyn, line 134:
```suggestion
      background-color: $udb-primary-color;
```

components/giftbox.vue
```diff
@@ -0,0 +1,280 @@
+<template>
+  <div class="giftbox-wrapper">
+    <div class="giftbox-button" @click="showModal">
+      <fa icon="gift" class="giftbox-icon" />
+    </div>
+    <b-modal
+      ref="giftbox-modal"
+      class="giftbox-modal"
+      size="xl"
+      title="Nieuw in onze API en widgets"
```
brampauwelyn, line 10: Use translatable strings(i18n 🌍)

brampauwelyn, line 10: e.g.
```suggestion
      :title="$t('modal.new_features.title')"
```

components/giftbox.vue
```diff
@@ -0,0 +1,280 @@
...
+    display: flex;
+    border-top: $grey 2px solid;
+
+    .features-list {
+      width: 30%;
+      height: 87vh;
+      border-right: $grey 2px solid;
+      overflow-y: scroll;
+
+      .feature-title-block {
```
brampauwelyn, line 187: Don't nest too many levels
`Be aware that overly nested rules will result in over-qualified CSS that could prove hard to maintain and is generally considered bad practice.`

I don't think it's necessary to nest these classes they are already quite specific.

components/giftbox.vue
```diff
@@ -0,0 +1,280 @@
...
+        }
+
+        .btn-call-to-action {
+          display: inline-block;
+          overflow: hidden;
+          white-space: nowrap;
+          text-overflow: ellipsis;
+          width: fit-content;
+          max-width: 100%;
+          background: #c0120c;
```
brampauwelyn, line 252: Use variables for colors

brampauwelyn, line 252:
```suggestion
          background: $udb-primary-color;
```

components/giftbox.vue
```diff
@@ -0,0 +1,280 @@
...
+              />
+              <fa v-else icon="eye-slash" class="eye-icon" />
+            </div>
+            <p>{{ feature.title }}</p>
+          </div>
+        </div>
+        <div v-if="selectedFeature" class="features-detail">
+          <div class="title-calltoaction-container">
+            <h6>{{ selectedFeature.title }}</h6>
+            <a
```
brampauwelyn, line 40: Maybe use a bootstrap button for the call to action?

components/giftbox.vue
```diff
@@ -0,0 +1,280 @@
...
+        }
+      });  
+    },
+    methods: {
+      async fetchFeatures() {
+        this.loading = true;
+        const response = await fetch(
+          'https://www.publiq.be/uitdatabank/articles.json',
+        );
+        const data = await response.json();
```
brampauwelyn, line 95: I thin you can do something like this
```suggestion
        const { data } = await response.json();
        return data;
```

brampauwelyn:
Also try to align the design with the mockup provided in the ticket.
CTA should be at the bottom.

components/giftbox.vue
bertramakers: True, I mentioned the same in my call with @simonpubliq earlier today. We weren't sure if it would be possible already without the i18n PR merged, but in the end the function name to call in templates to do translations didn't change after all (at first it changed from `$t()` to `$i18n.t()`, but that was reverted by using `nuxt-i18n` instead of `i18next`) so we agreed he could already do it here already.

components/giftbox.vue
bertramakers: Yes, that should work. Here's more info on that syntax @simonpubliq https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#Object_destructuring

Also more info on creating objects with variables and their names as properties and values that I showed you in our call earlier today: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Object_initializer#New_notations_in_ECMAScript_2015

bertramakers: But note that we don't have Babel in the AngularJS app, so don't use this new syntax there.

brampauwelyn: > But note that we don't have Babel in the AngularJS app, so don't use this new syntax there.

Shouldn't be an issue since the component lives inside the vue app?

bertramakers: Yes it's no issue here, that's why I said don't use it _there_ 😉 But yeah it was a bit unclear at first glance

components/giftbox.vue
```diff
@@ -0,0 +1,280 @@
...
+          );
+          this.seenFeatures.add(this.selectedFeature.uid);
+        }
+      },
+    },
+  };
+</script>
+
+<style lang="scss" scoped>
+  @import '../assets/styles/_colors';
```
brampauwelyn, line 121: Better to import via the `styleResources` setting in `nuxt.config.js` see [https://dev.to/ceppeu/using-sass-global-variables-in-nuxt-js-j0k](https://dev.to/ceppeu/using-sass-global-variables-in-nuxt-js-j0k)

Maybe you could make a `main.scss` and then import it in the `nuxt.config.js`

components/giftbox.vue
```diff
@@ -0,0 +1,280 @@
...
+      width: 65%;
+      height: 65%;
+      position: relative;
+      top: 17.5%;
+    }
+  }
+</style>
+
+<style lang="scss">
+  @import '../assets/styles/_colors';
```
brampauwelyn, line 160: Same here.
Better to import via `nuxt.config.js`

components/giftbox.vue
```diff
@@ -50,10 +50,16 @@
...
-            <b-button class="btn-call-to-action" variant="outline-primary">
-              {{ selectedFeature.callToActionLabel }}
-            </b-button>
-          </div>          
+            <a
+              v-if="selectedFeature.callToAction"
+              :href="selectedFeature.callToAction"
+              target="_blank"
+            >
+              <b-button class="btn-call-to-action" squared variant="primary">
```
bertramakers, line 58: If all our buttons are supposed to be `squared`, maybe we should change it in the bootstrap variables _or_ create our own button component that uses `b-button` and just proxies most props, but also adds `squared` so we don't always have to repeat that.

components/giftbox.vue
```diff
@@ -240,12 +246,20 @@
       }
 
       .btn-call-to-action {
-          display: inline-block;
-          overflow: hidden;
-          white-space: nowrap;
-          text-overflow: ellipsis;
-          width: fit-content;
-          max-width: 100%;
+        display: inline-block;
```
bertramakers, line 247: These indentation fixes should automatically happen using [prettier](https://prettier.io/).
Best to check with @brampauwelyn how to configure that in VS Code.

components/giftbox.vue
brampauwelyn: True.
Also why is there a button inside an `<a>` tag?

components/giftbox.vue
```diff
@@ -252,6 +252,7 @@
         text-overflow: ellipsis;
         width: fit-content;
         max-width: 100%;
+        margin-bottom: 1rem;
```
bertramakers, line 253: @simonpubliq I noticed your commits are a bit inconsistent in scope which makes them a bit harder to review sometimes. Make sure to read the article I sent you on commit messages, and also try to do atomic commits. This CSS change for example is totally fine on its own, but unrelated to the change in the `fetchFeatures` method. By committing changes like this separately instead, you can make it easier to revert, cherry-pick, and review each commit when needed. 😊

It should be possible to stage changes in the same file separately, so you wouldn't have to "undo" work and then "redo" it later to commit it: https://stackoverflow.com/questions/1085162/commit-only-part-of-a-file-in-git
Apps like [Fork](https://git-fork.com/) also support this and make it even easier. 
If you need help with staging only parts of your changes, definitely ping me!

bertramakers: Sadly I don't have a go-to resource on atomic commits, but this article seems to cover most of it: https://curiousprogrammer.dev/blog/why-i-create-atomic-commits-in-git/

bertramakers:
@simonpubliq I just wanted to mention that I think the overall approach is sound, and we're just giving feedback on the details that could be perfected now. It's great that you're taking ownership of this feature and making sure it's as good as possible! 💪

components/giftbox.vue
simonpubliq: @brampauwelyn @bertramakers I was wondering if this isn't best done in a separate PR, because it is not specific to the giftbox-feature

bertramakers: I'd rather do that as soon as possible, so we don't have to go back and remove the `squared` prop later when we do override the bootstrap radius variable, or change existing usages of `b-button` to our own button component once we have one if we go that route instead.

components/giftbox.vue
```diff
@@ -76,27 +76,72 @@
...
+      cookieOptions() {
+        return {
+          maxAge: 60 * 60 * 24 * 30,
+        };
+      },
+    },
     created() {
       this.fetchFeatures().then((res) => {
         this.features = res;
+        // sync the seenFeatures with the incomming features from the api call
```
bertramakers, line 93: Nit:
```suggestion
        // sync the seenFeatures with the incoming features from the api call
```

components/giftbox.vue
```diff
@@ -76,27 +76,72 @@
...
-          process.env.newFeaturesUrl
-        );
+        const response = await fetch(process.env.newFeaturesUrl);
         const { data } = await response.json();
         this.loading = false;
         return data;
       },
+      addToSeenFeatures(uid) {
+        const seenFeatures = this.seenFeatures;
+        seenFeatures[uid] = true;
```
bertramakers, line 112: Why not make it an array of uids, e.g. `[uid1, uid2, etc]` instead of `{uid1: true, uid2: true, etc}`?

If it's just to easily check if the uid is in the array, you can use [Array.prototype.includes](https://developer.mozilla.org/nl/docs/Web/JavaScript/Reference/Global_Objects/Array/includes)

To remove a uid from the array you could use [Array.prototype.filter](https://developer.mozilla.org/nl/docs/Web/JavaScript/Reference/Global_Objects/Array/filter)

I feel like this would make the logic a bit easier to reason about when someone has to re-read this in the future to make changes.

components/giftbox.vue
```diff
@@ -76,27 +76,72 @@
...
+        this.$cookies.set('seenFeatures', seenFeatures, this.cookieOptions);
+      },
+      isFeatureSeen(uid) {
+        const seenFeatures = this.seenFeatures;
+        return !!seenFeatures[uid];
+      },
+      syncSeenFeatures() {
+        const seenFeatures = this.seenFeatures;
+
+        // if there are no features
```
bertramakers, line 122: Nit: This comment is a bit redundant, same as `// loop over seenFeatures` below. Comments describing _what_ the code does are usually not that valuable and sometimes even distracting. Comments that describe _why_ the code does something on the other hand are very useful later on.

components/giftbox.vue
```diff
@@ -76,27 +76,72 @@
...
+        if (!this.features.length > 0) {
+          this.resetSeenFeatures();
+        } else {
+          // loop over seenFeatures
+          Object.keys(seenFeatures).forEach((seenFeature) => {
```
bertramakers, line 127: This loop could probably be simplified by using https://developer.mozilla.org/nl/docs/Web/JavaScript/Reference/Global_Objects/Array/filter, if you'd use an array like suggested an earlier comment.

components/giftbox.vue
```diff
@@ -76,27 +76,72 @@
...
     created() {
       this.fetchFeatures().then((res) => {
         this.features = res;
+        // sync the seenFeatures with the incomming features from the api call
+        this.syncSeenFeatures();
+        // if there are new features
         if (this.features.length > 0) {
           this.selectedFeature = this.features[0];
-          this.seenFeatures.add(this.selectedFeature.uid);
+          this.addToSeenFeatures(this.selectedFeature.uid);
```
bertramakers, line 98: Good idea to have descriptive methods to mark features as seen or check if they are seen before, instead of working with the `this.features` object directly throughout the whole component. Makes the code easier to follow by hiding the implementation details 👍

simonpubliq:
### After I resolved comments:
Schermafbeelding 2020-07-24 om 13 35 19
Schermafbeelding 2020-07-24 om 13 35 06
Schermafbeelding 2020-07-24 om 13 34 52

assets/styles/colors.scss
```diff
@@ -0,0 +1,5 @@
+$white: #ffffff;
```
brampauwelyn, line 1: Prefix .scss files which are going to be imported with `_`

assets/styles/global.scss
```diff
@@ -0,0 +1 @@
+
```
brampauwelyn, line 1: Empty global.scss?

nuxt.config.js
```diff
@@ -86,6 +92,9 @@ export default {
       },
     ],
   },
+  styleResources: {
+    scss: ['./assets/styles/*.scss'],
```
brampauwelyn, line 96: Import the global.scss file here
```suggestion
    scss: ['./assets/styles/global.scss'],
```

components/giftbox.vue
```diff
@@ -0,0 +1,301 @@
...
+      syncSeenFeatures() {
+        const seenFeatures = this.seenFeatures;
+
+        if (!this.features.length > 0) {
+          this.resetSeenFeatures();
+        } else {
+          seenFeatures.filter((seenFeature) => {
+            return this.features.find((feature) => feature.uid === seenFeature);
+          });
+        }
```
bertramakers, line 129: I'm a bit confused by this `if/else` at this point.

If there are no features at all, it calls `resetSeenFeatures()` which sets the cookie value to `[]`. But then at the end of this method (`syncSeenFeatures`) it also sets the cookie value to `seenFeatures`, which is also `[]` in that case. So `resetSeenFeatures` seems unnecessary.

So then it could be re-written as:
```
        if (this.features.length > 0) {
          seenFeatures.filter((seenFeature) => {
            return this.features.find((feature) => feature.uid === seenFeature);
          });
        }
```

However, the code in that new `if` block (originally the `else`) also does nothing.

1. It doesn't do anything with the return value of `.filter()` (which returns a new array)
2. It will never filter anything anyway, because you're checking that a feature in `seenFeatures` is also in `this.seenFeatures`. But those two arrays are exactly the same, because `const seenFeatures = this.seenFeatures`

So that code can also be removed, and you end up with:

```
syncSeenFeatures() {
  this.$cookies.set('seenFeatures', seenFeatures, this.cookieOptions);
}
```

Unless I'm mistaken/overlooking something?

But if it's only job is storing the seenFeatures in the cookie, maybe it should be called `persistSeenFeatures` instead of `syncSeenFeatures`.

bertramakers: Additionally, if you simplify it to `persistSeenFeatures` as suggested, you can also use it in `addToSeenFeatures` instead of doing `this.$cookies.set()` there.

bertramakers: Ah, I was confusing `this.seenFeatures` with `this.features`!

bertramakers: Still, the filter does nothing right now and you should assign the return value to something 😉

simon-debruijn:
```
seenFeatures.filter((seenFeature) => {
            return this.features.find((feature) => feature.uid === seenFeature);
});
```
 has to be reassigned to seenFeatures

PR #9, opened 2020-07-28
Feature/iii 3378

simon-debruijn:
### Changed
Changed layout of login page to resemble the old angular login page as close as possible

---
Ticket: https://jira.uitdatabank.be/browse/III-3378

Schermafbeelding 2020-07-28 om 14 39 24
Schermafbeelding 2020-07-28 om 14 39 44
Schermafbeelding 2020-07-28 om 14 40 00
Schermafbeelding 2020-07-28 om 14 40 08


pages/login.vue
```diff
@@ -161,7 +161,8 @@
               <p class="accent">500</p>
               <h2>{{ $t('main.channels') }}</h2>
             </blockquote>
-            <p class="info">{{ $t('main.channels_info') }}</p>
+            <!-- eslint-disable-next-line vue/no-v-html -->
+            <p class="info" v-html="$t('main.channels_info')"></p>
```
bertramakers, line 165: Why do we need the `v-html` here?

simon-debruijn: @bertramakers there is a link nested in an <a> tag in the translation

bertramakers: Alright. The Vue docs recommend using component interpolation for that though:

> Dynamically localizing arbitrary HTML on your website can be very dangerous because it can easily lead to XSS vulnerabilities. Only use HTML interpolation on trusted content and never on user-provided content.
> We recommended using the component interpolation feature.
https://kazupon.github.io/vue-i18n/guide/formatting.html#html-formatting

You can find the docs for component interpolation here: https://kazupon.github.io/vue-i18n/guide/interpolation.html#basic-usage

PR #14, opened 2020-08-04
Feature/iii 3300 - productions overview

simon-debruijn:
### Added

-  pagination component
-  button to create productions
-  stylings to match old angular app for productions

After changes:

Schermafbeelding 2020-08-04 om 10 46 38

---
Ticket: https://jira.uitdatabank.be/browse/III-3300

### DIFF
--- a/assets/styles/_custom-bootstrap.scss
+++ b/assets/styles/_custom-bootstrap.scss
@@ -1,4 +1,4 @@
-@import './_colors.scss';
+@import 'colors.scss';
 // color overrides
 $primary: $udb-blue;
 $secondary: $udb-grey;
--- a/assets/styles/_custom-bootstrap.scss
+++ b/assets/styles/_custom-bootstrap.scss
@@ -1,7 +1,9 @@
 @import './_colors.scss';
-// Your variable overrides
+// color overrides
 $primary: $udb-blue;
 $secondary: $udb-grey;
+
+// button overrides
 $btn-border-radius: 0;
 $btn-padding-y: 0.275rem;
 $btn-padding-x: 0.5rem;
@@ -14,6 +16,15 @@ $btn-padding-x: 0.5rem;
     border-color: #adadad !important;
   }
 }
+
+// pagination overrides
+$pagination-color: #222;
+$pagination-active-bg: #f3453f;
+$pagination-active-color: $white;
+$pagination-hover-bg: #f3453f;
+$pagination-hover-color: $white;
+$pagination-border-color: $udb-grey;
+$pagination-focus-box-shadow: none;
 
 // Bootstrap and its default variables
 @import '../../node_modules/bootstrap/scss/bootstrap';
--- a/assets/styles/_custom-bootstrap.scss
+++ b/assets/styles/_custom-bootstrap.scss
@@ -3,6 +3,9 @@
 $primary: $udb-blue;
 $secondary: $udb-grey;
 
+//font-size overrides
+$small-font-size: 65%;
+
 // button overrides
 $btn-border-radius: 0;
 $btn-padding-y: 0.275rem;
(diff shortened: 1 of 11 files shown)


layouts/default.vue
```diff
@@ -51,6 +51,30 @@
...
+    display: flex;
+    justify-content: left;
+    align-items: flex-start;
+    text-align: left;
+  }
+
+  .title {
+    font-family: 'Open Sans', Helvetica, Arial, sans-serif;
+    display: block;
+    font-weight: 400;
```
bertramakers, line 69: I see that it's impossible to get it to look exactly the same, but I'd go with font-weight 300 here. It looks better and closer to the original in the AngularJS app

pages/manage/productions/index.vue
```diff
@@ -1,47 +1,35 @@
 <template>
   <div class="wrapper">
-    <div>
+    <div class="container-fluid">
       <h1 class="title">
-       $t('menu.productions')
+        {{ $t('menu.productions') }}
+        <small>
```
bertramakers, line 6: This link is still too big compared to other pages in the AngularJS app. It's now `19.2px`, but it should be `15.6px`. The `<small>` in the AngularJS app was overwritten to be font-size `65%` instead of the default `85%` in Bootstrap 3 (and `80%` in Bootstrap 4). So we should probably also change it to `65%`.

assets/styles/_custom-bootstrap.scss
```diff
@@ -15,6 +17,15 @@ $btn-padding-x: 0.5rem;
...
 }
 
+// pagination overrides
+$pagination-color: #222;
+$pagination-active-bg: #f3453f;
+$pagination-active-color: $white;
+$pagination-hover-bg: #f3453f;
+$pagination-hover-color: $white;
+$pagination-border-color: $udb-grey;
+$pagination-focus-box-shadow: none;
```
bertramakers, line 27: It looks similar, but in the old app the buttons were square (34x34 px) and a slightly smaller font-size. In the Vue app they look more rectangular (33x38 px) and have a bigger font-size (16px vs 15px).

AngularJS:
Screenshot 2020-08-04 at 12 09 42

Vue:
Screenshot 2020-08-04 at 12 10 04

bertramakers: But if it's too hard to get to look more or less the same, leave it as it is.

i18n/nl.json
```diff
@@ -44,6 +44,11 @@
     "notifications": "Meldingen",
     "logout": "Afmelden"
   },
+  "actions": {
+    "add": "toevoegen",
```
bertramakers, line 48: Should we use the key `create` here? That's also what we call it in the routes.

i18n/nl.json
```diff
@@ -44,6 +44,11 @@
...
+    "previous": "Vorige",
+    "next": "Volgende"
```
bertramakers, line 50: Maybe move the `previous` and `next` actions to a `pagination` object instead of `actions`. That way the context is clearer.

simon-debruijn: the only counterargument I can give is that these "actions" can be reused for different purposes. And then we would have duplicate labels for the same thing

simon-debruijn:
after comments resolved: 

Schermafbeelding 2020-08-04 om 15 12 39

i18n/nl.json
simon-debruijn: @bertramakers

bertramakers: I don't think having duplicate translations for different contexts is a bad thing. We don't determine the exact copy as developers, the PO does. For them, having "duplicate" translations would have the advantage of:
- Easily finding the context of a specific key (e.g. `actions.next` is ambiguous, `pagination.next` is a lot clearer)
- Being able to differentiate in the final copy if they want (e.g. `rol toevoegen`, `label toevoegen`, etc instead of one `toevoegen`)

bertramakers: It's always a tough balance to find though. IMO `productions.pagination.next` would be too specific (we at least want the "next" and "previous" to be consistent in all paginations), while `actions.next` is too broad (who knows where this will be used over time).

PR #16, opened 2020-08-04
III-3347 Make nuxt env variables public runtime config

bertramakers:
### Removed

- Removed usages of `process.env.` outside of the `publicRuntimeConfig` in `nuxt.config.js`. Otherwise these env variables would get interpolated (i.e. hardcoded) during build, preventing us from running the same build on different environments with different `.env` files. See https://nuxtjs.org/blog/moving-from-nuxtjs-dotenv-to-runtime-config/

---
Ticket: https://jira.uitdatabank.be/browse/III-3347

### DIFF
--- a/nuxt.config.js
+++ b/nuxt.config.js
@@ -96,6 +96,7 @@ export default {
       {
         default: true,
         name: 'uitdatabank',
+        url: 'https://sockets.uitdatabank.dev',
       },
     ],
   },
--- a/nuxt.config.js
+++ b/nuxt.config.js
@@ -73,13 +73,20 @@ export default {
    ** See https://nuxtjs.org/api/configuration-build/
    */
   build: {},
-  env: {
+  publicRuntimeConfig: {
     apiKey: process.env.API_KEY,
     apiUrl: process.env.API_URL,
     legacyAppUrl: process.env.LEGACY_APP_URL,
     authUrl: process.env.AUTH_URL,
-    socketUrl: process.env.SOCKET_URL,
     newFeaturesUrl: process.env.NEW_FEATURES_URL,
+    io: {
+      sockets: [
+        {
+          name: 'uitdatabank',
+          url: process.env.SOCKET_URL,
+        },
+      ],
+    }
   },
   router: {
     middleware: ['auth'],
@@ -88,7 +95,7 @@ export default {
     sockets: [
       {
         default: true,
-        url: 'https://sockets.uitdatabank.dev',
+        name: 'uitdatabank',
       },
     ],
   },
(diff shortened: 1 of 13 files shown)


nuxt.config.js
```diff
@@ -88,6 +95,7 @@ export default {
     sockets: [
       {
         default: true,
+        name: 'uitdatabank',
         url: 'https://sockets.uitdatabank.dev',
```
bertramakers, line 99: Note that this URL is still hardcoded for now, due to a bug in `nuxt-socket-io`. See https://github.com/richardeschloss/nuxt-socket-io/issues/132

nuxt.config.js
```diff
@@ -73,25 +73,25 @@ export default {
    ** See https://nuxtjs.org/api/configuration-build/
    */
   build: {},
-  env: {
+  publicRuntimeConfig: {
     apiKey: process.env.API_KEY,
```
brampauwelyn, line 77: Shouldn't the API_KEY be under `privateRuntimeConfig`?

bertramakers: The Nuxt docs mention:

> privateRuntimeConfig is only available on server

But the app builds in SPA mode and we need to make http calls to the (PHP) backend from the client in the browser. So we need the api key to be available there. This is also the case with the AngularJS app. I think even if we'd manage to run it in universal mode (which gave some problems in the Vagrant box), we'd still need it for after the initial render of the first page is done and we need to perform API requests later inside the browser.

bertramakers: This is not an issue limited to this Nuxt app btw. Google summarizes it nicely:

> API keys are generally not considered secure; they are typically accessible to clients, making it easy for someone to steal an API key. Once the key is stolen, it has no expiration, so it may be used indefinitely, unless the project owner revokes or regenerates the key. While the restrictions you can set on an API key mitigate this, there are better approaches for authorization.
>
> When to use API keys
> 
> An API may restrict some or all of its methods to require API keys. It makes sense to do this if:
>
> - You do want to block anonymous traffic. API keys identify an application's traffic for the API producer, in case the application developer needs to work with the API producer to debug an issue or show their application's usage.
> - You want to control the number of calls made to your API.
> - You want to identify usage patterns in your API's traffic. You can see application usage in APIs & services.
> - You want to filter logs by API key.
>
> API keys cannot be used for:
>
> - Identifying individual users — API keys don't identify users, they identify projects.
> - Secure authorization.
> - Identifying the creators of a project.

https://cloud.google.com/endpoints/docs/openapi/when-why-api-key#security_of_api_keys

Nevertheless, even if Nuxt could route all http requests through some kind of proxy and add the api key inside that proxy, it would still be insecure. Because then someone could just call the proxy without an api key and achieve the same results as if they had stolen one.

brampauwelyn: Alright thx for the explanation!

PR #17, opened 2020-08-06
change arrow-functions to normal functions to be able to use the context of this

simon-debruijn:
### Changed

- change arrow-functions to normal functions to be able to use the context of this

### Fixed

-  login and logout creating an error


bertramakers:
Thanks for fixing this so quickly!

PR #21, opened 2020-08-14
add functionality to delete event from production

simon-debruijn:
### Added

- functionality to delete event from production

screenshot:
Schermafbeelding 2020-08-14 om 16 16 02

---
Ticket: https://jira.uitdatabank.be/browse/III-3304

### DIFF
--- a/pages/manage/productions/index.vue
+++ b/pages/manage/productions/index.vue
@@ -32,7 +32,7 @@
             @clickDeleteEvent="handleClickDeleteEvent"
           />
           <delete-modal
-            :production-title="selectedProductionName"
+            :production-name="selectedProductionName"
             :event-name="toBeDeletedEventName"
             @confirmDelete="handleConfirmDeleteEvent"
           />
--- a/pages/manage/productions/index.vue
+++ b/pages/manage/productions/index.vue
@@ -81,10 +81,12 @@
         isLoadingEvents: true,
         events: [],
         toBeDeletedEventId: '',
-        toBeDeletedEventName: '',
       };
     },
     computed: {
+      locale() {
+        return this.$i18n.locale;
+      },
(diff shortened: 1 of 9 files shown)


pages/manage/productions/index.vue
```diff
@@ -66,6 +80,8 @@
 
         isLoadingEvents: true,
         events: [],
+        toBeDeletedEventId: '',
+        toBeDeletedEventName: '',
```
geoffreydhuyvetters, line 84: can't we avoid this by passing the id to the modal as discussed? 

delete -> open modal with event information and production passed 
confirm -> trigger action back in index with the eventId passed down

simon-debruijn: yes, I could only pass the id down to the modal and then search for the eventById in the modal itself to retrieve the name. but toBeDeletedEventId still would have to be a data prop in index no?

Is this what you mean, or is it something different?

geoffreydhuyvetters: No, I don't think it needs to be, we can communicate through events from the modal. But do correct me if I'm oversimplifying it?

pages/manage/productions/index.vue
```diff
@@ -26,9 +29,18 @@
             :has-adding-error="hasAddingEventToProductionError"
             @addEventToProduction="handleAddEventToProduction"
             @inputEventId="handleInputEventId"
+            @clickDeleteEvent="handleClickDeleteEvent"
           />
+          <delete-modal
+            :production-name="selectedProductionName"
+            :event-name="toBeDeletedEventName"
+            @confirmDelete="handleConfirmDeleteEvent"
```
geoffreydhuyvetters, line 37: I think @confirm is ok here since we're in the delete modal

PR #22, opened 2020-08-18
fix cursor pointer for a tags in productions

simon-debruijn:
### Fixed

- fix cursor pointer for a tags in productions

### DIFF
--- a/pages/manage/productions/index.vue
+++ b/pages/manage/productions/index.vue
@@ -240,10 +240,6 @@
     a {
       font-weight: 400;
       color: $udb-blue;
-
-      &:hover {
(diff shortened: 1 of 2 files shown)


pages/manage/productions/index.vue
```diff
@@ -240,6 +240,10 @@
     a {
       font-weight: 400;
       color: $udb-blue;
+
+      &:hover {
+        cursor: pointer;
+      }
```
bertramakers, line 246: This will probably be an issue for every `a` tag that doesn't have an `href` though.
So we should probably add it to some more global scope?

PR #24, opened 2020-08-19
Feature/iii 3310 - Add basic setup similarity ui

simon-debruijn:
### Added

- basic setup similarity ui

---
Ticket: https://jira.uitdatabank.be/browse/III-3310

screenshot:
Schermafbeelding 2020-08-19 om 10 58 21


components/productions/create/event.vue
```diff
@@ -0,0 +1,87 @@
...
+      },
+    },
+    computed: {
+      period() {
+        const parsedStart = this.parseDate(this.startDate);
+        const parsedEnd = this.parseDate(this.endDate);
+        if (parsedStart === parsedEnd) {
+          return parsedStart;
+        }
+        return `${parsedStart} - ${parsedEnd}`;
```
bertramakers, line 52: It would probably be better to use the calendar summary here.
Start and end date can be very vague for events that run multiple times over a longer period of time.
@corneelwille what do you think?

You can fetch it by requesting https://<api>/events/{id}/calsum?format={sm|md|lg}&langCode=nl_BE

For example: https://udb-silex-acc.uitdatabank.be/events/5781ac48-6a92-4ee5-bb67-c0187349b3b4/calsum?format=lg&langCode=nl_BE

components/productions/index/event.vue
```diff
@@ -67,9 +67,12 @@
...
       period() {
-        return `${this.parseDate(this.event.startDate)} - ${this.parseDate(
-          this.event.endDate,
-        )}`;
+        const parsedStart = this.parseDate(this.event.startDate);
+        const parsedEnd = this.parseDate(this.event.endDate);
+        if (parsedStart === parsedEnd) {
+          return parsedStart;
+        }
+        return `${parsedStart} - ${parsedEnd}`;
```
bertramakers, line 75: Same suggestion about using the calendar summary.

layouts/default.vue
```diff
@@ -61,13 +61,16 @@
...
-  .title {
+  h1 {
     font-family: 'Open Sans', Helvetica, Arial, sans-serif;
     display: block;
-    font-weight: 300;
-    font-size: 1.6rem;
-    line-height: 3.74rem;
+    font-weight: 300 !important;
+    font-size: 1.6rem !important;
+    line-height: 3.74rem !important;
```
bertramakers, line 69: As @brampauwelyn recently suggested, we should try to avoid `!important` unless there's really no other way.

> The only thing that can override an !important tag is another !important tag. By using it one once, you potentially end up with a CSS file that is full of !important tags, which is not ideal.

https://uxengineer.com/css-specificity-avoid-important-css/

Seeing this change now, I wonder why you added them? Does that mean the `h1` tags never had these styles before either? Or is there a `h1` you added now in this PR that didn't get those styles?

simon-debruijn: the heading styles was on the class .title, I think if we want to reuse headings it should be on the h1-tag, so it can be used also without the .title class. But the !important seems to be needed to override bootstrap defaults for headings

simon-debruijn: unless I can override these by overriding global variables of bootstrap

bertramakers: Are you sure? That seems strange. Bootstrap doesn't use a very specific selector, just `h1`. It doesn't use `!important` either. So it should be easy to override? Unless Bootstrap gets included _after_ our own styling of course 🤔 That would be a big problem in general down the line. So maybe double-check that first.

If that fails, you could try a more specific selector like `body h1` for example. Even if Bootstrap overrides the style later based on the `h1` selector, the `body h1` selector will have precedence because it's more specific. The article I linked also explains that concept.

components/productions/create/event.vue
```diff
@@ -42,19 +46,28 @@
...
-        const parsedEnd = this.parseDate(this.endDate);
-        if (parsedStart === parsedEnd) {
-          return parsedStart;
-        }
-        return `${parsedStart} - ${parsedEnd}`;
+      locale() {
+        return this.$i18n.locale;
+      },
+      eventId() {
+        return parseEventId(this.id);
```
bertramakers, line 59: I'd suggest parsing the id before passing it as a prop to the components. Then you can avoid having an `id` and `eventId` in these components which is a bit confusing.

bertramakers (changes requested):

pages/manage/productions/create.vue
```diff
@@ -1,15 +1,110 @@
...
+          v-for="event in suggestedEvents"
+          :id="parseEventId(event['@id'])"
+          :key="event['@id']"
+          :type="getEventType(event.terms)"
+          :title="event.name[locale]"
+          :start-date="event.startDate"
+          :end-date="event.endDate"
+          :image-url="event.image"
+          :description="event.description[locale]"
+        />
```
bertramakers, line 24: Out of curiosity, can you explain why you an `event` component with `v-for` here and pass it only the props it needs, while in the ` pages/manage/productions/index.vue` you use an additional `events` (plural) component and pass it the whole event(s) data?

I have no preference for one or the other, but ideally, we'd pick one approach and stick with it for consistency I believe.

pages/manage/productions/index.vue
```diff
@@ -209,6 +210,9 @@
           })
           .filter((production) => production !== undefined);
       },
+      parseEventId(id) {
+        return parseId(id);
+      },
```
bertramakers, line 215: This doesn't seem to be used in this component?

pages/manage/productions/create.vue
simon-debruijn: The `events` component is a collection of multiple events, a list. In the create page there are always only 2 events. Either way, I've changed both `event` components to take id, name, ... as separate props now.

bertramakers (approved):

api/events.js
```diff
@@ -40,3 +36,15 @@ export const findByIds = (apiUrl, headers, fetch) => async (eventIds) => {
   const events = await Promise.all(mappedEvents);
   return events.filter((event) => !event.status);
 };
+
+export const getCalendarSummary = (apiUrl, headers, fetch) => async ({
+  id,
+  locale,
```
geoffreydhuyvetters, line 42: if you want to make sure it uses the defaults when you don't pass anything you need to add ` = {}` after the destructuring

api/events.js
```diff
@@ -40,3 +36,15 @@ export const findByIds = (apiUrl, headers, fetch) => async (eventIds) => {
...
+  format = 'lg',
+}) => {
+  const url = `${apiUrl}/events/${id.toString()}/calsum?format=${format}&langCode=${locale}_BE`;
+  const res = await fetch(url, {
+    headers: headers(),
+  });
+  return res.text();
```
geoffreydhuyvetters, line 49: Shouldn't this be JSON?

components/productions/create/event.vue
```diff
@@ -0,0 +1,95 @@
...
+        default: '',
+      },
+      startDate: {
+        type: String,
+        default: '',
+      },
+      endDate: {
+        type: String,
+        default: '',
+      },
```
geoffreydhuyvetters, line 37: A bit weird that start/endDate are strings here

components/productions/index/event.vue
```diff
@@ -48,40 +48,48 @@
...
       handleClickDelete(eventId) {
         this.$emit('clickDelete', eventId);
       },
+      async getPeriod() {
+        return await this.$api.events.getCalendarSummary({
+          id: this.id,
+          locale: this.locale,
+        });
+      },
     },
```
geoffreydhuyvetters, line 93: can't we just call the api helper directly in created? This just adds an extra layer of abstraction

components/productions/create/event.vue
```diff
@@ -0,0 +1,95 @@
...
+      this.period = await this.getPeriod();
+    },
+    methods: {
+      async getPeriod() {
+        return await this.$api.events.getCalendarSummary({
+          id: this.id,
+          locale: this.locale,
+        });
+      },
+    },
```
geoffreydhuyvetters, line 67: Can't we just call the API helper directly? this just adds another layer of abstraction that isn't needed

pages/manage/productions/create.vue
```diff
@@ -1,15 +1,110 @@
...
+        return MockSuggestedEvents;
+      },
+      getEventType(terms) {
+        const foundTerm =
+          terms.find((term) => term.domain === 'eventtype') || {};
+        return foundTerm.label ? foundTerm.label : '';
+      },
+      parseEventId(id) {
+        return parseId(id);
+      },
```
geoffreydhuyvetters, line 74: Why do we need a Vue method for this? We can just call the imported method directly

api/events.js
bertramakers: No, this endpoint returns plain text for some reason apparently.

PR #27, opened 2020-08-20
Feature/iii 3312 - link events to new production 

simon-debruijn:
### Added

- link events to new production 

---
Ticket: https://jira.uitdatabank.be/browse/III-3312

### DIFF
--- a/api/api.js
+++ b/api/api.js
@@ -41,40 +41,27 @@ export default (authUrl, apiUrl, apiKey, tokenCallback) => {
     return getHeaders(token, apiKey);
   };
 
+  const args = [apiUrl, headersCallback, fetch];
+
   return {
     events: {
-      findToModerate: events.findToModerate(apiUrl, headersCallback, fetch),
-      findById: events.findById(apiUrl, headersCallback, fetch),
-      findByIds: events.findByIds(apiUrl, headersCallback, fetch),
-      getCalendarSummary: events.getCalendarSummary(
-        apiUrl,
-        headersCallback,
-        fetch,
-      ),
+      findToModerate: events.findToModerate(...args),
+      findById: events.findById(...args),
+      findByIds: events.findByIds(...args),
+      getCalendarSummary: events.getCalendarSummary(...args),
     },
     productions: {
-      find: productions.find(apiUrl, headersCallback, fetch),
-      addEventById: productions.addEventById(apiUrl, headersCallback, fetch),
-      deleteEventById: productions.deleteEventById(
-        apiUrl,
-        headersCallback,
-        fetch,
-      ),
-      getSuggestedEvents: productions.getSuggestedEvents(
-        apiUrl,
-        headersCallback,
-        fetch,
-      ),
-      skipSuggestedEvents: productions.skipSuggestedEvents(
-        apiUrl,
-        headersCallback,
-        fetch,
-      ),
+      find: productions.find(...args),
+      addEventById: productions.addEventById(...args),
+      deleteEventById: productions.deleteEventById(...args),
+      getSuggestedEvents: productions.getSuggestedEvents(...args),
+      skipSuggestedEvents: productions.skipSuggestedEvents(...args),
+      createWithEvents: productions.createWithEvents(...args),
     },
     user: {
-      getMe: user.getMe(apiUrl, headersCallback, fetch),
-      getPermissions: user.getPermissions(apiUrl, headersCallback, fetch),
-      getRoles: user.getRoles(apiUrl, headersCallback, fetch),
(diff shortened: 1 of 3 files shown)


api/api.js
```diff
@@ -41,40 +41,27 @@ export default (authUrl, apiUrl, apiKey, tokenCallback) => {
...
-      findByIds: events.findByIds(apiUrl, headersCallback, fetch),
-      getCalendarSummary: events.getCalendarSummary(
-        apiUrl,
-        headersCallback,
-        fetch,
-      ),
+      findToModerate: events.findToModerate(...args),
+      findById: events.findById(...args),
+      findByIds: events.findByIds(...args),
+      getCalendarSummary: events.getCalendarSummary(...args),
```
bertramakers, line 51: Nice improvement in readability 👍

api/productions.js
```diff
@@ -1,3 +1,7 @@
+/* eslint-disable no-console */
+import MockSuggestedEvents from '../assets/suggested-events';
+const isDevelopmentEnviroment = process.env.NODE_ENV === 'development';
```
bertramakers, line 3: All `process.env.` variables are hardcoded during a build.

So usually, we should use the `publicRuntimeConfig` in `nuxt.config.js` instead. We can access those variables through the Nuxt context in plugins, components, middleware, etc.

If you cannot access the Nuxt context because your code is a vanilla JS function outside of Nuxt, you'll need to pass the config you need as arguments to those functions. An example where we do this is here: https://github.com/cultuurnet/udb3-frontend/blob/main/plugins/api.js#L4-L10

See also the official documentation: https://nuxtjs.org/blog/moving-from-nuxtjs-dotenv-to-runtime-config/

> At runtime, Node.js automatically loads environment variables into process.env so that they are available to use in your application. The reference to the environment variable is replaced with the correct value. For example, if you had an API_SECRET key with the value of 'my-secret' then in your application where you had used process.env.API_SECRET this would be replaced with the value of my-secret.
> 
> __Values are read during build time and persisted in the webpack bundle. Therefore if we change our API_SECRET we will need to rebuild our application so that it can read the new value.__

(Emphasis mine)

**However in this case of `process.env.NODE_ENV` I think it would be fine because the variable is either `dev` if we're running `yarn dev`, and `production` when we run `yarn build` to build it to run on the servers.** In other words, the variable indicates what kind of build you're running and thus requires a re-build anyway to change.

However there is one other issue: If we run `yarn build` locally to run the app in production mode in our local Nuxt server in the Vagrant box, it will also be running as `process.env.NODE_ENV === production`. So it would still try to fetch actual suggestions, while we'd probably want to fetch the mock suggestions even though we're running in `production` mode.

So I would still advise putting a custom `environment` variable in nuxt.config.js instead, so we can set that to `development` for our local installs (both with `yarn dev` and `yarn build`), and also have more options like `acceptance`, `test`, and `production` instead of just `development` & `production`.

(cc @geoffreydhuyvetters because I think you might have missed this change to how we deal with environment variables, we spotted this while setting up the build pipeline while you were on vacation)

api/productions.js
```diff
@@ -1,3 +1,7 @@
+/* eslint-disable no-console */
```
bertramakers, line 1: Instead of disabling this rule for the whole file, it might be better to create some kind of wrapper function around `console.log()` that takes the same arguments and passes them to console.log, but first checks if the app is running in development mode and if it isn't it doesn't do anything.

Something like:
```
debug => (...args) {
  if (isDevelopmentMode) { // See my other comment about `process.env` variables
    /* eslint-disable no-console */
    console.log(...args);
    /* eslint-enable no-console */
  }
}
```

api/productions.js
geoffreydhuyvetters, line 3: Did we test this in a production build?

api/productions.js
```diff
@@ -52,10 +56,18 @@ export const deleteEventById = (apiUrl, headers, fetch) => async (
 };
 
 export const getSuggestedEvents = (apiUrl, headers, fetch) => async () => {
-  const url = `${apiUrl}/productions/suggestion/`;
+  // TODO: change suggestion to suggestion/
```
bertramakers, line 59: Seems like you forgot a to-do here.

api/productions.js
```diff
@@ -52,10 +56,18 @@ export const deleteEventById = (apiUrl, headers, fetch) => async (
...
+  const url = `${apiUrl}/productions/suggestion`;
   const res = await fetch(url, {
     headers: headers(),
   });
+  if (isDevelopmentEnviroment) {
```
geoffreydhuyvetters, line 64: might make sense to do this before the fetch, no?

PR #31, opened 2020-08-25
Linking when 1 event is already linked to a production

geoffreydhuyvetters:
### Changed

- When an event is already linked to a production, we must link the other event to that production

---

Ticket: https://jira.uitdatabank.be/browse/III-3313

### DIFF
--- a/pages/manage/productions/create.vue
+++ b/pages/manage/productions/create.vue
@@ -18,16 +18,19 @@
           :type="getEventType(suggestedEvent.terms)"
           :title="suggestedEvent.name[locale]"
           :image-url="suggestedEvent.image"
+          :production-name="
+            suggestedEvent.production ? suggestedEvent.production.title : ''
+          "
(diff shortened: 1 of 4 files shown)


pages/manage/productions/create.vue
```diff
@@ -123,9 +131,9 @@
         return this.$i18n.locale;
       },
       suggestedEventIds() {
```
bertramakers, line 133:
```suggestion
      suggestedEventIdsWithoutProduction() {
```
Just a minor suggestion, to avoid confusion with `suggestedEventsWithProduction` (that one now seems a subset of `suggestedEventIds` to me at first glance, while it's actually two opposites)

bertramakers: Remember to also change the usage(s)

bertramakers (approved):

bertramakers (approved):

PR #32, opened 2020-08-26
Linking when there are multiple productions

geoffreydhuyvetters:
### Added

- show radio buttons when we have multiple productions
- merge productions depending on the selection

Ticket: 

- https://jira.uitdatabank.be/browse/III-3314
- https://jira.uitdatabank.be/browse/III-3439


simon-debruijn (approved):

pages/manage/productions/create.vue
```diff
@@ -3,86 +3,105 @@
     <section class="container-fluid">
       <h1 class="title create-title">{{ $t('productions.create') }}</h1>
 
-      <p>
+      <!-- <p>
```
bertramakers, line 6: This commented code can be left out probably?

pages/manage/productions/create.vue
```diff
@@ -3,86 +3,105 @@
...
+            autocomplete="off"
+            :disabled="availableProductions.length > 0"
+            @input="handleInputProductionName"
+            @focus="handleFocusProductionName"
+          />
+          <section
+            v-show="showSuggestedProductions"
+            class="container-table-production-name"
+          >
+            <table class="table table-hover">
```
bertramakers, line 48: Why are we using a table here? Isn't it just a list?

pages/manage/productions/create.vue
```diff
@@ -199,23 +232,40 @@
         }
       },
       async linkEventsToNewProduction() {
-        await this.$api.productions.createWithEvents({
+        const response = await this.$api.productions.createWithEvents({
           name: this.productionName,
           eventIds: this.suggestedEventIdsWithoutProduction,
         });
+        if (response.status) {
```
bertramakers, line 239: Shouldn't we also reset the messages to an empty array if the response of a re-try was succesfull?
Otherwise the message will stay visible and the suggestions won't be refreshed.

bertramakers (changes requested):

pages/manage/productions/create.vue
geoffreydhuyvetters: I think that happens on line 254 in this file

pages/manage/productions/create.vue
geoffreydhuyvetters: not new to this PR, but idd we can change this. I'll do this in another PR focussing on the semantics a bit (or when we tackle the blur/focus bug)

pages/manage/productions/create.vue
bertramakers: Yes indeed, my bad!

pages/manage/productions/create.vue
bertramakers: Okay. I just noticed that the autocomplete is also not navigatable by keyboard. Other autocomplete fields in the AngularJS pages do have this (for example the field for entering the location of a new event). And it's not uncommon for users to pick a result with the arrow keys and enter key when they're already typing anyway.

Maybe we should look into a 3rd party component to handle this autocomplete instead? Like https://mattzollinhofer.github.io/vue-typeahead-bootstrap-docs/ (Not to be confused with `vue-bootstrap-typeahead` which the one I linked is a fork of, because the original did not support keyboard navigation either.)

bertramakers: This is the github: https://github.com/mattzollinhofer/vue-typeahead-bootstrap/
And npm: https://www.npmjs.com/package/vue-typeahead-bootstrap

geoffreydhuyvetters: Yeah that was my preferred solution as well, but it's not really in the scope of this PR.

geoffreydhuyvetters: There is even an option to do it with Bootstrap, but that didn't yield the wanted results.

geoffreydhuyvetters:
> Ah, it was the eventSimilarityScore. I had missed that because the diff of the PR as a whole was a bit unclear. I see why you left that. In any case we can always re-add it later 👍

We can always revert the commit when we need it back, no worries 👍

pages/manage/productions/create.vue
bertramakers: Okay, can we pick that up in the PR focusing on semantics as well then?

bertramakers (approved):

PR #35, opened 2020-08-26
Limit to 15 productions on a page

geoffreydhuyvetters:
### Changed

- limit number of productions on a page to 15

---

Ticket: https://jira.uitdatabank.be/browse/III-3449

### DIFF
--- a/pages/manage/productions/index.vue
+++ b/pages/manage/productions/index.vue
@@ -4,9 +4,9 @@
       <h1 class="title">
         {{ $t('menu.productions') }}
         <small>
-          <nuxt-link to="productions/create">
-            {{ $t('productions.create') }}
-          </nuxt-link>
+          <nuxt-link to="productions/create">{{
+            $t('productions.create')
+          }}</nuxt-link>
         </small>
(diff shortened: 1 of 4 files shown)


pages/manage/productions/index.vue
```diff
@@ -74,7 +74,7 @@
     data() {
       return {
         pagesProductions: 1,
-        productionsPerPage: 30,
+        productionsPerPage: 10,
```
bertramakers, line 77: Should be `15` according to the ticket 😉

bertramakers:
The ticket mentions `to avoid a double scroll bar`. I think we have to remove the `overflow-y` and `max-height` from the table body to achieve that. Otherwise it will keep showing an (empty) scroll bar because`overflow-y` is specifically set  to `scroll` instead of `auto`.

Not sure why the `max-height` was necessary. I think it's overkill and can be removed.

Screenshot 2020-08-26 at 15 55 57

bertramakers (approved):

simon-debruijn (approved):

PR #36, opened 2020-08-26
Add target and rel to external link

geoffreydhuyvetters:
### Changed

- add target and rel to external link

---

Ticket: https://jira.uitdatabank.be/browse/III-3443

### DIFF
--- a/pages/login.vue
+++ b/pages/login.vue
@@ -233,7 +233,13 @@
           <div class="col-sm-6 footer-logo-and-language">
             <p class="footer-by">
               <span>{{ $t('footer.by') }}</span>
-              <a href="http://www.publiq.be" class="footer-link">publiq vzw</a>
+              <a
+                href="http://www.publiq.be"
+                target="_blank"
+                rel="noopener noreferrer"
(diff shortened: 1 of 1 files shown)


pages/login.vue
```diff
@@ -233,7 +233,13 @@
...
             <p class="footer-by">
               <span>{{ $t('footer.by') }}</span>
-              <a href="http://www.publiq.be" class="footer-link">publiq vzw</a>
+              <a
+                href="http://www.publiq.be"
+                target="_blank"
+                rel="noopener noreferrer"
+                class="footer-link"
+                >publiq vzw</a
+              >
```
bertramakers, line 242: This is okay by itself but doesn't fix the issue. The issue is that this link _in the AngularJS app_ which we show on the dashboard inside the iframe doesn't have the `target="blank"` attribute.

bertramakers: Although, we shouldn't even use `target="_blank"` in this footer on the start page. No other links in the footer on this page have it.

bertramakers:
Closing because it's the incorrect code base.

PR #37, opened 2020-08-27
changed productions info copy

geoffreydhuyvetters:
### Changed

- productions info copy changes

---

Ticket: https://jira.uitdatabank.be/browse/III-3301


bertramakers (approved):
Good that you also changed the French copy, I probably would've forgotten about that 😄

PR #38, opened 2020-08-27
Use third party solution for autocomplete

geoffreydhuyvetters:
### Changed

- use vue-typeahead-bootstrap for autocomplete

---

Ticket: https://jira.uitdatabank.be/browse/III-3450

### DIFF
--- a/pages/manage/productions/create.vue
+++ b/pages/manage/productions/create.vue
@@ -28,44 +28,14 @@
           <label for="production-name">{{
             $t('productions.production_name')
           }}</label>
-          <b-input
-            id="production-name"
+          <vue-typeahead-bootstrap
             v-model="productionName"
-            autocomplete="off"
+            class="production-input"
             :disabled="availableProductions.length > 0"
-            @input="handleInputProductionName"
-            @focus="handleFocusProductionName"
+            :data="suggestedProductionNames"
+            @keyup="handleProductionInputKeyup"
+            @hit="handleProductionInputHit"
           />
-          <section
-            v-show="showSuggestedProductions"
-            class="container-table-production-name"
-          >
-            <table class="table table-hover">
-              <tbody v-if="!isLoadingSuggestedProductions">
-                <tr
-                  v-for="suggestedProduction in suggestedProductions"
-                  :key="suggestedProduction.production_id"
-                >
-                  <td
-                    @click="
-                      handleClickProductionName(
-                        suggestedProduction.production_id,
-                      )
-                    "
-                  >
-                    {{ suggestedProduction.name }}
-                  </td>
-                </tr>
-              </tbody>
-              <tbody v-else>
-                <tr>
-                  <td>
-                    <loading-spinner />
-                  </td>
-                </tr>
-              </tbody>
-            </table>
-          </section>
         </section>
         <section v-else>
           <b-form-group :label="$t('productions.production_name')">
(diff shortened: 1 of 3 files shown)


pages/manage/productions/create.vue
```diff
@@ -28,44 +28,14 @@
           <label for="production-name">{{
             $t('productions.production_name')
           }}</label>
-          <b-input
-            id="production-name"
+          <vue-typeahead-bootstrap
```
simon-debruijn, line 31: can you give a selected color that matches the udb colors? Instead of the default blue, the $udb-blue for example

bertramakers (approved):

yarn.lock
```diff
@@ -11814,6 +11819,19 @@ vue-template-es2015-compiler@^1.6.0, vue-template-es2015-compiler@^1.9.0:
...
+  resolved "https://registry.yarnpkg.com/vue-typeahead-bootstrap/-/vue-typeahead-bootstrap-2.5.1.tgz#b0adfd082148b0687a0b930aea94fe4f9df38dd6"
+  integrity sha512-KVWFDih0A+qhZTW4BVgsCKj1UYDjfKufRPTQQO4vjpPnymGAtYXBeP7Yx53nOvZumDPpU1zoGUZlAPa4gBFVjg==
+  dependencies:
+    resize-observer-polyfill "^1.5.0"
+    vue "^2.5.17"
+
+vue@^2.5.17:
+  version "2.6.12"
+  resolved "https://registry.yarnpkg.com/vue/-/vue-2.6.12.tgz#f5ebd4fa6bd2869403e29a896aed4904456c9123"
+  integrity sha512-uhmLFETqPPNyuLLbsKz6ioJ4q7AZHzD8ZVFNATNyICSZouqP2Sz0rotWQC8UNBF6VGSCs5abnKJoStA6JbCbfg==
```
bertramakers, line 11833: It's a bit weird that we're suddenly including another version of Vue in our lock file (we also have 2.6.11 just below this one), but I'm assuming that's just how yarn works 🤷‍♂️

simon-debruijn (commented):
@geoffreydhuyvetters can you add a screenshot of how it looks now?

pages/manage/productions/create.vue
geoffreydhuyvetters: Screenshot 2020-08-27 at 12 33 03

it's not blue, it's light grey. I think it's neutral enough to work here. (screenshot is the selected state)

yarn.lock
geoffreydhuyvetters: It's because it's defined as a dependency in vue-typeahead-bootstrap. 

A peerDependency would be better, but we don't have any control over it

simon-debruijn (approved):

pages/manage/productions/create.vue
bertramakers: I agree that it looks fine and works, but I just wanted to add that the current autocomplete fields in the AngularJS app look like this:

Screenshot 2020-08-27 at 13 45 41

I don't think this is necessarily that much better, but it might be helpful to already replicate that look so that when we add more and more Vue pages or re-build ones from the AngularJS app, we don't have two types of autocomplete fields across the app.

yarn.lock
bertramakers: Yeah, I figured it was not something we had control over. Thanks for the explanation though!

PR #40, opened 2020-08-28
Feature/iii 3434 - Styling optimizations of productions frontend

simon-debruijn:
### Changed

- Styling optimizations of productions frontend

---

Screenshots:
Schermafbeelding 2020-08-31 om 09 36 46
Schermafbeelding 2020-08-31 om 09 37 03
Schermafbeelding 2020-08-31 om 09 37 27
Schermafbeelding 2020-08-31 om 09 37 42
Schermafbeelding 2020-08-31 om 09 37 57

Ticket: https://jira.uitdatabank.be/browse/III-3434


components/productions/index/events.vue
```diff
@@ -98,6 +110,10 @@
         type: Boolean,
         default: false,
       },
+      canEnableDeleteButton: {
```
geoffreydhuyvetters, line 113: can't we compute this from selectedEventIds?

components/productions/index/productions.vue
```diff
@@ -56,13 +57,17 @@
 </script>
 
 <style lang="scss">
-  .table-productions {
+  .list-productions {
```
geoffreydhuyvetters, line 60: I would try to avoid prefixing this with the semantic element, .productions-container maybe?

bertramakers: Or make the `style` `scoped` and just call it `.productions`?

simon-debruijn: @bertramakers If you make the component scoped, styling from parent components will be excluded. Currently the production component does need certain styling from higher up.

@geoffreydhuyvetters productions-container already exists in index, but maybe I can rename the one in index

bertramakers: @simon-debruijn  Why would the parent component, in this case, need to style the child component? Wouldn't it be better to make sure each component has its own styles? Global styles like Bootstrap etc keep working even if you add scoped CSS. You can even add a mix of scoped and non-scoped CSS to a component (https://vue-loader.vuejs.org/guide/scoped-css.html#mixing-local-and-global-styles), but I think the default should be to use `scoped` as much as possible to avoid leaking CSS throughout all our components.

assets/styles/global.scss
```diff
@@ -51,10 +51,34 @@ h1 {
...
+a.link {
+  font-weight: 400;
+  color: $udb-blue;
+
+  &:hover {
+    text-decoration: underline;
+  }
+}
+
+.panel {
```
bertramakers, line 76: Feels like something that should go in a separate `panel` component, which would also fit the idea to create a component library and use Storybook.

pages/manage/productions/index.vue
```diff
@@ -215,53 +224,23 @@
...
+      h2 {
+        font-size: 1rem;
+        font-weight: 700;
       }
     }
 
-    a {
-      font-weight: 400;
-      color: $udb-blue;
+    .panel-footer {
```
bertramakers, line 240: Same suggestions as for `panel` to create a separate component for this.

components/productions/index/delete-modal.vue
```diff
@@ -3,15 +3,18 @@
     id="deleteModal"
     :ok-title="$t('productions.delete')"
     :cancel-title="$t('productions.cancel')"
+    cancel-variant="outline-secondary"
```
bertramakers, line 6: The `cancel-variant` will probably be `outline-secondary` in 99% of our cases. So in line with `panel` and `panel-footer`, I'd suggest to add a custom `modal` component. It can just use `b-modal` internally and (for now) always set `cancel-variant` to `outline-secondary`, as well as set `hide-header`, and pass any other props through to `b-modal`.

pages/manage/productions/create.vue
```diff
@@ -254,63 +254,69 @@
 </script>
 
 <style lang="scss">
-  label {
-    font-size: 1rem;
-    font-weight: 700;
-  }
+  .productions-create-page {
+    .create-title {
```
bertramakers, line 258: If you put this in a `scoped` style block, you wouldn't have to add the `productions-create-page` class just for this selector.

pages/manage/productions/index.vue
```diff
@@ -215,53 +224,23 @@
...
-
-      tr {
-        display: inline-flex;
-        width: 100%;
-      }
-
-      th {
-        width: 100%;
-      }
+  .productions-index-page {
```
bertramakers, line 227: Same suggestion, you can just make this `style` block `scoped` instead of wrapping all the rules inside a single container class.

components/productions/index/productions.vue
simon-debruijn: Will be implemented in next PR

assets/styles/global.scss
simon-debruijn: Will be implemented in next PR

pages/manage/productions/index.vue
simon-debruijn: Will be implemented in next PR

components/productions/index/delete-modal.vue
simon-debruijn: Will be implemented in next PR

pages/manage/productions/create.vue
simon-debruijn: Will be implemented in next PR

pages/manage/productions/index.vue
simon-debruijn: Will be implemented in next PR

simon-debruijn:
@bertramakers @geoffreydhuyvetters I created 2 new tickets for the remaining comments about scoped styles and reusable components:
[III-3456](https://jira.uitdatabank.be/browse/III-3456)
[III-3457](https://jira.uitdatabank.be/browse/III-3457)

bertramakers:
Maybe not in scope, but when I add a new event to an existing production but the event gets added to the top of the list. When I then refresh the page, or just switch to another production and back in the list on the left, the new event is at the bottom of the list.
I'd change the front-end to also add the event to the end of the list when adding a new event.

bertramakers:
The submit and cancel buttons for adding a new event use the success and danger variants:
Screenshot 2020-08-31 at 11 57 41

But in other places we use the primary and secondary variants for such forms:
Screenshot 2020-08-31 at 11 58 18

bertramakers:
The overview looks a bit crowded with all the borders, especially because the content of the panel sticks against the sides:
Screenshot 2020-08-31 at 11 55 56

I know this was based on the dashboard, but there it makes sense that everything sticks to the sides and there's borders around the content and pagination:
Screenshot 2020-08-31 at 11 55 46

However because the overview of productions is not a single list and there's some spacing around elements, I'd suggest to to with the approach we used for the search pages:

- Don't use `panel`
- Don't put the pagination in a `panel-footer`

Screenshot 2020-08-31 at 11 55 31

Quick mockup as an example:

Screenshot 2020-08-31 at 12 04 23

Going one step further, we could also remove the info box and put that info as a placeholder in the search input field instead (a bit shortened). But that might need approval from @corneelwille.

For example:

Screenshot 2020-08-31 at 12 08 02

bertramakers:
The selected production has a red highlight that's very similar to the `danger` variant we use for info boxes and buttons. It might be better to use a (light) blue for that instead.

Screenshot 2020-08-31 at 12 10 33

bertramakers:
Suggestion: Maybe we should make the checkboxes next to the events disabled when the "Toevoegen" form is shown? Since you cannot click the delete button then because it's not visible.

bertramakers:
Nit: The title "Evenementen in ..." jumps up and down when opening/closing the "Toevoegen" form:

Aug-31-2020 12-54-54

Can we give this a fixed line-height that's big enough so it stays the same even when the buttons on that line are hidden?

corneelwille:
I really like your suggestions, they'll definitely contribute to a cleaner & more user friendly ui

**About the alert info**
At first I wasn't a huge fan of changing the alert-info into a placeholder for 2 reasons:
- consistency with other pages (of the legacy app)
- making it a placeholder means we'll have to drop the second sentence of the copy 'Laat het veld leeg om alle producties op te vragen'

But holding on to solutions of the past (alert-info) is a bad reason if a new solution is more logic. Also, the second sentence won't be that important (since the ui will only be used by heavy users / admins). 

**About the background color of selected items**
I also prefer the "brand" red (#f3453f), since using a light blue would also be confusing with the color of alert-infos (in the legacy app)

components/productions/index/events.vue
```diff
@@ -1,70 +1,78 @@
...
+        <b-button
+          variant="danger"
+          @click="handleClickCancelAddEventToProduction"
+        >
+          <fa icon="times" />
+          {{ $t('productions.cancel') }}
+        </b-button>
+      </div>
+      <ul class="list-group">
+        <event
```
bertramakers, line 59: The root element of `event` is `section`. But semantically each list item should be a `li`.

components/productions/index/productions.vue
```diff
@@ -1,29 +1,30 @@
...
-      >
-        <td>{{ production.name }}</td>
-      </tr>
-    </tbody>
-    <tbody v-else>
+  <section class="productions-container" aria-label="List of productions">
+    <h2>{{ $t('productions.production') }}</h2>
+    <template v-if="!isLoading && productions.length > 0">
+      <ul class="list-group">
+        <a
```
bertramakers, line 6: Semantically, each list item should be a `li`

bertramakers: (You can just wrap the `a` in a `li` of course)

bertramakers:
> Also, the second sentence won't be that important (since the ui will only be used by heavy users / admins).

It's also the default behaviour which you get when you go to the Productions page from the sidebar, so it feels kind of obvious?

simon-debruijn:
After changes:
Schermafbeelding 2020-08-31 om 17 00 43
Schermafbeelding 2020-08-31 om 17 01 02

bertramakers:
@simon-debruijn I think some feedback wasn't processed yet:

- The pagination should be underneath the list of productions, as confirmed by @corneelwille https://cultuurnet.slack.com/archives/C019V99CW84/p1598875843057000?thread_ts=1598873869.033300&cid=C019V99CW84
- The info text box can be removed and replaced with a `placeholder` inside the search field if I interpret https://github.com/cultuurnet/udb3-frontend/pull/40#issuecomment-683725351 correctly

Also some last small comments.

The "Evenementen in ..." title now doesn't jump around anymore, but those titles and buttons are now not aligned. The lists on the left and right are also not aligned anymore (with ~1px difference).

Screenshot 2020-08-31 at 18 04 33

I was able to align them in my web inspector by putting the `align-items: center` back, giving the `h2`'s a padding of `0.5rem 0`, and setting the `margin-bottom` of the left `h2` to `0.5rem` instead of `1.1rem`. If you want I can commit that fix myself.

Screenshot 2020-08-31 at 18 08 27

The disabled state for selected checkboxes is still light blue. That feels a bit weird if the normal state is dark gray for selected checkboxes. Can you change the disabled state to a lighter gray instead?

Screenshot 2020-08-31 at 18 03 02

Screenshot 2020-08-31 at 18 03 06

Lastly, the "Annuleren" button should have a `#fff` background color. Only on hover should it be `#e6e6e6`. The borders also change on hover. Best to check an example in the AngularJS app. You can find an example on http://localhost:3000/search and then clicking on the "Geavanceerd" link.

bertramakers (approved):
Looks perfect now 👌 Good job @simon-debruijn , it took some iterations but the end result looks and works very good!

PR #42, opened 2020-09-01
Setup storybook

geoffreydhuyvetters:
### Added

- storybook config and setup
- udb3-button component and story

---

Ticket: https://jira.uitdatabank.be/browse/III-3460


simon-debruijn (approved):

nuxt.config.js
```diff
@@ -92,6 +92,9 @@ export default {
   styleResources: {
     scss: ['./assets/styles/_colors.scss'],
   },
+  storybook: {
+    stories: ['~/udb3-ui/**/*.stories.js'],
```
bertramakers, line 96: If we prefix the ui library name, component names, etc with `udb3`, how hard would it be to rename those later if we ever decide to use this in other projects and it should become more generic? I do understand the usefulness of a prefix though. Just trying to estimate if it would be okay to go with something as specific as `udb3` right now or if we should think for a moment about a more generic prefix. But that will also be hard. Do you think it would be doable to rename this later if we have 100+ components? I genuinely don't know.

bertramakers: (One specific place we might want to re-use these is projectaanvraag for example, where external parties register projects and get an API key etc.)

udb3-ui/udb3-button.stories.js
```diff
@@ -0,0 +1,30 @@
+import Udb3Button, { ButtonVariant } from './udb3-button.vue';
```
bertramakers, line 1: Just as an FYI, Storybook also supports the `MDX` format for stories as an alternative to JS. Might be a bit easier to write stories, but I have no strong preference if you both prefer pure JS. Just throwing it out there so we can decide now on a format before we have a lot of stories to convert to one or the other. https://storybook.js.org/docs/vue/api/mdx

bertramakers (approved):
Setup itself looks good to me! 👍 
No mayor comments that have to block a merge unless you yourself think they require changes.

udb3-ui/udb3-button.stories.js
simon-debruijn: Javascript is life though ❤️ But in all seriousness, also could be an option 🙂

bertramakers: Haha 😃 It seems to require a bit less boilerplate code at first glance, but maybe the best way to decide would be to rewrite these same stories in MDX and compare them.

nuxt.config.js
simon-debruijn: just `Button` or `button` wont be possible. But we could use `s-button` for storybook-button for example. It's shorter and similar to b-button from bootstrap

simon-debruijn: and for the library we could also go for something more general `s-ui` or `sb-ui` or `sb-components` are options or nothing to do with storybook, maybe `custom-components` and then `c-button` `c-modal` and so on?

PR #46, opened 2020-09-03
Feature/iii 3457 modal component

simon-debruijn:
### Added

- added pub-modal-content component
- added pub-modal-question component
- added pub-modal-content story
- added pub-modal-question story

### Changed

- changed modals to use new pub-modal components

---


publiq-ui/pub-modal.stories.mdx
```diff
@@ -0,0 +1,100 @@
...
+  variant: {
+    control: {
+      type: false,
+    },
+  },
+};
+
+**
+This component is a wrapper around the [b-modal](https://bootstrap-vue.org/docs/components/modal) component from bootstrap-vue.
+You can use all of the defined methods and props from the bootstrap component.
```
bertramakers, line 45: IMHO this allows way too much freedom and defeats the purpose of wrapping it with our own component? For example you can pass a lot of options that influence the styling of the modal: https://bootstrap-vue.org/docs/components/modal

It also makes it hard to decouple from this implementation later if we ever want to do a redesign, maybe using another component library underneath. I know that's an "if" at this point, but Corneel is open to that idea (but in the long run).

bertramakers:
The styling of the header at the top of the modal is not consistent with that in the AngularJS app.

Screenshot 2020-09-07 at 18 25 17
Screenshot 2020-09-07 at 18 25 37

publiq-ui/pub-modal.stories.mdx
geoffreydhuyvetters: I just don't want to end up in a situation where we need to keep our props in sync with the b-bootstrap ones. But let's discuss it in a meeting later, see the frontend channel

geoffreydhuyvetters:
Can we make the status of this PR a bit clearer? I only review when it's out of draft (if not asked specifically)

publiq-ui/pub-modal.stories.mdx
geoffreydhuyvetters: Maybe there is a way to do it backwards? ignore style-specific props but pass all the other ones?

publiq-ui/pub-modal.vue
```diff
@@ -0,0 +1,80 @@
...
+    title-class="title"
+    modal-class="modal"
+    content-class="content"
+    body-class="body"
+    :visible="visible"
+    :scrollable="false"
+    :hide-footer="isContentVariant"
+    :hide-header="isQuestionVariant"
+    :size="isContentVariant ? 'xl' : 'md'"
+    v-on="$listeners"
```
bertramakers, line 16: We should probably follow the same principle for listeners as for props: Only expose/pass those that we actually use.

publiq-ui/pub-modal.vue
```diff
@@ -0,0 +1,80 @@
...
+    :cancel-title="cancelTitle"
+    title-class="title"
+    modal-class="modal"
+    content-class="content"
+    body-class="body"
+    :visible="visible"
+    :scrollable="false"
+    :hide-footer="isContentVariant"
+    :hide-header="isQuestionVariant"
+    :size="isContentVariant ? 'xl' : 'md'"
```
bertramakers, line 15:
```
:hide-footer="isContentVariant"
:hide-header="isQuestionVariant"
:size="isContentVariant ? 'xl' : 'md'"
```

With all of these differences, I'd consider having 2 separate components like `pub-modal` and `pub-confirmation-modal` or something like that.

Also because the following props are not relevant for the content variant:
```
:ok-title="confirmTitle"
:cancel-title="cancelTitle"
```

There's no clear threshold when it's best to split up, but my gut feeling says these modals are quite distinct.

When we also manually add the listeners as I suggested previously, we'll also have listeners that are not relevant for one variant or the other.

publiq-ui/pub-modal.vue
```diff
@@ -0,0 +1,80 @@
...
+    :ok-title="confirmTitle"
+    :cancel-title="cancelTitle"
+    title-class="title"
+    modal-class="modal"
+    content-class="content"
+    body-class="body"
+    :visible="visible"
+    :scrollable="false"
+    :hide-footer="isContentVariant"
+    :hide-header="isQuestionVariant"
```
bertramakers, line 14: Should this actually be hidden? See the screenshots in my comment here https://github.com/cultuurnet/udb3-frontend/pull/46#issuecomment-688422917

The AngularJS example does have a header with a title.

simon-debruijn: I looked at another example 
Schermafbeelding 2020-09-09 om 09 39 36

I'll hide-header it as a prop we can pass to pub-modal

publiq-ui/pub-modal.vue
simon-debruijn: I was wondering though, with listeners, what is the harm in exposing them all? If we expose the ones we want it is true in storybook they'll be documented better

publiq-ui/pub-modal.vue
simon-debruijn: They're both modals though, if these two are already too different for them to be variants, won't we want to split up every variant of components? except maybe the buttons and the alerts, since their only difference is the color.
I get that there are some differences in the props inside pub-modal, but visually I think they are the same component with slight differences.

bertramakers: _Personally_ I'd draw the line when the difference is not just styling/colors. Once we start displaying/hiding certain parts and have props that only apply to one specific variant, it usually quickly results in a lot of `if/else` switches and confusing documentation. For example, I wouldn't be surprised if at some point we have to introduce an `alert` variant for modals, which only shows the `ok` button but not the cancel button. Then you need to add some more checks to show the footer but only the ok button etc and the code will become very unclear with all of these conditions in one component I believe. But curious what @geoffreydhuyvetters or @brampauwelyn think of this.

publiq-ui/pub-modal.vue
bertramakers: Maybe we can show/hide the header depending on the value of the `title` prop instead? If it's empty hide it, if it has a value show it?

publiq-ui/pub-modal.vue
bertramakers: Yeah I was thinking mostly in having a definite list of props/listeners we have "support" so we don't have to go look at what listeners are registered on each usage of a component if we ever move away from Bootstrap. Maybe it's a bit overkill though. To be honest I didn't even think of this difference with props yet because in React event listeners are also just props (at least in functional components). So I (incorrectly assumed) we were also in agreement about event listeners when we said we would only allow specific props we actually need.

bertramakers: It would be nice to see the events in Storybook though. E.g. when you open a modal, close a modal, click a button, etc. You could just play around with it and see what events are being emitted instead of having to go look for them in the Bootstrap docs.

publiq-ui/pub-modal.vue
simon-debruijn: Also a possibility.

publiq-ui/pub-modal.vue
simon-debruijn: @geoffreydhuyvetters @bertramakers It is ok for me if we also only emit specific events. it will be better documented in storybook. We'll only have a bit more work maintaining them (maybe). But when potentially switching away from bootstrap it will be clearer which events we use and need. Ok for everyone? If so @bertramakers can add this rule to the documentation

publiq-ui/pub-modal.vue
brampauwelyn: IMO I think the modal component should only contain the black overlay + the dialog box + event to close the modal.
I wouldn't add too much markup (such as a title) in the component itself. 
In this case you're always free to add all the elements which are necessary for a specific case.

publiq-ui/pub-modal.vue
bertramakers: I think @geoffreydhuyvetters is actually working on another project today, so if we both agree I propose we do it like this and document it indeed, and we can still discuss it next week if he sees any problems with it we might have missed.

publiq-ui/pub-modal.vue
bertramakers: Under normal circumstances I'd agree, but sadly we (have to) use Bootstrap components underneath right now to replicate the look and feel of the AngularJS app in the Vue app, and the Bootstrap modal component is quite bloated in that aspect.

> \<b-modal\>, by default, has an OK and Cancel buttons in the footer. These buttons can be customized by setting various props on the component. You can customize the size of the buttons, disable buttons, hide the Cancel button (i.e. ok-only), choose a variant (e.g. danger for a red OK button) using the ok-variant and cancel-variant props, and provide custom button content using the ok-title and cancel-title props, or using the named slots modal-ok and modal-cancel.

https://bootstrap-vue.org/docs/components/modal

So it seems hard to get a "blank" modal unless we build it ourselves with Bootstrap classes to replicate the look of the AngularJS app.

bertramakers: @geoffreydhuyvetters we (Simon, Bram, I) discussed this in a call and after weighing the options the consencus was that multiple components would probably be best in this case. So `pub-content-modal`, `pub-question-modal`, ... (for example, exact naming wasn't discussed).

Only downside is that the two variants right now share the same styling, so we'd need to copy that to the two components. One possible solution would be to put those in a separate `pub-modal.scss` file and import it as `scoped` in the components like so: `<style src="./pub-modal.scss" scoped>`. We tested that and it worked. But because it's only 4 selectors and very little rules, I think for now we'll just copy it in 2 places if I remember correctly. (@simon-debruijn can you maybe confirm this last part?)

Other things that came up:
- Listeners: As with props we're not going to emit all bootstrap events, just those we're interested in. This prevents breaking changes if a bootstrap event name changes for example, then we only have to fix it in one place.
- Constants for e.g. variants on other components like `pub-button`: We briefly discussed the upsides and downsides of this approach with constants. While there are some good advantages like having a clear list of possible values and only having to change the value in one place if it would ever change in Bootstrap, we felt it was very cumbersome to have to import the object with the constants, then create a computed value for it so we can access it in the template, and then finally use it in the template. Because of this we decided to go with "hardcoded" values in the template after all (so `variant="danger"` instead of `variant="PubButtonVariant.DANGER"`. We think we can achieve the same advantages as the constants by a) not passing the variant value directly to the bootstrap component but having a mapping in-between in our custom component, to maintain compatibility when a value changes, and b) by having a documented list of values in Storybook (has more or less the same discoverability as seeing the constants in Storybook)

I'll also document these conclusions in the docs in Storybook later on.

simon-debruijn: I'll implement the `<style src="./pub-modal.scss" scoped>` for pub-modal now

bertramakers (approved):
LGTM! 👍

PR #47, opened 2020-09-03
Feature/iii 3457 heading 1 component

simon-debruijn:
### Added

- heading 1 component

---


pages/manage/productions/create.vue
```diff
@@ -1,7 +1,9 @@
 <template>
   <div class="wrapper">
     <section class="container-fluid productions-create-page">
-      <h1 class="title create-title">{{ $t('productions.create') }}</h1>
+      <pub-h1 :variant="heading1Variant.TITLE" class="create-title">{{
```
bertramakers, line 4: I don't think we need a variant for this. So far the only place where it's not used is the start page, but the `h1` there is actually hidden. It's used as a replacement text for screen readers that do not see the UiTdatabank logo in the SVG. So the styling doesn't matter there. So I'd suggest to always add the `border-bottom` by default and not have a variant for that.

bertramakers: See the style that hides the h1 on the start page here: https://github.com/cultuurnet/udb3-frontend/pull/47/files#diff-8df541b53e4cb5088c52e8d2c73e2b09L484

(We could even let that stay a normal `h1` since it's not visible anyway...)

pages/manage/productions/create.vue
```diff
@@ -111,6 +115,9 @@
       locale() {
         return this.$i18n.locale;
       },
+      heading1Variant() {
+        return Heading1Variant;
+      },
```
bertramakers, line 120: If we were to keep this variant, would we need to put the enum object in a computed object? Can't we just access it directly in the template if we import it? 🤔 @geoffreydhuyvetters do you maybe know more about the technical details here?

simon-debruijn: a constant imported in the script tag doesn't seem to be usable in the template unless you expose it via a method, data or computed or the last and maybe best option is to set them as part of the vue object by setting `this.HeadingVariant = HeadingVariant` in the `created` method

bertramakers: This all feels quite complex for using a simple constant in a prop. Not sure if there's a better solution though, except for not using a constant in the first place and just using the hardcoded string 😕

In any case if we don't need the variant here anyway we can leave it out altogether and check what @geoffreydhuyvetters thinks once he's feeling better about how to approach these enums in other components.

publiq-ui/pub-h1.stories.mdx
```diff
@@ -31,20 +21,8 @@ export const commonArgTypes = {
...
-  </Story>
-</Canvas>
-
 ```javascript
-<pub-h1 variant="Heading1Variant.TITLE">Heading 1</pub-h1>
+<pub-h1>Heading 1</pub-h1>
 ```
 
-<ArgsTable story="Title" />
+<ArgsTable story="Default" />
```
bertramakers, line 28: We already show the `ArgsTable` on line 22 https://github.com/cultuurnet/udb3-frontend/pull/47/commits/8ae6bded0d324224fc12fb7140059b09025903d1#diff-d5175fa21352de12428f109d27859cfeR22

bertramakers: But maybe that first one should be removed instead? As it's now above the example codeblock.

pages/manage/productions/index.vue
```diff
@@ -56,7 +56,7 @@
   import DeleteModal from '@/components/productions/index/delete-modal';
   import { parseId as parseEventId } from '@/functions/events';
   import PubWrapper from '@/publiq-ui/pub-wrapper';
-  import pubH1, { Heading1Variant } from '@/publiq-ui/pub-h1';
+  import pubH1 from '@/publiq-ui/pub-h1';
```
bertramakers, line 59: Nit: `pubH1` should be capitalized.
```suggestion
  import PubH1 from '@/publiq-ui/pub-h1';
```
If you change it, make sure to change the occurence in `components` as well down below.

bertramakers (approved):

PR #48, opened 2020-09-03
Feature/iii 3457 wrapper component

simon-debruijn:
### Added

- wrapper component

---

Ticket: https://jira.uitdatabank.be/browse/III-3457

PR #49, opened 2020-09-04
Feature/iii 3457 add pub-pagination

simon-debruijn:
### Added

- added pub-pagination component
- added story for pub-pagination

---

### DIFF
--- a/publiq-ui/pub-pagination.stories.mdx
+++ b/publiq-ui/pub-pagination.stories.mdx
@@ -11,16 +11,31 @@ export const Template = (args, { argTypes }) => {
   };
 };
 
-# Variants
+export const commonArgs = {
+  rows: 50,
+  prevText: 'Previous one',
+  nextText: 'Next one',
+};
+
+export const commonArgTypes = {
+  prevText: {
+    defaultValue: "this.$t('pagination.previous')",
+  },
+  nextText: {
+    defaultValue: "this.$t('pagination.next')",
+  },
+};
 
 ## Default
 
 <Canvas>
-  <Story name="Default">{Template.bind({})}</Story>
+  <Story name="Default" args={commonArgs} argTypes={commonArgTypes}>
+    {Template.bind({})}
+  </Story>
 </Canvas>
 
 ```javascript
-<pub-pagination />
+<pub-pagination rows="50" prev-text="Previous one" next-text="Previous one" />
 ```
 
 <ArgsTable story="Default" />
--- a/publiq-ui/pub-pagination.stories.mdx
+++ b/publiq-ui/pub-pagination.stories.mdx
@@ -1,0 +1,26 @@
+import { Meta, Story, Canvas, ArgsTable } from '@storybook/addon-docs/blocks';
+import PubPagination from './pub-pagination.vue';
+
+<Meta title="pub-pagination" component={PubPagination} />
+
+export const Template = (args, { argTypes }) => {
+  return {
+    props: Object.keys(argTypes),
+    components: { PubPagination },
+    template: '<pub-pagination v-bind="$props" />',
+  };
+};
+
+# Variants
+
+## Default
+
+<Canvas>
+  <Story name="Default">{Template.bind({})}</Story>
+</Canvas>
+
+```javascript
+<pub-pagination />
+```
+
+<ArgsTable story="Default" />
--- a/publiq-ui/pub-pagination.stories.mdx
+++ b/publiq-ui/pub-pagination.stories.mdx
@@ -19,10 +19,14 @@ export const commonArgs = {
 
 export const commonArgTypes = {
   prevText: {
-    defaultValue: "this.$t('pagination.previous')",
+    table: {
+      defaultValue: { summary: '$t(pagination.previous)' },
+    },
   },
   nextText: {
-    defaultValue: "this.$t('pagination.next')",
+    table: {
+      defaultValue: { summary: '$t(pagination.next)' },
+    },
   },
 };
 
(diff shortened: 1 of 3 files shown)


publiq-ui/pub-pagination.stories.mdx
```diff
@@ -0,0 +1,41 @@
...
+
+export const commonArgs = {
+  rows: 50,
+  prevText: 'Previous one',
+  nextText: 'Next one',
+};
+
+export const commonArgTypes = {
+  prevText: {
+    defaultValue: "this.$t('pagination.previous')",
```
simon-debruijn, line 22: @bertramakers these default values don't seem to come through. Although it seems to be following the [docs](https://storybook.js.org/docs/react/api/argtypes)

simon-debruijn: it is meant just as a string, because I don't think we have the ability to use the $t function here, correct me if I'm wrong

bertramakers: Indeed you can't use Nuxt functions here since Storybook is not running in Nuxt. It's just importing our ui components. It's okay to use hardcoded (untranslated) strings here IMHO though. It's purely technical documentation.

publiq-ui/pub-pagination.vue
```diff
@@ -41,24 +41,24 @@
   .pagination {
     justify-content: center;
 
-    ::v-deep .page-item.active .page-link {
+    /deep/ .page-item.active .page-link {
       z-index: 0;
     }
 
-    ::v-deep .prev-btn .page-link {
+    /deep/ .prev-btn .page-link {
```
bertramakers, line 60: Is this just a personal preference or is there a functional/technical difference?

bertramakers: I'm okay with either but we should pick one and stick with it if there's no difference.

publiq-ui/pub-pagination.stories.mdx
```diff
@@ -0,0 +1,26 @@
...
+
+export const Template = (args, { argTypes }) => {
+  return {
+    props: Object.keys(argTypes),
+    components: { PubPagination },
+    template: '<pub-pagination v-bind="$props" />',
+  };
+};
+
+# Variants
```
bertramakers: As mentioned in #48, there's no reason to mention Variants here if there are none 😃 Not every components will have variants.

publiq-ui/pub-pagination.stories.mdx
simon-debruijn: Yes that is also my intention, but the defaultValue doesn't work in the docs part in storybook

publiq-ui/pub-pagination.vue
simon-debruijn: I was planning to do a separate refactor for al occurrence into `/deep/`. it does exactly the same but my IDE picks up better on the syntax

publiq-ui/pub-pagination.vue
```diff
@@ -3,10 +3,12 @@
     v-model="currentPage"
     :total-rows="rows"
```
bertramakers, line 4: Only noticed this now, but maybe `total` would be better instead of `rows`? Or `total-rows`. Because `rows` can easily be misinterpreted as the amount of visible rows instead. I was confused myself at first when I saw the prop in Storybook.

publiq-ui/pub-pagination.stories.mdx
bertramakers: Additionally, you're passing a value already in `commonArgs` so it's normal the default values aren't used.

publiq-ui/pub-pagination.stories.mdx
bertramakers: Ah looks like it was fixed in a more recent commit already 👍

publiq-ui/pub-pagination.vue
bertramakers: Get a better IDE maybe haha 😉 
But okay for me to replace them 👍

publiq-ui/pub-pagination.stories.mdx
bertramakers: If I set
```
export const commonArgs = {
  rows: 50,
};

export const commonArgTypes = {
  prevText: {
    defaultValue: "Previous",
  },
  nextText: {
    defaultValue: "Next",
  },
};
```

I see:
Screenshot 2020-09-04 at 10 56 07

Maybe you're confusing the default value of the control, and the default value of the prop on the component itself? The latter one should be changed in the component itself.

simon-debruijn: Ok I thought I could override the default value coming from the vue component to be a string we decided, because I would have liked to change the long string of `function() { this.$t...`

simon-debruijn: Ah it is possible! with `table: {
      defaultValue: { summary: 'Eigen default' },
    }`

bertramakers: It would be better to have a default that's not reliant on a Nuxt plugin (`nuxt-i18n`) of course. Not sure how to handle this. Maybe also something to discuss with @geoffreydhuyvetters once he feels better.

bertramakers: > Ah it is possible! with `table: {
> 
>       defaultValue: { summary: 'Eigen default' },
> 
>     }`

But wouldn't it be weird to show some kind of "random" default value in Storybook while it's not actually the default? 🤔

bertramakers: Ah I see you set `$t(...)` as the default now to clean it up visually. Still it raises the questions whether it makes sense to access `nuxt-i18n` in a component library. See for example https://stackoverflow.com/a/59218742

bertramakers (approved):
Let's merge this already and discuss how to deal with localization in the ui library next week with @geoffreydhuyvetters 👍

PR #51, opened 2020-09-04
Feature/iii 3457 panel

simon-debruijn:
### Added

- added pub-panel component
- added pub-panel-footer component

---


bertramakers:
Maybe in this case we could remove the separate `pub-panel-footer.stories.mdx` and put an extra "with footer" story in `pub-panel.stories.mdx`? Since there's no real use case for using `pub-panel-footer` outside of `pub-panel`.

bertramakers:
Code changes themselves look good 👍

simon-debruijn:
> Maybe in this case we could remove the separate `pub-panel-footer.stories.mdx` and put an extra "with footer" story in `pub-panel.stories.mdx`? Since there's no real use case for using `pub-panel-footer` outside of `pub-panel`.

done

bertramakers (approved):

PR #53, opened 2020-09-04
Feature/iii 3457 main component

simon-debruijn:
### Added

- added main component
- added main component story

---


bertramakers:
IMHO `pub-main` feels a bit unclear as a component name. Can you think of a more descriptive name?

simon-debruijn:
> IMHO `pub-main` feels a bit unclear as a component name. Can you think of a more descriptive name?
> 
> **Edit**: Do we even need a reusable component for this? It'll probably only be used in one place. I'd rather not put every possible div/class in this library, only components we will actually re-use in multiple places.

True, I'll delete this PR then?

bertramakers:
I'll close it 😉

PR #56, opened 2020-09-07
Feature/iii 3457 list component

simon-debruijn:
### Added

- pub-list component
- pub-list-item component
- pub-list story
- pub-list-item story

---


bertramakers (approved):
Perfect, no remarks! 👌 
Would be cool to also add a story to `pub-panel` of an example with a list inside since that is a common use case I believe, but that can also be done later / in another PR.

PR #57, opened 2020-09-07
Feature/iii 3457 checkbox

simon-debruijn:
### Added

- pub-checkbox component
- pub-checkbox story

---

### DIFF
--- a/publiq-ui/pub-checkbox.stories.mdx
+++ b/publiq-ui/pub-checkbox.stories.mdx
@@ -1,0 +1,24 @@
+import { Meta, Story, Canvas, ArgsTable } from '@storybook/addon-docs/blocks';
+import PubCheckbox from './pub-checkbox.vue';
+
+<Meta title="pub-checkbox" component={PubCheckbox} />
+
+export const Template = (args, { argTypes }) => {
+  return {
+    props: Object.keys(argTypes),
+    components: { PubCheckbox },
+    template: '<pub-checkbox v-bind="$props">Click me</pub-checkbox>',
+  };
+};
+
+## Default
+
+<Canvas>
+  <Story name="Default">{Template.bind({})}</Story>
+</Canvas>
+
+```javascript
+<pub-checkbox>Click me</pub-checkbox>
+```
+
+<ArgsTable story="Default" />
--- a/publiq-ui/pub-checkbox.vue
+++ b/publiq-ui/pub-checkbox.vue
@@ -22,7 +22,7 @@
     },
     methods: {
       handleInput() {
-        this.$emit('input');
+        this.$emit('toggle');
       },
     },
   };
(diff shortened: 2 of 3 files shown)


publiq-ui/pub-checkbox.vue
```diff
@@ -0,0 +1,29 @@
...
+        default: false,
+      },
+      isDisabled: {
+        type: Boolean,
+        default: false,
+      },
+    },
+    methods: {
+      handleInput() {
+        this.$emit('input');
```
bertramakers, line 25: Is the name `input` a Vue convention, a Bootstrap convention, or another kind of convention for input changes? If it's a Bootstrap convention I'm not sure we should re-use it. The Bootstrap components should be the implementation we use underneath, but should not define the public API of our custom components. That way we can easily switch later if we do a redesign and maybe switch to e.g. Material Design components.

bertramakers: Other question: shouldn't we pass some kind of event object to `this.$emit('input');` containing the state of the checkbox? Or how would a listener know that state?

simon-debruijn: The way we do it right now is that the parent component knows which checkbox is checked and passes it on to each checkbox by the isChecked prop. The input handler is used to toggle the known boolean state of that checkbox. We emit this event because we keep the state of the checkbox not in the checkbox itself but in the parent components

bertramakers: Okay thanks, that answers my second question. My first question still stands 😊

simon-debruijn: I think the `input` handler is not necessarily a vue convention, in this case a bootstrap specific event I think. But what would be a better alternative then. To rename the emitted event to something more specific like `select`?

bertramakers: `toggle` perhaps? Or `check`? Although the latter might indicate it's `checked` while it seems like the event is also emitted when unchecking. So that's why I'd go with `toggle` personally if you agree.

simon-debruijn: done

bertramakers (approved):
Thanks, looks great! 👍

publiq-ui/pub-checkbox.stories.mdx
```diff
@@ -0,0 +1,24 @@
+import { Meta, Story, Canvas, ArgsTable } from '@storybook/addon-docs/blocks';
+import PubCheckbox from './pub-checkbox.vue';
+
+<Meta title="pub-checkbox" component={PubCheckbox} />
```
bertramakers, line 4: Oh, don't forget to put this in the `Components` group now that that PR was merged.
So:
```suggestion
<Meta title="Components/pub-checkbox" component={PubCheckbox} />
```

PR #59, opened 2020-09-08
Feature/iii 3465 Add similarityScore on events

geoffreydhuyvetters:
### Added

- Add similarity score on events create page

Ticket: https://jira.uitdatabank.be/browse/III-3465

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -12,6 +12,8 @@
 
 ### Fixed
 
-- ***
+-
+
+---
 
 Ticket: https://jira.uitdatabank.be/browse/...
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -12,8 +12,6 @@
 
 ### Fixed
 
--
-
----
+- ***
 
 Ticket: https://jira.uitdatabank.be/browse/...
--- a/api/productions.js
+++ b/api/productions.js
@@ -93,7 +93,7 @@ export const getSuggestedEvents = (
   });
 
   if (response.status === 204 || response.status === 404) {
-    return [];
+    return { events: [], similarity: 0 };
   }
 
   return await response.json();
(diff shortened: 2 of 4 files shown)


functions/getMockedEventsSet.js
```diff
@@ -407,7 +407,9 @@ const getMockedEventsSet = () => {
     suggestedEventsWithOneProduction,
   ];
   const index = Math.floor(Math.random() * events.length);
-  return events[index];
+  const similarity = Math.random();
```
simon-debruijn, line 410: I know this is just dummy bet shouldn't it be a percentage like `Math.floor((Math.random() * 100) + 1)`

simon-debruijn (approved):

.github/pull_request_template.md
```diff
@@ -12,8 +12,6 @@
 
 ### Fixed
 
--
-
----
+- ***
```
bertramakers, line 15: I'm a bit confused, can you explain what this change is for? The way I see it you removed the horizontal line between the change log and the extra notes (ticket etc), and added `***` in the first bullet point of `Fixed`?

api/productions.js
```diff
@@ -96,8 +96,7 @@ export const getSuggestedEvents = (
     return [];
```
bertramakers, line 96: We should probably also update this to return an object now, right?

bertramakers (changes requested):
I think the error handling in `getSuggestedEvents` will need to be changed now.

api/productions.js
geoffreydhuyvetters: True

.github/pull_request_template.md
geoffreydhuyvetters: Hmm, this was done by prettier. Probably because it's an empty list

geoffreydhuyvetters: I'll revert it

bertramakers (approved):
Thanks for the quick changes @geoffreydhuyvetters, looks good to me now! ✅

PR #61, opened 2020-09-08
Use textarea trick for html entities

geoffreydhuyvetters:
### Added

- use textarea trick for HTML entities

---

Ticket: https://jira.uitdatabank.be/browse/III-3472

### DIFF
--- a/components/productions/create/event.vue
+++ b/components/productions/create/event.vue
@@ -62,6 +62,8 @@
         return this.$i18n.locale;
       },
       truncatedDescription() {
+        // Convert HTML entities like &nbsp; in the description by putting it in a textarea and reading out its value.
+        // See https://stackoverflow.com/a/7394787 for more info.
         const dummyInput = document.createElement('textarea');
         dummyInput.innerHTML = stripHTML(this.description).result;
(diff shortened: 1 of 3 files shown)


simon-debruijn (approved):

bertramakers:
Doesn't this keep the HTML tags inside the content?

The answer you linked in the ticket mentions this: https://stackoverflow.com/questions/7394748/whats-the-right-way-to-decode-a-string-that-has-special-html-entities-in-it/7394787#7394787

> The advantage of using this code is that tags are also preserved.

The JSFiddle also seems to show this.

The goal was to strip the HTML tags out of the content since it's not really readable right now.

bertramakers (changes requested):
Marking as "requesting changes" because I don't think this achieves the requested goal.

geoffreydhuyvetters:
@bertramakers I don't see any HTML in the actual examples, only HTML entities. Which I'm replacing with this technique.

bertramakers (approved):
I've added a comment to clarify the `textarea` trick. Looks good to me otherwise! 👍

PR #62, opened 2020-09-08
Feature/iii 3459 - show error message when a event is being added that already exists

simon-debruijn:
### Added

- add error message when a event is being added that already exists

---

Ticket: https://jira.uitdatabank.be/browse/III-3459

### DIFF
--- a/components/productions/index/events.vue
+++ b/components/productions/index/events.vue
@@ -55,14 +55,13 @@
           {{ $t('productions.overview.cancel') }}
         </b-button>
       </div>
-      <b-alert
+      <pub-alert
         v-for="(errorMessage, index) in errorMessages"
         :key="index"
-        variant="danger"
+        :variant="alertVariant.DANGER"
         :show="errorMessages.length > 0"
-        dismissible
-        >{{ errorMessage }}</b-alert
-      >
+        >{{ errorMessage }}
+      </pub-alert>
       <pub-panel>
         <pub-list>
           <event
@@ -92,6 +91,7 @@
   import { parseId } from '@/functions/events';
   import PubPanel from '@/publiq-ui/pub-panel';
   import PubList from '@/publiq-ui/pub-list';
+  import PubAlert, { AlertVariant } from '@/publiq-ui/pub-alert';
 
   export default {
     components: {
@@ -99,6 +99,7 @@
       Event,
       PubPanel,
       PubList,
+      PubAlert,
     },
     props: {
       events: {
@@ -143,6 +144,9 @@
       },
       hasErrorMessages() {
         return this.errorMessages.length > 0;
+      },
+      alertVariant() {
+        return AlertVariant;
       },
     },
     watch: {
--- a/components/productions/index/events.vue
+++ b/components/productions/index/events.vue
@@ -55,6 +55,14 @@
           {{ $t('productions.overview.cancel') }}
         </b-button>
       </div>
+      <b-alert
+        v-for="(errorMessage, index) in errorMessages"
+        :key="index"
+        variant="danger"
+        :show="errorMessages.length > 0"
+        dismissible
+        >{{ errorMessage }}</b-alert
+      >
       <pub-panel>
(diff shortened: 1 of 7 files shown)


simon-debruijn:
There's possibly going to be a new jira ticket where we add translations for all cases when showing an error.

components/productions/index/events.vue
```diff
@@ -55,6 +55,14 @@
...
         </b-button>
       </div>
+      <b-alert
+        v-for="(errorMessage, index) in errorMessages"
+        :key="index"
+        variant="danger"
+        :show="errorMessages.length > 0"
+        dismissible
+        >{{ errorMessage }}</b-alert
+      >
```
bertramakers, line 65: Seems like a good candidate for a component in the ui library by the way?
If this PR is urgent it doesn't have to block it though.
Just wanted to point out that probably every usage of a `b-...` component should be wrapped at some point?

pages/manage/productions/index.vue
```diff
@@ -141,30 +143,45 @@
       },
       async handleAddEventToProduction(eventId) {
         this.isAddingEventToProduction = true;
-        this.hasAddingEventToProductionError = false;
-        const res = await this.$api.productions.addEventById(
+
+        const foundEvent = this.events.find(
+          (event) => parseEventId(event['@id']) === eventId,
+        );
```
bertramakers, line 149: This approach feels a bit convoluted to me. If I'm not mistaken you already get an error message back from the API, but it's just no translated yet so you try to guess the error by seeing if the event is already part of the production. But there could be other errors. For example the event can not exist, the event can already be part of another production, ... And in that case you do show the untranslated error message anyway. So I would simplify this to just show the untranslated error message, and deal with the localization of error messages another time. It's a more widespread problem than this specific place.

simon-debruijn: we don't get an error message back from the API for this case, so maybe that is the fix needed rather than a frontend fix. 
When I add an event 'Hobbit 1' to the production 'The Hobbit' while it is already there, I get a 204 instead of a 400. Whereas for 'it already belongs to another production' it does work. And then the code without my addition should work I think

components/productions/index/events.vue
simon-debruijn: done

pages/manage/productions/index.vue
bertramakers: Okay let's fix that on the backend then 👍 We still need some changes from this PR to actually show the messages from the backend right? Maybe remove this extra logic already and we'll fix it on the backend asap.

bertramakers: I've added a ticket for the backend change https://jira.uitdatabank.be/browse/III-3481

publiq-ui/pub-alert.vue
```diff
@@ -0,0 +1,26 @@
...
+    SUCCESS: 'success',
+  };
+
+  export default {
+    props: {
+      variant: {
+        type: String,
+        default: AlertVariant.INFO,
+      },
+      show: {
```
bertramakers, line 20: Nit: In the modal PR we use `visible`. I know it's just copied from the Bootstrap prop name, but on our own components we could of course name it whatever we like and use `visible` or `show` consistently. Not blocking though, I leave it up to you to decide what you think is best / worth it.

bertramakers (approved):

publiq-ui/pub-alert.stories.mdx
```diff
@@ -0,0 +1,76 @@
+import { Meta, Story, Canvas, ArgsTable } from '@storybook/addon-docs/blocks';
+import PubAlert, { AlertVariant } from './pub-alert.vue';
+
+<Meta title="pub-alert" component={PubAlert} />
```
bertramakers, line 4: Remember to put this in the `Components` group now since the PR with the grouping was merged.
So:
```suggestion
<Meta title="Components/pub-alert" component={PubAlert} />
```

PR #64, opened 2020-09-09
Feature/iii 3457 refactor variants

simon-debruijn:
### Added

- pub-button components where needed

### Changed

- use strings instead of Variant enums in templates
- stories of pub-button and pub-alert

---

### DIFF
--- a/components/giftbox.vue
+++ b/components/giftbox.vue
@@ -65,9 +65,9 @@
               :href="selectedFeature.callToAction"
               target="_blank"
             >
-              <b-button class="btn-call-to-action" variant="primary">
+              <pub-button class="btn-call-to-action" variant="primary">
                 {{ selectedFeature.callToActionLabel }}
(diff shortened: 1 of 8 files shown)


components/giftbox.vue
```diff
@@ -12,7 +12,7 @@
     <pub-modal-content
       :visible="isModalVisible"
       :title="$t('giftbox.new_features')"
-      @hidden="handleHiddenModal"
+      @hide="handleHiddenModal"
```
bertramakers, line 15: Nit: Should be `handleHideModal` according to the convention the, no? 😉 Okay for me though, it's clear enough 👍

bertramakers (approved):

bertramakers:
Nice work! 👌

components/giftbox.vue
simon-debruijn: true!

PR #65, opened 2020-09-09
III-3457 rename pub-h1 and pub-wrapper into pub-page and pub-page-title

simon-debruijn:
### Changed

- name of pub-h1 to pub-page-title
- name of pub-wrapper to pub-page

---


bertramakers (approved):
I like the idea of `pub-page-title` instead of `pub-h1` a lot 👍 
The fact that it's a `h1` underneath is just an implementation detail. `pub-page-title` is much clearer at a glance imho. 👌

PR #66, opened 2020-09-10
Feature/iii 3457 pub-typeahead

simon-debruijn:
### Added

- pub-typeahead component
- pub-typeahead story
- pub-label component
- pub-label story
- pub-input component
- pub-input story
- pub-content-title component
- pub-content-title story

### Changed

- other components to use pub-typeahead
- other components to use pub-label
- other components to use pub-input

---


publiq-ui/pub-input.vue
```diff
@@ -0,0 +1,53 @@
+<template>
+  <div class="input-container">
+    <pub-label v-if="label" id="form-input">{{ label }}</pub-label>
+    <b-form-input
+      id="form-input"
```
jdrieghe, line 5: Will this create issues when we render this component multiple times on the page?

publiq-ui/pub-input.vue
```diff
@@ -0,0 +1,53 @@
+<template>
+  <div class="input-container">
+    <pub-label v-if="label" id="form-input">{{ label }}</pub-label>
```
bertramakers, line 3: I think this needs a `for=` to link it to the input field (so you can click it to focus on the field, so screen readers know what field it's related to, ...)

publiq-ui/pub-input.vue
```diff
@@ -0,0 +1,53 @@
...
+    align-items: center;
+    line-height: 1rem;
+  }
+
+  label {
+    margin-right: 1rem;
+  }
+
+  #form-input {
+    max-width: 43rem;
```
bertramakers, line 51: I'm wondering if it wouldn't be better to leave this out and set max-widths where we use this (on the container) so we can make it different per use case?

publiq-ui/pub-label.stories.mdx
```diff
@@ -0,0 +1,34 @@
...
+    args={{ id: 'custom-input' }}
+    argTypes={{ id: { control: { type: false } } }}
+  >
+    {Template.bind({})}
+  </Story>
+</Canvas>
+
+```javascript
+<div>
+  <pub-label id="custom-input">Label</pub-label>
```
bertramakers, line 29: Nit: The label and input field stick together in the story. It's good that they don't have a pre-set margin, but maybe we can make this a bit prettier on the eyes by adding a `<br />` so the input is on a newline in this story specifically.

publiq-ui/pub-typeahead.vue
```diff
@@ -0,0 +1,67 @@
...
+      handleHit(value) {
+        this.$emit('hit', value);
+      },
+    },
+  };
+</script>
+
+<style lang="scss" scoped>
+  .typeahead {
+    max-width: 43rem;
```
bertramakers, line 65: Here is this magic `43rem` again 😄 I'm curious where it's coming from.

publiq-ui/pub-input.vue
```diff
@@ -0,0 +1,53 @@
...
+        default: '',
+      },
+      placeholder: {
+        type: String,
+        default: '',
+      },
+    },
+    methods: {
+      handleInput() {
+        this.$emit('input', this.value);
```
bertramakers, line 32: Nit: Maybe `change` would be clearer? (Like we changed the `input` on checkbox to `toggle`)

publiq-ui/pub-input.vue
simon-debruijn: the id prop is passing it to for="" on the actual label, first I was passing a prop 'for' but there was complaining because of the naming

bertramakers: I don't seem to be able to focus on the text field in the `pub-input` story by clicking on the label though.

bertramakers: Ah, I think it's actually caused by what @jdrieghe mentioned, that we re-use the same id in multiple places now.
Because the `for` is indeed there in the HTML

Screenshot 2020-09-15 at 14 36 44

simon-debruijn: I could just create a uuid to be sure, will need to add a new dependency though

bertramakers: How about we add a `name` prop to the `pub-input` component and make that required? Then we can use that for the `id` and `for`? A `uuid` feels a bit... random 😄 But I can see where you're coming from, you probably want to avoid having to set some kind of id/name manually since we probably don't need to identify the field from the outside of the component?

simon-debruijn: Yes, since the only link is to match the label with the input, it seemed appropriate  😉 But your idea is also a good option I think. There will be no way to be sure if the given name isn't already used elsewhere though.

bertramakers: > There will be no way to be sure if the given name isn't already used elsewhere though.

True. Maybe for now, we can set an id based on the `this._uid` that exists in each component (but prefixed for example)? See https://stackoverflow.com/a/34957052 

Then we don't have to set it manually since we have no use case for that right now, but the "random" id is still a bit "predictable" in a sense. If we do need the ability to set a custom id later, we can always add that as a prop then?

simon-debruijn: Cool to know that each component has access to its uid! 🤙  I'll do it using the component's idea then

jdrieghe: I think the uid is a good solution for now. I would be very careful with using a `name` property in the context of an input element. Those have an entirely different meaning in html.

bertramakers: After @simon-debruijn changed it to use `this._uid` we read that it's not supposed to be used actually, it's considered a "private" property 😕 The alternative is a uuid, but that would introduce a dependency on another package/function. Maybe the `name` prop but with another name, like `id`? Makes more sense maybe as it would set that as the HTML id then.

publiq-ui/pub-input.vue
simon-debruijn: Isn't input logical for pub-input? In case of the searchbox, where it's being used it would be `handleChangeSearch` instead of `handleInputSearch`

bertramakers: Fair enough 👍

publiq-ui/pub-input.stories.mdx
```diff
@@ -25,6 +26,7 @@ export const Template = (args, { argTypes }) => {
   </Story>
 </Canvas>
 
+
```
bertramakers, line 29: Nit: Unnecessary double newline
```suggestion
```

components/productions/index/search.vue
```diff
@@ -1,6 +1,7 @@
 <template>
   <pub-input
     class="searchbox"
+    name="productions-overview-search"
```
bertramakers, line 4: I think @jdrieghe had a comment about `name` being a bad choice of name for this prop because it means something else in HTML. Maybe `id`?

bertramakers: See https://github.com/cultuurnet/udb3-frontend/pull/66#discussion_r488866894

simon-debruijn: `id` like `name` or `class` also already has a meaning in html

components/productions/index/search.vue
```diff
@@ -29,5 +29,6 @@
 <style lang="scss" scoped>
   .searchbox {
     margin-bottom: 1rem;
+    max-width: 43rem;
```
bertramakers, line 33: Thanks for moving this, but it's still used in multiple places. Maybe we should move it to a variable to prevent a wide range of max-widths in the long run. That way the number can also get a descriptive name to make it seem less arbitrary. (I'm still not sure why `43rem` exactly?)

pages/manage/productions/index.vue
```diff
@@ -233,17 +233,15 @@
...
-        font-size: 1rem;
-        font-weight: 700;
-      }
+  .productions-events-container {
+    display: flex;
+    width: 100%;
+    font-weight: 400;
+    margin-bottom: 1rem;
+
+    /deep/ h2 {
```
bertramakers, line 242: Also seems like something we should move to a component?

components/productions/index/search.vue
bertramakers: Yes but `id` is would be the same as what it becomes in the HTML. It's more predictable.

simon-debruijn: but maybe `id` matches better with the use of id in plain html, while `name` in html is used for a different reason than identifying solely

bertramakers: Exactly 😄

pages/manage/productions/index.vue
simon-debruijn: h2 is styled differently over multiple pages, so I don't really see how we can make it a component efficiently

bertramakers: Can you give some examples? Maybe we can come up with something that looks mostly the same but is consistent. Or if they're completely different in some cases, maybe those cases should become something else than a h2?
We also don't have to call the component `h2`. It could have a more specific name as we have for h1's (`pub-page-title`).

components/productions/index/search.vue
simon-debruijn: done

pages/manage/productions/index.vue
bertramakers: Looking at productions alone, I see an `h2` on the create page that's `1.2rem` while this one is `1rem`. Maybe we should make both `1.2rem` to start with? As an `h2` it feels like it should be bigger than normal text imho?

bertramakers: Another `h2` I found, but in the AngularJS app, are the titles on the "Invoeren" form. Those also happen to be `1.2rem` (and `font-weight: 700`). They're also capitalized and have border-bottom and a number in front of them, but those are extras we could add in this specific case without changing the `h2` itself I think.

Screenshot 2020-09-16 at 09 39 47

simon-debruijn: Schermafbeelding 2020-09-16 om 09 34 25
Schermafbeelding 2020-09-16 om 09 34 39
Schermafbeelding 2020-09-16 om 09 41 15
Schermafbeelding 2020-09-16 om 09 41 23
Schermafbeelding 2020-09-16 om 09 41 29
Schermafbeelding 2020-09-16 om 09 41 35

simon-debruijn: So maybe all occurrences to `1.2rem` and `font-weight: 700 ` and then border, padding and such as styling in the specific pages? The capitalization we could add as a variant or a prop capitalized?

bertramakers: Yep that sounds good to me 👍 
Maybe leave the capitalization out, for now, we can figure that out once we have to rebuild the "Invoeren" page.

components/productions/index/search.vue
simon-debruijn: This is also a relevant discussion for the margins, maybe we should have global variables like `width sm, md, lg` and `margin sm, md, lg` in our global.scss?

simon-debruijn: It can be more values and differently named

pages/manage/productions/index.vue
bertramakers: Sorry I meant that we should probably also move that to a separate component in another PR, not in this one 😄 Because now this whole PR is blocked on a potential discussion how h2's should look and be called. Let's create an issue for that like @geoffreydhuyvetters suggested and leave this out of this PR. My comment was more of a suggestion to do it later anyway but I could've worded that a bit better.

bertramakers (approved):

PR #67, opened 2020-09-10
Feature/iii 3457 link component

simon-debruijn:
### Added

- pub-link component
- pub-link story

---


pages/manage/productions/index.vue
```diff
@@ -4,9 +4,9 @@
       <pub-page-title>
         {{ $t('menu.productions') }}
         <small>
-          <nuxt-link class="link" to="productions/create">{{
-            $t('productions.overview.create')
-          }}</nuxt-link>
+          <nuxt-link to="productions/create">
+            <pub-link>{{ $t('productions.overview.create') }}</pub-link>
+          </nuxt-link>
```
bertramakers, line 9: Wrapping `pub-link` in `nuxt-link` results in a double `a` tag in the final HTML:

Screenshot 2020-09-11 at 10 12 05

I think you could fix it with an approach like this:
```
<pub-link :as="nuxt-link" to="productions/create">{{ $t('productions.overview.create') }}</pub-link>
```

To make that `:as` work, see https://medium.com/bitbond/vue-load-dynamic-components-based-on-a-string-variable-52f54860a0d

bertramakers:
Rest of the changes look good 👍

publiq-ui/pub-link.vue
```diff
@@ -1,13 +1,25 @@
...
-  <a :href="href" @click="handleClick"><slot /></a>
+  <component :is="as" v-bind="$attrs" v-on="$listeners" @click="handleClick">
+    <slot />
+  </component>
 </template>
 
 <script>
+  import PubButton from './pub-button';
+
+  export const SupportedTypes = ['a', 'nuxt-link', 'pub-button'];
```
bertramakers, line 10: I'd remove this validation because from the ui lib's perspective we "cannot" know the definitive list of what other types of link components we'd need to support. So if we'd extract this UI lib it will hinder us.

For example `nuxt-link` is an implementation detail specific to Udb3 / Nuxt. But another project could have `other-framework-link` so to speak. `nuxt-link` is also technically not known by the UI lib and Storybook. We could import it because we happen to have all the code in a single repository, but that's not always going to be the case.

publiq-ui/pub-link.vue
```diff
@@ -1,13 +1,25 @@
 <template>
-  <a :href="href" @click="handleClick"><slot /></a>
+  <component :is="as" v-bind="$attrs" v-on="$listeners" @click="handleClick">
```
bertramakers, line 2: I'm curious, is `v-bind="$attrs"` different from `v-bind="$props"` and when should we use which one?
Also good idea to bind the `listeners` in this case 👍

publiq-ui/pub-link.vue
simon-debruijn: Yes but we have to import components like PubButton, so we can't just accept PubAlert, because it isn't imported in pub-link

simon-debruijn: So that was the idea behind the validation

publiq-ui/pub-link.vue
simon-debruijn: `v-bind="$props"` only passes the props defined in the pub-link component, `v-bind="$attrs"` passes all the props given, no matter if they've been defined as props in pub-link

publiq-ui/pub-link.vue
simon-debruijn: I mean I could import the whole components directory but that doesn't really make sense to me

bertramakers: Ah, I see. That makes sense! I'm just wondering now if there's no other way to achieve that, for example by registering the component we put in the `:as` globally in the Vue app itself? 🤔

bertramakers: For example: https://github.com/nuxt/nuxt.js/issues/421

publiq-ui/pub-link.vue
bertramakers: Okay thanks for explaining!

publiq-ui/pub-link.vue
simon-debruijn: I'm not totally sure what you mean. Registering all components as globally known?

bertramakers: No, just `nuxt-link` in the Nuxt app.

bertramakers (approved):

bertramakers (approved):

PR #71, opened 2020-09-14
Feature/iii 3456 radio group

simon-debruijn:
### Added

- pub-radio-group component
- pub-radio-group story

### Changed

- implementation in components for to pub-radio-group

---

Ticket: https://jira.uitdatabank.be/browse/III-3456

### DIFF
--- a/pages/manage/productions/create.vue
+++ b/pages/manage/productions/create.vue
@@ -56,8 +56,7 @@
           <pub-radio-group
             v-model="selectedSuggestedProductionId"
             :group-label="$t('productions.create.production_name')"
-            :values="availableProductionIds"
-            :labels="availableProductionTitles"
+            :items="mappedAvailableProductions"
             name="production"
             class="choose-suggested-production"
           />
@@ -156,11 +155,11 @@
           .filter((event) => event.production)
           .map((events) => events.production);
       },
-      availableProductionIds() {
-        return this.availableProductions.map((production) => production.id);
-      },
-      availableProductionTitles() {
-        return this.availableProductions.map((production) => production.title);
+      mappedAvailableProductions() {
+        return this.availableProductions.map((production) => ({
+          label: production.title,
+          value: production.id,
+        }));
       },
       suggestedProductionNames() {
         return this.suggestedProductions.map((production) => production.name);
--- a/pages/manage/productions/create.vue
+++ b/pages/manage/productions/create.vue
@@ -129,14 +129,6 @@
 
       isLinkingEventsWithProduction: false,
     }),
-    watch: {
-      selectedSuggestedProductionId() {
-        console.log(
-          'selectedSuggestedProductionId',
-          this.selectedSuggestedProductionId,
-        );
-      },
-    },
     computed: {
       locale() {
         return this.$i18n.locale;
--- a/pages/manage/productions/create.vue
+++ b/pages/manage/productions/create.vue
@@ -325,6 +325,7 @@
   .production-name-input {
     position: relative;
     max-width: 43rem;
+    margin-bottom: 1rem;
   }
 
   .container-table-production-name {
--- a/pages/manage/productions/create.vue
+++ b/pages/manage/productions/create.vue
@@ -42,16 +42,6 @@
           />
         </section>
         <section v-else>
-          <!-- <b-form-group :label="$t('productions.create.production_name')">
-            <b-form-radio
-              v-for="production in availableProductions"
-              :key="production.id"
-              v-model="selectedSuggestedProductionId"
-              :value="production.id"
-              name="production"
-              >{{ production.title }}</b-form-radio
-            >
-          </b-form-group> -->
           <pub-radio-group
             v-model="selectedSuggestedProductionId"
             :group-label="$t('productions.create.production_name')"
--- a/pages/manage/productions/create.vue
+++ b/pages/manage/productions/create.vue
@@ -322,11 +322,7 @@
     margin-bottom: 1rem;
   }
 
-    .production-name-input {
-      max-width: 43rem;
-    }
-
-  .production-input {
+  .production-name-input {
     max-width: 43rem;
   }
 
--- a/pages/manage/productions/create.vue
+++ b/pages/manage/productions/create.vue
@@ -42,7 +42,7 @@
           />
         </section>
         <section v-else>
-          <b-form-group :label="$t('productions.create.production_name')">
+          <!-- <b-form-group :label="$t('productions.create.production_name')">
             <b-form-radio
               v-for="production in availableProductions"
               :key="production.id"
@@ -51,7 +51,14 @@
               name="production"
               >{{ production.title }}</b-form-radio
             >
-          </b-form-group>
+          </b-form-group> -->
+          <pub-radio-group
+            v-model="selectedSuggestedProductionId"
+            :group-label="$t('productions.create.production_name')"
+            :values="availableProductionIds"
+            :labels="availableProductionTitles"
+            name="production"
+          />
         </section>
 
         <section class="button-container">
@@ -94,6 +101,7 @@
   import PubPage from '@/publiq-ui/pub-page';
   import PubAlert from '@/publiq-ui/pub-alert';
   import PubButton from '@/publiq-ui/pub-button';
+  import PubRadioGroup from '@/publiq-ui/pub-radio-group';
 
   export default {
     components: {
@@ -104,6 +112,7 @@
       PubPageTitle,
       PubAlert,
       PubButton,
+      PubRadioGroup,
     },
     data: () => ({
       eventSimilarityScore: 0,
@@ -119,6 +128,14 @@
 
       isLinkingEventsWithProduction: false,
     }),
+    watch: {
+      selectedSuggestedProductionId() {
+        console.log(
+          'selectedSuggestedProductionId',
+          this.selectedSuggestedProductionId,
+        );
+      },
+    },
     computed: {
       locale() {
         return this.$i18n.locale;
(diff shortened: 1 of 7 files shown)


publiq-ui/pub-radio-group.vue
```diff
@@ -0,0 +1,60 @@
...
+  <b-form-group :label="groupLabel" class="form-group">
+    <b-form-radio-group v-model="selectedModel" stacked>
+      <b-form-radio
+        v-for="(value, index) in values"
+        :key="value"
+        :value="value"
+        :name="name"
+        class="custom-radio"
+        >{{ labels[index] }}</b-form-radio
+      >
```
jdrieghe, line 11: This formatting looks really weird. Was this suggested by the linter? 🤔

jdrieghe (approved):

publiq-ui/pub-radio-group.vue
bertramakers: It's automatically formatted by prettier. It seems to be correct since the prettier check on Github Actions is green for this PR.

bertramakers:
Nit: The white circle in the selected radio button seems off-center in Firefox.

Screenshot 2020-09-15 at 13 51 35

Due to the contrast between the black and white, the circle also looks quite jagged imho.

In the same browser, it is centered in Bootstrap-Vue:
Screenshot 2020-09-15 at 13 55 26

(https://bootstrap-vue.org/docs/components/form-radio#form-radio-inputs)

Note that we use the system radio button in e.g. the entry form:
Screenshot 2020-09-15 at 13 57 03

Bootstrap 4 itself also uses this: https://getbootstrap.com/docs/4.0/components/forms/#default-stacked

Screenshot 2020-09-15 at 14 00 44

I'm not sure why Bootstrap-Vue in particular deviates from this. Maybe it's easier to just stick to the system radio buttons / checkboxes instead of customizing them and possibly losing UX / accessibility features that we might not think of and/or Bootstrap-Vue has overlooked?

At the very least I would stick to the default color scheme for these components. (Grey border, blue background + white dot if selected, ...). It could be jagging for Windows users though.

pages/manage/productions/create.vue
```diff
@@ -42,16 +42,14 @@
...
-            >
-          </b-form-group>
+          <pub-radio-group
+            v-model="selectedSuggestedProductionId"
+            :group-label="$t('productions.create.production_name')"
+            :values="availableProductionIds"
+            :labels="availableProductionTitles"
+            name="production"
+            class="choose-suggested-production"
+          />
```
bertramakers, line 52: I like how you made it a single component that takes a list of values and labels and emits the selected value instead of forcing us to replicate this logic in every place we show radio buttons 👍

simon-debruijn:
> Nit: The white circle in the selected radio button seems off-center in Firefox.
> 
> Screenshot 2020-09-15 at 13 51 35
> 
> Due to the contrast between the black and white, the circle also looks quite jagged imho.
> 
> In the same browser, it is centered in Bootstrap-Vue:
> Screenshot 2020-09-15 at 13 55 26
> 
> (https://bootstrap-vue.org/docs/components/form-radio#form-radio-inputs)
> 
> Note that we use the system radio button in e.g. the entry form:
> Screenshot 2020-09-15 at 13 57 03
> 
> (Edit: In that place we also don't have bold labels next to the radio buttons, so unless we do make them bold in other places we should probably stick to that style.)
> 
> Bootstrap 4 itself also uses this: https://getbootstrap.com/docs/4.0/components/forms/#default-stacked
> 
> Screenshot 2020-09-15 at 14 00 44
> 
> I'm not sure why Bootstrap-Vue in particular deviates from this. Maybe it's easier to just stick to the system radio buttons / checkboxes instead of customizing them and possibly losing UX / accessibility features that we might not think of and/or Bootstrap-Vue has overlooked?
> 
> At the very least I would stick to the default color scheme for these components. (Grey border, blue background + white dot if selected, ...). It could be jagging for Windows users though.

This styling happens because for the checkbox we have overwritten certain bootstrap defaults to achieve the look of our checkboxes in productions. Because there were no specific variables for checkbox, they apply to several selectable components.
A solution could be to change the default for 'controls' back to the default blue, like bootstrap does it, and just create a variant with specific styling for the checkboxes like in productions.

Schermafbeelding 2020-09-15 om 15 43 19

bertramakers:
Yep, I understand _why_ it looks like that now (except for the white circle being off-center). 😄 But I think we should try to take a step back and look at the bigger picture for a moment. Personally, I'm not sure why we need custom radios/checkboxes to begin with. The only place we had a custom checkboxes in the past is the multi-select on the search pages for events and places. In other places we've always used the system checkboxes (and radios). Recently, we copied the custom styling of checkboxes for productions, but we could just as well have used the regular ones. 

About more variants, it's like @geoffreydhuyvetters said a while back: We should copy the current look and feel, but if we see three different styles of tables for example we should pick one and stick with it. I think it's the same for form elements here. Either we go with custom ones like in the search page (and now productions), or we go with the regular ones (as in all other places at the moment).

The reason I'd personally stick to the regular system ones are:
- More familiar to all users, regardless of their OS
- No behaviour we or bootstrap-vue might forgot to re-implement while re-inventing these elements (e.g. is it tab-able, etc...?).
- The styling of these elements seems to be hard to get right, see below for an example as I just noticed that the white circle in the selected radio is apparently only off-centre in _some_ cases (so probably influenced by cascading styles).

Incorrect in Docs:
Screenshot 2020-09-15 at 15 57 44

Correct in Canvas:
Screenshot 2020-09-15 at 15 57 49

I know this probably feels a lot like nitpicking over a very small detail, I just think it's important to get the fundamental elements like this right from the beginning when building a UI library and get these decisions out of the way moving forward. 😊

bertramakers:
Okay so conclusions after some Slack threads:

- We're going to stick with the system checkboxes / radio buttons for now. Once all (most) pages use these centralized components we can always change them later
- We should discuss these things first in an RFC issue so we don't need to constantly change direction in the actual implementation

simon-debruijn:
> Don't think we should include the label now. We didn't before either and a checkbox can be used in a lot of scenarios (label to the left, label to the right, ...)

The bootstrap checkbox also included a label, won't we be using checkbox with labels on the right in 99% of the time?

publiq-ui/pub-checkbox.vue
```diff
@@ -28,11 +28,6 @@
         type: String,
         default: '',
       },
-      label: {
-        type: String,
-        default: '',
-        required: true,
-      },
```
bertramakers: There's been some confusion I think. My suggestion was to remove the label altogether, and just let this component be a checkbox. Not to remove the prop but keep the `pub-label`.

simon-debruijn: I know this was a unused prop, I commented on your suggestion

pages/manage/productions/create.vue
```diff
@@ -53,16 +53,14 @@
...
-              :value="production.id"
-              name="production"
-              >{{ production.title }}</b-form-radio
-            >
-          </b-form-group>
+          <pub-radio-group
+            v-model="selectedSuggestedProductionId"
+            :group-label="$t('productions.create.production_name')"
+            :values="availableProductionIds"
+            :labels="availableProductionTitles"
```
geoffreydhuyvetters, line 60: Can't we use collection here (array with objects)? 
That would make it more clear when your values and labels don't align.

simon-debruijn: Done

bertramakers:
> The bootstrap checkbox also included a label, won't we be using checkbox with labels on the right in 99% of the time?

Bootstrap doesn't do that, bootstrap-vue does 😉

We could make it optional and put a `v-if` on the `pub-label`? That way it's there for those 99% and we can also leave it out in those 1%?

simon-debruijn:
> > The bootstrap checkbox also included a label, won't we be using checkbox with labels on the right in 99% of the time?
> 
> Bootstrap doesn't do that, bootstrap-vue does 😉
> 
> We could make it optional and put a `v-if` on the `pub-label`? That way it's there for those 99% and we can also leave it out in those 1%?

I have made the change

publiq-ui/pub-checkbox.vue
```diff
@@ -1,12 +1,33 @@
 <template>
-  <b-form-checkbox :checked="checked" :disabled="disabled" @input="handleInput">
-    <slot />
-  </b-form-checkbox>
+  <div class="container-checkbox">
+    <input
+      :id="id"
+      type="checkbox"
+      :name="name"
```
bertramakers, line 6: Do we need both the `id` and `name`?

publiq-ui/pub-radio-group.stories.mdx
```diff
@@ -0,0 +1,46 @@
...
+  </Story>
+</Canvas>
+
+```javascript
+<pub-radio-group
+  v-model="selected"
+  name="city"
+  groupLabel="Select a city:"
+  :labels="['Rome', 'Paris', 'Prague']",
+  :values="['rome', 'paris', 'prague']",
```
bertramakers, line 41: The example in Storybook is broken since the latest changes to use a collection instead of two arrays.

publiq-ui/pub-checkbox.vue
simon-debruijn: They serve different purposes. Traditionally name is something you have to pass to a checkbox input to then later get the selected values by name. But in vue we could leave it, I think

simon-debruijn: I'll leave the name prop, which is optional. If we need it it's there

bertramakers (approved):

PR #72, opened 2020-09-14
feature/III-3491 - add locationName and locationCity information to event cards on create page

simon-debruijn:
### Added

- feature/III-3491 - add locationName and locationCity information to event cards on create page

---

Ticket: https://jira.uitdatabank.be/browse/III-3491

### DIFF
--- a/pages/manage/productions/create.vue
+++ b/pages/manage/productions/create.vue
@@ -14,6 +14,10 @@
             :key="suggestedEvent['@id']"
             :type="getEventType(suggestedEvent.terms)"
             :title="suggestedEvent.name[locale]"
+            :location-name="suggestedEvent.location.name[locale]"
+            :location-city="
+              suggestedEvent.location.address[locale].addressLocality
+            "
             :image-url="suggestedEvent.image"
             :production-name="
               suggestedEvent.production ? suggestedEvent.production.title : ''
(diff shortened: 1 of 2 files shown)


jdrieghe (approved):
👍  looking good

pages/manage/productions/create.vue
```diff
@@ -14,6 +14,10 @@
             :key="suggestedEvent['@id']"
             :type="getEventType(suggestedEvent.terms)"
             :title="suggestedEvent.name[locale]"
+            :location-name="suggestedEvent.location.name[locale]"
```
bertramakers, line 17: It's possible that there's no name for the given `locale`. In that case we should fall back on the `mainLanguage` of the `location` (the location is its own separate entity with its own mainLanguage, separate from the event - it can be that the event was originally entered in Dutch but the location in French).

So something like:
```suggestion
            :location-name="suggestedEvent.location.name[locale] ?? suggestedEvent.location.name[suggestedEvent.location.mainLanguage]"
```

For `title` we should probably do the same, but with the `mainLanguage` of `suggestedEvent`:
```
:title="suggestedEvent.name[locale] ?? suggestedEvent.name[suggestedEvent.mainLanguage]"
```

pages/manage/productions/create.vue
```diff
@@ -14,6 +14,10 @@
...
+            :location-city="
+              suggestedEvent.location.address[locale].addressLocality
```
bertramakers, line 19: (Same comment as for `location-name`)

bertramakers (changes requested):
We need to provide a fallback for missing translations on the event/location

bertramakers (approved):
Looking good now, and much more readable than my quick example 😉 👍

PR #73, opened 2020-09-14
Feature/iii 3478 spinner create page

simon-debruijn:
### Added

- pub-loading-spinner to create page

### Changed

- alertbox to be pub-alert

---

Ticket: https://jira.uitdatabank.be/browse/III-3478

### DIFF
--- a/pages/manage/productions/create.vue
+++ b/pages/manage/productions/create.vue
@@ -5,7 +5,7 @@
       <div v-if="isLoadingEvents">
         <pub-loading-spinner />
       </div>
-      <div v-else-if="!isLoadingEvents && suggestedEvents.length > 0">
+      <div v-else-if="suggestedEvents.length > 0">
         <p>
           <strong>{{ $t('productions.create.suggested_events') }}</strong>
           {{ eventSimilarityScore }}%
--- a/pages/manage/productions/create.vue
+++ b/pages/manage/productions/create.vue
@@ -69,14 +69,15 @@
           </pub-button>
         </section>
       </div>
-      <section v-else class="list-group-item list-group-item-warning">
+      <pub-alert v-else variant="warning" visible>
         {{ $t('productions.create.no_suggested_events_found') }}
-      </section>
+      </pub-alert>
       <pub-alert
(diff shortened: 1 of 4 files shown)


pages/manage/productions/create.vue
```diff
@@ -2,7 +2,10 @@
   <pub-page>
     <section class="container-fluid productions-create-page">
       <pub-page-title>{{ $t('productions.create.title') }}</pub-page-title>
-      <div v-if="suggestedEvents.length > 0">
+      <div v-if="isLoadingEvents">
+        <pub-loading-spinner />
+      </div>
+      <div v-else-if="!isLoadingEvents && suggestedEvents.length > 0">
```
jdrieghe, line 8: Do you still need the `!isLoadingEvents` when you are in the else-if branch?

jdrieghe:
```suggestion
      <div v-else-if="suggestedEvents.length > 0">
```

jdrieghe (approved):
Would remove the `!isLoadingEvents` from the else-if branch before merging

pages/manage/productions/create.vue
bertramakers: I like this article that Tony shared a while back about checks like this: https://dev.to/davidkpiano/no-disabling-a-button-is-not-app-logic-598i

It's a long read (but super interesting and very clear), but the gist is that instead of having multiple boolean flags/checks like this (`isLoadingEvents`, `suggestedEvents.length > 0` which is essentially something like `hasSuggestedEvent`, etc) you define a list of (internal) states that your component can have and transition between them.

For example in this case the state could possibly be one of `loading`, `empty`, `success` (we can argue about the exact names, they're just for illustration).

You can then do checks like `state === 'loading'`, or make it a computed value like `isLoading()` _based on the current state_.

The advantage of this approach is that you can never be in two states at once, as you could with two (or more) boolean flags/checks.

Maybe in this case it's still manageable with boolean flags, although this suggestion already illustrates a bit that it becomes confusing pretty quickly 😉 In any case I think it's something we could consider applying as much as possible.

bertramakers (approved):

PR #75, opened 2020-09-16
Set a width for the table headers

geoffreydhuyvetters:
### Added

- provide width for the table headers

---

Ticket: https://jira.uitdatabank.be/browse/III-3483

### DIFF
--- a/components/productions/index/event.vue
+++ b/components/productions/index/event.vue
@@ -126,4 +126,10 @@
       display: none;
     }
   }
+
+  .table {
+    th {
+      width: 10rem;
+    }
+  }
 </style>
--- a/components/productions/index/event.vue
+++ b/components/productions/index/event.vue
@@ -127,7 +127,7 @@
     }
   }
 
-  .table {
+  table {
     th {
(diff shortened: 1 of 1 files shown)


bertramakers (approved):
Clear and simple! 👍

components/productions/index/event.vue
```diff
@@ -126,4 +126,10 @@
       display: none;
     }
   }
+
+  .table {
```
bertramakers, line 130: Nit: The `.table` class is a bit redundant though, no? 😛 But I know it was already there in the template.

simon-debruijn (approved):

bertramakers:
@geoffreydhuyvetters because the selector was now too broad and the `style` tag is not scoped (didn't know that at first) this will target other `th` elements as well. We discussed yesterday that you would add a class name to the `th` elements instead, but since we agreed a bit later that you would focus on the RFC for primitive components + test setup, I've simply reverted the last commit for now so we can merge this with the `.table` class name and clean it up later.

PR #79, opened 2020-09-18
Make label font-weight consistent

bertramakers:
### Changed

- Changed `font-weight` of `pub-label` component to always be `400` instead of `700` and then resetting it everywhere.

### DIFF
--- a/publiq-ui/pub-checkbox.vue
+++ b/publiq-ui/pub-checkbox.vue
@@ -59,10 +59,6 @@
     cursor: pointer;
   }
 
-  label {
-    font-weight: normal;
-  }
-
   label:hover {
     cursor: pointer;
   }
--- a/publiq-ui/pub-label.vue
+++ b/publiq-ui/pub-label.vue
@@ -17,7 +17,7 @@
 <style lang="scss" scoped>
   label {
     font-size: 1rem;
-    font-weight: 700;
+    font-weight: 400;
     margin-bottom: 0;
   }
 </style>
--- a/publiq-ui/pub-radio-group.vue
+++ b/publiq-ui/pub-radio-group.vue
@@ -76,10 +76,6 @@
     cursor: pointer;
   }
 
-  label {
-    font-weight: normal;
-  }
-
   label:hover {
     cursor: pointer;
   }


simon-debruijn:
Do we also want the typeahead and input labels to be normal font-weight instead of the bold?

Schermafbeelding 2020-09-21 om 09 16 22
Schermafbeelding 2020-09-21 om 09 16 31

bertramakers:
> Do we also want the typeahead and input labels to be normal font-weight instead of the bold?

Looking at the pages where we already use those, there seems to be a lot of bold text already. Personally I wouldn't mind making the labels normal font-size.

Input with bold label:
Screenshot 2020-09-21 at 09 55 26

Input with normal label:
Screenshot 2020-09-21 at 09 55 47

Typeahead with bold label:
Screenshot 2020-09-21 at 09 56 36

Typeahead with normal label:
Screenshot 2020-09-21 at 09 56 14

I'm curious how it would look with bold/normal text when we put multiple form elements together. Maybe we should add a story with multiple input fields, a typehead, a radio group, a checkbox etc and see how it looks & feels together?

simon-debruijn:
> > Do we also want the typeahead and input labels to be normal font-weight instead of the bold?
> 
> Looking at the pages where we already use those, there seems to be a lot of bold text already. Personally I wouldn't mind making the labels normal font-size.
> 
> Input with bold label:
> Screenshot 2020-09-21 at 09 55 26
> 
> Input with normal label:
> Screenshot 2020-09-21 at 09 55 47
> 
> Typeahead with bold label:
> Screenshot 2020-09-21 at 09 56 36
> 
> Typeahead with normal label:
> Screenshot 2020-09-21 at 09 56 14
> 
> I'm curious how it would look with bold/normal text when we put multiple form elements together. Maybe we should add a story with multiple input fields, a typehead, a radio group, a checkbox etc and see how it looks & feels together?

We can definitely try a story with multiple Labels. From my quick research on the internet, labels are almost always bold text, and are preferred to be above the input field, preferred  over a label on the left and then input. But maybe the radio group and checkbox text shouldn't be labels, just a normal span?

bertramakers:
Checkboxes and radio buttons should have a `label` for accessibility though: https://www.w3.org/WAI/tutorials/forms/labels/

bertramakers:
Let's keep it as it is for now and create a story with a combination of these form elements, as well as an RFC if we see the need to make changes then.

Issue #80, opened 2020-09-21
[RFC] [pub-page-title-action]


simon-debruijn:
## Use case

When a title contains a action like 'toevoegen', this component makes it have the correct styling.

## Existing examples in the Legacy UI and/or Vue app

![Schermafbeelding 2020-09-21 om 13 50 35](https://user-images.githubusercontent.com/66943525/93763486-d2d46580-fc11-11ea-81fa-45b1fab35c1b.png)

## Proposed styling

Wrap the content in a `small html tag` and apply `text-transform: lowercase`. 

## Actions

`@click`

bertramakers:
Good idea. Will it use `pub-link` underneath?
Can you also add a code block with an example of how to use it in combination with `pub-page-title`?

simon-debruijn:
That was the thing I wasn't sure about yet. Maybe leave it separate and just have:
`<pub-page-title-action>
      <pub-link as="nuxt-link" to="productions/create">
          toevoegen
      </ pub-link>
</ pub-page-title-action>
`
the 'to' and 'as' props will probably make it unnecessarily complicated if we have to pass them through

bertramakers:
Okay yeah seeing it like that I agree. It would also give some more flexibility, so we could also have a `<pub-page-title-action>` with another component inside with a `@click` to e.g. open a modal or something like that instead of a `pub-link` to another page.

bertramakers:
So it would work like this?

```
<pub-page-title>
    Producties
    <pub-page-title-action>
        <pub-link as="nuxt-link" to="productions/create">toevoegen</ pub-link>
    </pub-page-title-action>
</pub-page-title>
```

simon-debruijn:
> So it would work like this?
> 
> ```
> <pub-page-title>
>     Producties
>     <pub-page-title-action>
>         <pub-link as="nuxt-link" to="productions/create">toevoegen</ pub-link>
>     </pub-page-title-action>
> </pub-page-title>
> ```

Yes, sorry I was struggling with the spacing in markdown for the codeblocks 😋 But that is exactly how it would be

bertramakers:
It's a lot of tags but gives more flexibility about what the action can be (a link, @click, etc) and have that separated from the styling so I'm okay with that trade-off. @geoffreydhuyvetters do you have some input or do we implement it like this?

PR #81, opened 2020-09-21
Feature/iii 3457 make scoped

simon-debruijn:
### Changed

- change style tags to be scoped

---

Ticket: https://jira.uitdatabank.be/browse/III-3456

### DIFF
--- a/components/job/job-logger.vue
+++ b/components/job/job-logger.vue
@@ -238,7 +238,7 @@
   };
 </script>
 
-<style lang="scss">
+<style lang="scss" scoped>
   .list-unstyled {
     padding-left: 0;
     list-style: none;
--- a/components/job/job-logger.vue
+++ b/components/job/job-logger.vue
@@ -295,9 +295,6 @@
     font-size: 0.75em;
     display: block;
   }
-  .badge {
-    background-color: $danger;
-  }
   .udb-job-success {
     color: #5cb85c;
   }
--- a/components/productions/index/productions.vue
+++ b/components/productions/index/productions.vue
@@ -1,5 +1,5 @@
 <template>
-  <section class="productions-container" aria-label="List of productions">
+  <section aria-label="List of productions">
     <h2>{{ $t('productions.overview.production') }}</h2>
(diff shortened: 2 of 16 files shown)


pages/manage/productions/index.vue
```diff
@@ -3,7 +3,7 @@
     <div class="container-fluid">
       <pub-page-title>
         {{ $t('menu.productions') }}
-        <small>
+        <small class="title-action">
```
simon-debruijn, line 6: will become its own component [pub-page-title-action](https://github.com/cultuurnet/udb3-frontend/issues/80)

components/productions/index/productions.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <section class="productions-container" aria-label="List of productions">
+  <section aria-label="List of productions">
```
bertramakers, line 2: Nit: I'm not sure if we need the `aria-label` here by the way, we also have the `h2` right below. So screen readers would read both "List of productions" (in English!) and then "Producties" right afterward. But technically unrelated to this PR.

bertramakers:
The code changes look good to me! I clicked around through the app, focusing on the components/pages that changed, and noticed that the indicator of events to validate is no longer a badge though:

Screenshot 2020-09-22 at 12 52 10

And neither does the unread count of the announcements:
Screenshot 2020-09-22 at 12 54 28

components/job/job-logger.vue
```diff
@@ -238,7 +238,7 @@
   };
 </script>
 
-<style lang="scss">
+<style lang="scss" scoped>
```
bertramakers, line 241: The issue with the badges seems to be because this component got scoped, but also has a rule for `.badge`.
Maybe we should add a component for the badge to fix that?

bertramakers: @simon-debruijn I fixed it myself, was a small change. See the commits below.
I didn't create an RFC for the component because it was pretty simple.

bertramakers (approved):

PR #82, opened 2020-09-21
III-3459 - change variant from 'alert' to 'warning'

simon-debruijn:
'alert' isn't a known variant, thus was a typo

### Changed

- change variant from 'alert' to 'warning'

### Fixed

- pub-alert styling

---

Screenshot:
Schermafbeelding 2020-09-21 om 16 05 35

Ticket: https://jira.uitdatabank.be/browse/III-3459

### DIFF
--- a/publiq-ui/pub-alert.stories.mdx
+++ b/publiq-ui/pub-alert.stories.mdx
@@ -70,10 +70,10 @@ export const commonArgTypes = {
 </Canvas>
 
 ```javascript
-<pub-alert variant="danger">Lorem Ipsum</pub-alert>
+<pub-alert variant="warning">Lorem Ipsum</pub-alert>
 ```
 
-<ArgsTable story="Danger" />
+<ArgsTable story="Warning" />
 
 ## Danger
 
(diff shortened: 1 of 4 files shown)


bertramakers:
Fix itself looks good to me. Do you think we should also add validation to the prop itself to prevent this from happening in the future? https://vuejs.org/v2/guide/components-props.html#Prop-Validation

simon-debruijn:
> https://vuejs.org/v2/guide/components-props.html#Prop-Validation

Yes I think it can only help! And in this case we want this to be restricted, whereas this wasn't the case with pub-link.

bertramakers:
Okay, can you add that as well then? I'll review it again afterward!

simon-debruijn:
> Okay, can you add that as well then? I'll review it again afterward!

done

publiq-ui/pub-alert.stories.mdx
```diff
@@ -70,10 +70,10 @@ export const commonArgTypes = {
 </Canvas>
 
 ```javascript
-<pub-alert variant="danger">Lorem Ipsum</pub-alert>
+<pub-alert variant="warning">Lorem Ipsum</pub-alert>
 ```
 
-<ArgsTable story="Danger" />
+<ArgsTable story="Warning" />
```
bertramakers, line 76: Good spot! 👍

bertramakers (approved):

PR #83, opened 2020-09-21
Feature/iii 3457 pub title

simon-debruijn:
### Added

- pub-title component
- pub-title story

---

### DIFF
--- a/publiq-ui/pub-title.vue
+++ b/publiq-ui/pub-title.vue
@@ -1,0 +1,36 @@
+<template>
+  <component :is="heading"><slot /></component>
+</template>
+
+<script>
+  export default {
+    props: {
+      size: {
+        type: Number,
+        default: 2,
+        required: true,
+        validator(val) {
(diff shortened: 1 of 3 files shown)


publiq-ui/pub-title.vue
```diff
@@ -0,0 +1,39 @@
...
+</template>
+
+<script>
+  export default {
+    props: {
+      size: {
+        type: Number,
+        default: 2,
+        validator(val) {
+          return [1, 2, 3, 4].includes(val);
```
bertramakers, line 12: I'm wondering if it wouldn't be safer to leave out `3` and `4` for now, since we don't have styling for them?
Or if you put them there intentionally, we should probably also add stories for them?

simon-debruijn: No indeed I forgot to remove them from the validator function

bertramakers (approved):

PR #85, opened 2020-09-22
Fix pub-link's color

bertramakers:
### Fixed

- Fixed `pub-link`'s color. Reboot.css was overwriting the color of all `a` tags. By adding a class and making our selector more specific, our selector has precedence.

### DIFF
--- a/publiq-ui/pub-link.vue
+++ b/publiq-ui/pub-link.vue
@@ -1,5 +1,5 @@
 <template>
-  <component :is="as" v-bind="$attrs" v-on="$listeners" @click="handleClick" class="link">
+  <component :is="as" :class="className" v-bind="$attrs" v-on="$listeners" @click="handleClick">
     <slot />
   </component>
 </template>
@@ -12,6 +12,11 @@
         default: 'a',
       },
     },
+    computed: {
+      className() {
+          return this.as === 'a' ? 'link' : '';
+      }
+    },
     methods: {
       handleClick() {
         this.$emit('click');
@@ -21,7 +26,7 @@
 </script>
 
 <style lang="scss" scoped>
-  .link {
+  a.link {
     font-weight: 400;
     color: $udb-blue;
 
--- a/publiq-ui/pub-link.vue
+++ b/publiq-ui/pub-link.vue
@@ -1,5 +1,5 @@
 <template>
-  <component :is="as" :class="className" v-bind="$attrs" v-on="$listeners" @click="handleClick">
+  <component :is="as" :class="pub-link" v-bind="$attrs" v-on="$listeners" @click="handleClick">
     <slot />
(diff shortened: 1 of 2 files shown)


simon-debruijn (approved):

publiq-ui/pub-link.vue
```diff
@@ -21,7 +21,7 @@
 </script>
 
 <style lang="scss" scoped>
-  a {
+  .link {
```
simon-debruijn, line 24: what if I pass something else than a 'a'-tag. Like an Icon or button, it will get the color udb-blue then. Why doesn't the a-tag selector work?

simon-debruijn: maybe a.link as selector?

bertramakers: Ah yes, good spot!

Two reasons we shouldn't use `a` on its own:
- It seems like we still have some issues with bootstrap being included _after_ our own styles. But only in Storybook. We should fix that on its own, but it also indicates that our components' selectors are too generic and can easily be overwritten from the outside. Using a class makes our selectors more specific and harder to accidentally overwrite.
- Vue itself recommends not to use html tags as selectors when using scoped for performance reasons: https://vue-loader.vuejs.org/guide/scoped-css.html#also-keep-in-mind

> Scoped styles do not eliminate the need for classes. Due to the way browsers render various CSS selectors, p { color: red } will be many times slower when scoped (i.e. when combined with an attribute selector). If you use classes or ids instead, such as in .example { color: red }, then you virtually eliminate that performance hit.

I'll see if I can apply the class only if it's an `a`, or go with your suggestion 👍

bertramakers: Here's a screenshot of reboot overwriting our own styles in Storybook (the stylesheets file names are only visible in dev mode btw, on Chromatic it's all inline but also in this order)

Screenshot 2020-09-22 at 14 36 09

bertramakers: Ah, I just realized it's not the order in which the CSS is included or loaded, but the precedence of the selectors. Reboot's `a:not([href]):not([class])` is simply more specific than our own `a` selector which is why it takes precedence.

bertramakers: Adding `href="#"` to the template of the story fixes it. (It was in the example docblock, but not the template of the actual story.) But then on hover I get a similar problem:

Screenshot 2020-09-22 at 14 42 26

bertramakers: @simon-debruijn I've implemented your suggestion + it now only adds the class if the link is an `a` (to avoid any conflicts with another `.link` class outside the component) + added an `href` to the story template.

simon-debruijn (approved):

PR #86, opened 2020-09-22
Testing setup

geoffreydhuyvetters:
### Added

- initial testing setup
- mock API
- create `mountWithEnvironment` helper


simon-debruijn:
Btw: we are using [vue-test-utils](https://vue-test-utils.vuejs.org/api/) and not [testing-library](https://testing-library.com/)

geoffreydhuyvetters:
We struggled to make it work with this setup (especially setting up the mocks)

test/api.js
```diff
@@ -0,0 +1,29 @@
+import events from './data/events';
+import projects from './data/projects';
```
bertramakers, line 2: Nit: Filename should probably be `productions`, not `projects`?

pages/manage/productions/index.test.js
```diff
@@ -0,0 +1,58 @@
...
+
+import { mountWithEnvironment } from '../../../test/mountWithEnvironment';
+
+test('returns full list when entering the page', async () => {
+  const wrapper = mountWithEnvironment(Index);
+
+  await Vue.nextTick();
+
+  const listItems = wrapper.findAll('.productions .list-group-item');
+  expect(listItems).toHaveLength(2);
```
bertramakers, line 12: Could we do an assertion on the expected title(s) of the productions instead, since we mock them anyway? To avoid having a dom selector that could change?

pages/manage/productions/index.test.js
```diff
@@ -0,0 +1,58 @@
...
+});
+
+test('returns a found item when searching with an existing production name', async () => {
+  const wrapper = mountWithEnvironment(Index);
+
+  const searchInput = wrapper.find('#productions-overview-search');
```
bertramakers, line 18: Is there no way to find an input element, checkbox, etc by its label text with the setup we have now? Or if we add another package? I think I remember that was one point we all seemed to agree on, that we wanted to avoid having DOM selectors.

publiq-ui/pub-page-title.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <pub-title size="1" class="page-title"><slot /></pub-title>
+  <pub-title :size="1" class="page-title"><slot /></pub-title>
```
bertramakers, line 2: Curious: Does it need the `:` because it expects a number instead of a string? Or is there another reason?

test/api.js
```diff
@@ -0,0 +1,29 @@
+import events from './data/events';
```
bertramakers, line 1: I like this approach of having a centralized set of test events that we can then access through mocked API methods, should avoid a lot of repetitive data mocking 👍

pages/manage/productions/index.test.js
```diff
@@ -0,0 +1,58 @@
...
+  expect(productionsContainer.text()).toContain(
+    'productions.overview.no_productions',
+  );
+});
+
+test('Enable remove button when selecting event', async () => {
+  const wrapper = mountWithEnvironment(Index);
+
+  await Vue.nextTick();
+  await Vue.nextTick();
```
bertramakers, line 49: I may lack some knowledge about these testing tools, but it's a bit unclear to me why we need two `await Vue.nextTick()`'s here instead of the usual one in other places.

bertramakers (commented):
Overall I like the structure and way we mock test data. 👍  I'm less a fan of using CSS/DOM selectors to find elements in the pages/components since those could change, and focus less on how the user sees the app. I think doing assertions on text and querying on text would also make the tests even more readable. But not sure if that's possible with the tools we have now so I can live with it if there's really no other way.

bertramakers:
Oh, can you also add a Github action to run them for each PR and push to master?

pages/manage/productions/index.test.js
bertramakers: By the way, this package is a lightweight adapter built on top of DOM Testing Library (which we originally meant to use) and @vue/test-utils (which we use now if I'm not mistaken). That might be a solution to have the best of both worlds? https://github.com/testing-library/vue-testing-library

An example in their README:

Screenshot 2020-09-22 at 17 55 12

publiq-ui/pub-page-title.vue
bertramakers: Ah I see we also had to change it in another PR to make it pass a number. Makes sense 👍

pages/manage/productions/index.test.js
bertramakers: An example from vue test utils documentation: https://vue-test-utils.vuejs.org/guides/#getting-started

`expect(wrapper.text()).toContain('Hello world')`

geoffreydhuyvetters:
> Overall I like the structure and way we mock test data. 👍  I'm less a fan of using CSS/DOM selectors to find elements in the pages/components since those could change, and focus less on how the user sees the app. I think doing assertions on text and querying on text would also make the tests even more readable. But not sure if that's possible with the tools we have now so I can live with it if there's really no other way.

I'm afraid not, maybe a safer way is to work with specific data attributes and use as selectors?

pages/manage/productions/index.test.js
geoffreydhuyvetters: We tried that one, but it wasn't clear how to mock the next specific helpers (API, i18n) when rendering

test/api.js
geoffreydhuyvetters: yeah definitely

pages/manage/productions/index.test.js
simon-debruijn: We need to wait for vue to update the ui, so we have to use `Vue.nextTick()`. In this example we noticed it only renders fully after using `Vue.nextTick()` 2x.

bertramakers: Okay thanks for the explanation. I assumed `Vue.nextTick()` waited for the next render yes. Weird that we have to do it twice here but okay, at this point the Vue magic doesn't surprise me anymore 😉 As long as it works consistently and doesn't fail every so often because it needs a _third_ `nextTick` it's fine by me 👍

test/renderWithEnvironment.js
```diff
@@ -1,15 +1,16 @@
...
 
 import $api from './api';
 
 const localVue = createLocalVue();
 localVue.use(BootstrapVue);
 
-export const mountWithEnvironment = (Component, options = {}) =>
-  mount(
+export const renderWithEnvironment = (Component, options = {}) =>
+  render(
```
bertramakers, line 13: Such a small change, yet such a big impact 🎉

components/productions/index/event.vue
```diff
@@ -6,6 +6,7 @@
         :name="name"
         :checked="isSelected"
         :disabled="isDisabled"
+        :aria-label="name"
```
bertramakers, line 9: I think `pub-checkbox` uses `<label>` underneath, so maybe we don't even need this? If we do, maybe we should add it to the component so it sets it automatically for every checkbox based on the `:name` so we don't need to duplicate it everywhere?

components/productions/index/events.vue
```diff
@@ -64,11 +64,12 @@
         >{{ errorMessage }}
       </pub-alert>
       <pub-panel>
-        <pub-list>
+        <pub-list aria-label="events">
```
bertramakers, line 67: Will this be read out loud by screen readers? Should we make it translatable? Probably not that big a deal for an internal feature like productions, but want to make sure we don't discover later that we have a lot of `aria-label` attributes throughout the app that need to become translatable.

bertramakers: (Also for the other aria-label attributes added below that are not based on user input)

pages/manage/productions/index.test.js
```diff
@@ -1,58 +1,62 @@
...
+  expect(noProductions).toBeInTheDocument();
 });
 
 test('Enable remove button when selecting event', async () => {
-  const wrapper = mountWithEnvironment(Index);
+  const checkbox = screen.getByRole('checkbox', { name: 'eee' });
 
-  await Vue.nextTick();
-  await Vue.nextTick();
+  const removeButton = await screen.findByText('productions.overview.delete');
```
bertramakers, line 55: Just to clarify, this `productions.overview.delete` is the actual text that's rendered right? I'm guessing because we have no translations/default messages in the tests? Just want to make sure I understand it right.

simon-debruijn: Yes we mock the $t function

components/productions/index/search.vue
```diff
@@ -24,6 +24,8 @@
         this.$emit('inputSearch', value);
       },
       handleInputSearch(value) {
+        // ugly fix for 'this' binding issue when trying to mock debounce
+        // setting it to 0 on test also doesn't resolve it
```
bertramakers, line 28: Thanks for documenting the reason!

pages/manage/productions/index.test.js
bertramakers: Too bad that it makes the test less readable, but makes sense. 👍

bertramakers:
Weird that Github is not running the actions for this branch. Maybe because of the merge conflict? 🤔

pages/manage/productions/index.test.js
geoffreydhuyvetters: What would you suggest, using the actual translation values? The translation keys are most likely to stay no?

bertramakers: Nah it's fine, it was mostly to make sure I understood it right so I know what "text" to look for when I write tests myself 😄 In another project I worked on (React) we used a translation package that also made it possible to set a default message whenever you called it with a key. So if the key doesn't exist, it would print the default message. We used that in the tests because it wouldn't change either (since the defaults never changed in the code, but would be overwritten in the translations if someone needed to change the exact copy). It's technically more or less the same, just a bit more readable in the tests, that's all. But I don't know if that's even possible with `vue-i18n` / `nuxt-i18n`. So this is definitely okay as well for me! 👍

bertramakers (approved):
I'm approving already because all of my leftover comments are small optimizations, nothing blocking. Assuming the tests are green of course after fixing the merge conflict 😉

simon-debruijn (approved):
✨✨✨ WOW ✨✨✨

PR #87, opened 2020-09-22
Fixes pub title

simon-debruijn:
### Changed

- replaced usage of h2 by pub-title

### Fixed

- props being passed as strings instead of numbers


components/productions/index/events.vue
```diff
@@ -2,13 +2,13 @@
   <section>
     <template v-if="!isLoading">
       <div class="heading-container">
-        <h2>
+        <pub-title class="page-title">
```
bertramakers, line 5: Nit: The `page-title` class name feels a bit confusing because `pub-page-title` is also a component.

bertramakers: It's also not really a page title I think?

components/productions/index/productions.vue
```diff
@@ -1,6 +1,8 @@
 <template>
   <section>
-    <h2>{{ $t('productions.overview.production') }}</h2>
+    <pub-title class="page-title">{{
```
bertramakers, line 3: (Same comment about the class name.)

pages/manage/productions/index.vue
```diff
@@ -253,10 +253,5 @@
     .events-container {
       width: 60%;
     }
-
-    /deep/ h2 {
```
bertramakers, line 257: 🙌

bertramakers (approved):

components/productions/index/events.vue
simon-debruijn: I mean, they're all titles. The ones I replaced are all h2's. But if you have a better name we can still change it.

bertramakers: I meant it's not really a _page_ title 😉 (I was talking about the class name, not the component name) 

I think it's especially confusing with the `pub-page-title` component. Because what would be the difference between `<pub-title class="page-title">` and `<pub-page-title>` for someone going through the code to make a change?

bertramakers: Since you already have `.heading-container`, I guess `.heading` would be okay?

simon-debruijn: ahhh ok, I totally missed your point! Yeah sure I'll change them.

simon-debruijn: done

bertramakers (approved):

PR #93, opened 2020-09-24
Fix layout loading spinner button

simon-debruijn:
### Added

- new variants for pub-spinner

### Changed

- renamed pub-loading-spinner to pub-spinner

### Removed

- styling applied on spinner in button

### Fixed

- layout problem with spinner in button

---

Screenshot problem:
Schermafbeelding 2020-09-24 om 17 44 45 (2)

Screenshot after fix:
Schermafbeelding 2020-09-24 om 17 50 20


publiq-ui/pub-spinner.vue
```diff
@@ -0,0 +1,38 @@
+<template>
+  <div class="text-center d-flex justify-content-center spinner-container">
+    <b-spinner :small="isInButtonVariant" :variant="mappedVariant"></b-spinner>
+  </div>
+</template>
+
+<script>
+  const SpinnerVariant = {
+    onPage: 'primary',
+    inButton: 'light',
```
bertramakers, line 10: Suggestion: Instead of the context (`onPage`, `inButton`) we could make it a prop that says more about the look of the spinner itself? Like `<pub-spinner color="primary">` and `<pub-spinner color="light">`.
Because I imagine we'll have other places we can use spinners than on a page or in a button.
Of course, it would be even better if those colors would be actual design tokens that are the same everywhere we use them.

bertramakers: Also in a `secondary` button, you'd need a blue spinner. So then you end up with something like this with the `onPage` and `inButton` variants:

```
<pub-spinner variant="onPage" />

<pub-button variant="primary">
  <pub-spinner variant="inButton" />
</pub-button>

<pub-button variant="secondary">
  <pub-spinner variant="onPage" />
</pub-button>
```

Compare this to:

```
<pub-spinner variant="primary" />

<pub-button variant="primary">
  <pub-spinner variant="light" />
</pub-button>

<pub-button variant="secondary">
  <pub-spinner variant="primary" />
</pub-button>
```

bertramakers: Additionally, we could add a `loading=true|false` prop to `pub-button` that would automatically show a spinner of the right variant for the button variant? It could also disable it automatically.

So we would only have to do this:
```
Blue button with white spinner:
<pub-button variant="primary" :loading="true"></pub-button>

White button with blue spinner:
<pub-button variant="secondary" :loading="true"></pub-button>
```

simon-debruijn: Not only the color is the difference also the size of the `inButton` is smaller. But what do you think about this approach:
`<pub-spinner size="small" variant="light" :loading="true" />`
`<pub-spinner size="normal" variant="primary" :loading="true" />` 

`size="normal"` and `variant="primary"` would be the defaults if nothing is passed.

simon-debruijn: Yeah btw your approach also can work, but that can be used in combination with my suggestion

bertramakers: Yup good idea to add a size prop as well then. 👍 
The `:loading="true"` was meant for `pub-button` btw.

publiq-ui/pub-spinner.stories.mdx
```diff
@@ -14,36 +14,48 @@ export const Template = (args, { argTypes }) => {
...
+  <Story name="InButton" args={{ loading: true }}>
     {TemplateInButton.bind({})}
   </Story>
 </Canvas>
 
 ```javascript
-<pub-button>
-  <pub-spinner variant="inButton" />
-</pub-button>
+<pub-button :loading="true">Click me</pub-button>
```
bertramakers, line 58: Maybe this should be a story for `pub-button` instead?

bertramakers: And could you also add a story for the sizes?

bertramakers (approved):
Good improvement aside from the bug fix! 👍

PR #97, opened 2020-10-06
Feature/iii 3151 - use different urls for the login page according to the language

simon-debruijn:
### Added

- new route `/login/nl`
- new route `/login/fr`

### Changed

- place of `login.vue` from `pages` to `components`

---

Ticket: https://jira.uitdatabank.be/browse/III-3151

### DIFF
--- a/i18n/fr.json
+++ b/i18n/fr.json
@@ -3,6 +3,8 @@
   "main": {
     "activities_info": "Vous proposez un concert, une représentation, une exposition ou un film? Ou plutôt un événement sportif, une randonnée, une soirée cartes, une kermesse? Ou encore des cours, des ateliers ou des événements? Tout cela trouvera sa place dans l’UiTdatabank.",
     "activities": "activités par an",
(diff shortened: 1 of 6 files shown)


i18n/fr.json
```diff
@@ -3,6 +3,8 @@
   "main": {
     "activities_info": "Vous proposez un concert, une représentation, une exposition ou un film? Ou plutôt un événement sportif, une randonnée, une soirée cartes, une kermesse? Ou encore des cours, des ateliers ou des événements? Tout cela trouvera sa place dans l’UiTdatabank.",
     "activities": "activités par an",
+    "channels_info_link_text": "",
+    "channels_info_link_url": "",
```
bertramakers, line 7: Are these supposed to be empty?

bertramakers (approved):

i18n/fr.json
simon-debruijn: Yes these aren't used on the french page, but there where warnings of missing keys when you pick french

PR #101, opened 2020-10-27
Feature/set up component system

geoffreydhuyvetters:
### Added

- styled-components and theming setup

### Removed

- most of the global styling

---


src/functions/getValueFromTheme.js
```diff
@@ -0,0 +1,8 @@
+import { get } from 'lodash';
+import { theme } from '../components/publiq-ui/theme';
+
+const getValueFromTheme = (path) => {
+  return get(theme, path);
+};
+
+export { getValueFromTheme };
```
geoffreydhuyvetters, line 8: Don't really like the idea of a 'functions' folder, looks super weird to me. I prefer to keep those as close to where we're using them. So in this scenario, I would just keep it in the `publiq-ui` folder, which also makes a lot more sense if we want to distribute it later.

package.json
```diff
@@ -24,13 +24,15 @@
     "bootstrap": "^4.5.3",
     "i18next": "^19.8.3",
     "i18next-browser-languagedetector": "^6.0.1",
+    "lodash-es": "^4.17.15",
```
geoffreydhuyvetters, line 27: We're importing `lodash-es` but using `lodash` in the codebase?

Also, in the Vue version, we ended up using `lodash` because this complicated things with Jest (needing to add specific configuration to transpile `lodash-es` for the tests). There might be a nice approach to handle this, but just a heads up here.

src/styles/global/_custom-bootstrap.scss
```diff
@@ -8,47 +8,13 @@ $light: $white;
...
-    background-color: #e6e6e6;
-    border-color: #adadad;
-  }
-}
-
-//alert overrides
-.alert-info {
-  background-color: #d9edf6;
-  border-color: #bce8ef;
-  color: #3e88ab;
```
geoffreydhuyvetters, line 53: Maybe we could also store these colors in the theme as we're doing with btn-outline-secondary

src/components/publiq-ui/theme.js
```diff
@@ -0,0 +1,40 @@
...
+  colors,
+  components: {
+    button: {
+      borderRadius: 0,
+      paddingX: '0.8rem',
+      paddingY: '0.267rem',
+      secondary: {
+        color: '#333',
+        backgroundColor: colors.white,
+      },
```
geoffreydhuyvetters, line 23: Shouldn't we go for outlineSecondary here?

or create a specific property?

```js
outline: {
  secondary: {

  }
}
```

src/functions/getValueFromTheme.js
geoffreydhuyvetters: This is also not how the theming works, you have to retrieve it from the props in the styled-component.

src/components/publiq-ui/Alert.js
```diff
@@ -8,10 +9,22 @@ const AlertVariants = {
   WARNING: 'warning',
 };
 
+const StyledBootstrapAlert = styled(BootstrapAlert)`
+  &.alert-info {
+    background-color: #d9edf6;
+    border-color: #bce8ef;
+    color: #3e88ab;
+  }
```
bertramakers, line 17: For simple components like `Alert` and `Button` we might be able to just roll our own with the same styles as the Bootstrap components? We're already explicitly setting these colours. All that's left to just roll our own would basically be some `padding`, `margin` and `border-radius`. (But do we even have a border-radius on alerts in the original design?) And then a transition and extra padding for dismissible alerts.

Screenshot 2020-10-27 at 16 37 05

For pagination, modals etc I'd agree to keep re-using Bootstrap components because of their complexity. But in this case if we roll our own that might make it easier to make adjustments later?

simon-debruijn: @bertramakers Your suggestion is to step away from Bootstrap as much as possible?

bertramakers: I would just evaluate it per component. Like in this case, it's only a couple of extra CSS properties and then we don't need the Bootstrap component.

Not saying we have to ditch Bootstrap completely right now, but if we can recreate the same look with little effort maybe it's easier that way so if we ever want to move away from Bootstrap there's less refactoring.

bertramakers: @geoffreydhuyvetters what do you think?

geoffreydhuyvetters: The only reason why I wouldn't do this is to keep the built-in logic Bootstrap provides.

src/components/publiq-ui/Button.js
```diff
@@ -8,12 +10,32 @@ const ButtonVariants = {
   DANGER: 'danger',
 };
 
+const getValue = (path) => (props) =>
+  getValueFromTheme(props, `components.button.${path}`);
```
geoffreydhuyvetters, line 14: I think we can change the function so it becomes

```js
const getValue =
  getValueFromTheme(`components.button`);
```

here

src/components/publiq-ui/Alert.js
bertramakers: For pagination, modal, etc definitely! For `Alert` and `Button`, I think that's relatively simple at first glance. For `Alert` it's the dismiss click handler that needs to hide/remove it, and for `Button` it's the click event that needs to call an injectable handler? Unless I'm oversimplifying it. Just basing this on what I see from reviewing the PRs. So if you both believe it would be much more work than what it's worth it's also okay to leave it as-is of course!

src/styles/global/_custom-bootstrap.scss
```diff
@@ -1,9 +1,9 @@
...
-$primary: $udb-blue;
-$secondary: $udb-grey;
-$success: #5cb85c;
-$danger: #d9534f;
-$light: $white;
+// $primary: $udb-blue;
+// $secondary: $udb-grey;
+// $success: #5cb85c;
+// $danger: #d9534f;
+// $light: $white;
```
bertramakers, line 6: (The PR is still a draft so you're probably still actively making changes, but just a reminder to not forget to remove these later if we end up not needing them here.)

src/components/publiq-ui/Alert.js
bertramakers: Btw we should also overwrite the border-radius here, as we normally don't have one on alerts but Bootstrap adds one.

Can probably be done the same way as on `Button`.

Screenshot 2020-10-28 at 10 05 59

simon-debruijn:
Bring it on with your comments! I'm mentally ready for it, for now 😁

src/components/publiq-ui/Button.js
simon-debruijn: I would have to change `getValueFromTheme` first tho, or not?

geoffreydhuyvetters: with 'change the function' I meant that, yes

src/components/publiq-ui/Button.stories.mdx
```diff
@@ -4,10 +4,11 @@ import { Button, ButtonVariants } from './Button.js';
 <Meta title="Components/Button" />
 
 export const commonArgs = {
-  visible: true,
-  dismissible: true,
+  disabled: false,
 };
 
+export const commonArgTypes = { onClick: { action: 'clicked' } };
```
geoffreydhuyvetters, line 10: Where is the name 'commonArgs' coming from? Aren't these the actual props of the component?

src/components/publiq-ui/Container.js
```diff
@@ -0,0 +1,29 @@
...
+    max-width: 970px;
+  }
+
+  @media (min-width: 1200px) {
+    width: 1170px;
+    max-width: 1170px;
+  }
+`;
+
+const Container = ({ children }) => {
```
geoffreydhuyvetters, line 21: Make sure we always pass down className too. If we would like to wrap this component with a styled() again, it needs access to the className.

src/components/publiq-ui/Small.js
```diff
@@ -0,0 +1,16 @@
+import PropTypes from 'prop-types';
+import styled from 'styled-components';
+
+const StyledSmall = styled.small`
+  font-size: 65%;
```
geoffreydhuyvetters, line 5: I think I would prefer to use rem's for font-sizes as much as we can

src/components/publiq-ui/Button.stories.mdx
```diff
@@ -4,10 +4,11 @@ import { Button, ButtonVariants } from './Button.js';
 <Meta title="Components/Button" />
 
 export const commonArgs = {
```
geoffreydhuyvetters, line 6: Where is the name 'commonArgs' coming from? Aren't these the actual props of the component?

simon-debruijn: You could name name `commonArgs` `commonProps` for example, it's named this way because of the storybook naming of `args`.

simon-debruijn:
@geoffreydhuyvetters Can I merge this PR? Then we can do specific PR's for the components themself

PR #103, opened 2020-11-03
Feature/iii 3588 fix checkbox

simon-debruijn:
### Fixed

- label id (htmlFor)
- label with checkbox

---

### DIFF
--- a/src/components/publiq-ui/Label.js
+++ b/src/components/publiq-ui/Label.js
@@ -6,14 +6,14 @@ const StyledLabel = styled(Form.Label)`
   font-weight: 700;
 `;
 
-const Label = ({ id, children, className }) => (
-  <StyledLabel htmlFor={id} className={className}>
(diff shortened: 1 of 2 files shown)


brampauwelyn (approved):

src/components/publiq-ui/Label.js
```diff
@@ -6,11 +6,14 @@ const StyledLabel = styled(Form.Label)`
   font-weight: 700;
 `;
 
-const Label = ({ children, className }) => (
-  <StyledLabel className={className}>{children}</StyledLabel>
+const Label = ({ id, children, className }) => (
+  <StyledLabel htmlFor={id} className={className}>
```
geoffreydhuyvetters, line 10: Shouldn't we just keep the htmlFor prop for this? `id` looks as if you're providing the label an id, which is not the case

geoffreydhuyvetters (approved):

PR #105, opened 2020-11-03
Feature/iii 3588 title component

simon-debruijn:
### Added

- Title component
- Title story
- PageTitle component
- PageTitle story

---

Ticket: https://jira.uitdatabank.be/browse/III-3588

PR #110, opened 2020-11-04
Feature/iii 3588 page component

simon-debruijn:
### Added

- Page component
- Page story

---

**!!! First this [PR](https://github.com/cultuurnet/udb3-frontend/pull/105) must be merged**


src/components/publiq-ui/Page.js
```diff
@@ -0,0 +1,26 @@
+import PropTypes from 'prop-types';
+import styled from 'styled-components';
+
+const StyledSection = styled.section`
+  background-color: #f0f0f0;
```
brampauwelyn, line 5: Just an idea. Not sure if it's the way to go...
Maybe centralize colors in `Theme.js`or `Colors.js` file? 
And use a variable here instead e.g. Colors.lightGrey

What do you think @geoffreydhuyvetters  ?

brampauwelyn (approved):

src/components/publiq-ui/Page.js
geoffreydhuyvetters: In this case, it might make sense to store them in the theme, yes. Since this is an application-wide thing that will probably not change when we refactor to more primitives.

geoffreydhuyvetters (approved):

PR #117, opened 2020-11-09
III-3588 / Primitive components

geoffreydhuyvetters:
### Added

- 

---


src/components/publiq-ui/Alert.js
```diff
@@ -1,9 +1,8 @@
-import { pick } from 'lodash';
 import PropTypes from 'prop-types';
 import { Alert as BootstrapAlert } from 'react-bootstrap';
 import styled from 'styled-components';
 import { getValueFromTheme } from './theme';
-import { spacingProps, spacingPropTypes } from './Box';
+import { getLayoutProps, spacingProps, spacingPropTypes } from './Box';
```
geoffreydhuyvetters, line 5: Maybe we should unify it to getBoxProps, boxProps, boxPropTypes? Since I can imagine some props being added to Box later eventually? What do you think?

PR #118, opened 2020-11-10
Feature/iii 3588 give components box props

simon-debruijn:
### Added

- box props to components 

---


src/components/publiq-ui/Inline.js
```diff
@@ -34,30 +27,50 @@ const Inline = ({
   justifyContent,
   ...props
 }) => {
-  const clonedChildren = Children.map(children, (child, i) =>
-    cloneElement(child, {
+  const clonedChildren = Children.map(children, (child, i) => {
+    // if child is normal text
+    if (typeof child === 'string')
```
bertramakers, line 32: Nit: The `if` statement should probably use braches `{}` here, like below?

src/components/publiq-ui/Alert.js
```diff
@@ -32,15 +21,22 @@ const Alert = ({
   ...props
 }) => {
   return (
-    <StyledBootstrapAlert
+    <BootstrapAlert
+      forwardedAs={Box}
```
bertramakers, line 25: Here we do `<BootstrapAlert forwardedAs={Box} ...>` and in another component we do `<Box forwardedAs={BootstrapBadge}>`. In the other components we also switch between both approaches. What is the reasoning behind this? When should we use which approach? @geoffreydhuyvetters can you clarify this maybe? Simon doesn't know for sure he told me.

PR #119, opened 2020-11-13
Feature/iii-3590 - menubar and iframe

simon-debruijn:
### Added

- Icon component
- Icon story
- SideBar
- Iframe
- redirecting routes 
- part of login flow

---


package.json
```diff
@@ -21,14 +21,19 @@
...
+    "@fortawesome/fontawesome-svg-core": "^1.2.32",
+    "@fortawesome/free-brands-svg-icons": "^5.15.1",
+    "@fortawesome/free-solid-svg-icons": "^5.15.1",
+    "@fortawesome/react-fontawesome": "^0.1.12",
     "bootstrap": "^4.5.3",
     "i18next": "^19.8.3",
     "i18next-browser-languagedetector": "^6.0.1",
     "lodash": "^4.17.20",
-    "next": "^10.0.0",
+    "next": "10.0.1",
```
geoffreydhuyvetters, line 32: Why are we removing the semver range here?

src/components/SideBar.js
```diff
@@ -0,0 +1,249 @@
...
+                  'hover.backgroundColor',
+                )};
+              }
+            }
+          `}
+          onClick={onClick}
+        >
+          <Content iconName={iconName}>{children}</Content>
+        </Button>
+      )}
```
geoffreydhuyvetters, line 92: Not sure if it makes sense to change MenuItem for 1 scenario. a completely custom menu for the notifications would make more sense imo

src/components/SideBar.js
```diff
@@ -0,0 +1,249 @@
...
+          > :not(:first-child) {
+            border-top: 1px solid ${getValueForMenu('borderColor')};
+          }
+        `}
+      >
+        <Menu items={userMenu} />
+        {/* TODO: this should work 'justifyContent="space-between"' */}
+        <Stack css="flex: 1; justify-content: space-between;">
+          <Menu items={manageMenu} title={t('menu.management')} />
+          <Menu items={notificationMenu} />
```
geoffreydhuyvetters, line 242: See comment above.

src/components/SideBar.js
```diff
@@ -0,0 +1,249 @@
...
+        css={`
+          flex: 1;
+          > :not(:first-child) {
+            border-top: 1px solid ${getValueForMenu('borderColor')};
+          }
+        `}
+      >
+        <Menu items={userMenu} />
+        {/* TODO: this should work 'justifyContent="space-between"' */}
+        <Stack css="flex: 1; justify-content: space-between;">
```
geoffreydhuyvetters, line 240: Should investigate this, because this should work.

src/components/publiq-ui/Link.js
```diff
@@ -22,7 +22,7 @@ const Link = ({ href, children, as, className, ...props }) => {
 
   if (isInternalLink) {
     return (
-      <NextLink prefetch href={href} passHref>
+      <NextLink href={href} passHref>
```
geoffreydhuyvetters, line 25: Why did you remove the prefetch here?

src/layouts/default.js
```diff
@@ -0,0 +1,28 @@
...
+      </Box>
+    </Inline>
+  );
+};
+
+Default.propTypes = {
+  children: PropTypes.node,
+};
+
+export default Default;
```
geoffreydhuyvetters, line 28: I think this makes sense to name default in Nuxt, because Nuxt automatically loads it as the default template.

But, it looks really weird when you use the component yourself. Could we give it a more descriptive name?

src/pages/_app.js
```diff
@@ -3,14 +3,19 @@ import '../styles/global.scss';
...
-    </ThemeProvider>
-  </I18nextProvider>
-);
+const App = ({ Component, pageProps }) => {
+  return (
+    <I18nextProvider i18n={i18n}>
+      <ThemeProvider>
+        <CookiesProvider>
+          <Component {...pageProps} />
+        </CookiesProvider>
```
geoffreydhuyvetters, line 15: I guess choosing the cookie solution depends a lot of whether we need this Provider here (mainly for SSR reasons). If we want to go for a 100% SPA approach, we need to configure Next a bit more https://colinhacks.com/essays/building-a-spa-with-nextjs

src/pages/login/nl.js
```diff
@@ -0,0 +1,19 @@
...
+  setCookie('udb-language', 'nl', cookieOptions);
+
+  return (
+    <div>
+      <LoginButton language="nl" />
+    </div>
+  );
+};
+
+export default Nl;
```
geoffreydhuyvetters, line 19: nl and fr should be the same page but with different copy, can't we solve this by using a [...params].js file?

public/assets/udb-logo-mobile.svg
```diff
@@ -0,0 +1,16 @@
...
+      />
+    </g>
+</svg>
```
geoffreydhuyvetters, line 16: We can make React components that render SVG code out of them, that way we don't have to load them via image tags, same the component below

next.config.js
```diff
@@ -5,4 +5,22 @@ module.exports = {
...
+  },
+  env: {
+    API_KEY: process.env.API_KEY,
+    API_URL: process.env.API_URL,
+    LEGACY_APP_URL: process.env.LEGACY_APP_URL,
+    AUTH_URL: process.env.AUTH_URL,
+    SOCKET_URL: process.env.SOCKET_URL,
+    NEW_FEATURES_URL: process.env.NEW_FEATURES_URL,
+    ENV_NAME: process.env.ENV_NAME,
+  },
```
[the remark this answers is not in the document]
simon-debruijn: I wasn't able to make it work by only adding .local, found this as the solution

src/components/LoginButton.js
```diff
@@ -0,0 +1,39 @@
+import PropTypes from 'prop-types';
+import { useCookies } from 'react-cookie';
```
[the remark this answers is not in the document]
simon-debruijn: We don't I picked this one because it matches well with the hooks-philosophy of React. But I was also worried this will possibly have to be replaced by another package if we want to use the cookies in plain javascript files

next.config.js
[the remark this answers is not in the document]
geoffreydhuyvetters: should work just fine, let's investigate together

src/components/LoginButton.js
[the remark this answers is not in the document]
geoffreydhuyvetters: Not every function in a component needs to be a hook, what would be the advantage of using a hook here? Does it automatically handle internal state or re-renders?

next.config.js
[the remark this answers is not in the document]
geoffreydhuyvetters: https://nextjs.org/docs/basic-features/environment-variables#exposing-environment-variables-to-the-browser

src/components/publiq-ui/Link.js
simon-debruijn: I got console info that prefetch wasn't needed anymore

geoffreydhuyvetters: Oh, nice. Probably new since 10 then.

PR #120, opened 2020-11-18
Feature/iii 3591 refactor giftbox (announcements)

simon-debruijn:
### Added

- announcements modal

---

Ticket: https://jira.uitdatabank.be/browse/III-3591

Screenshots:
Schermafbeelding 2020-11-19 om 09 47 08
Schermafbeelding 2020-11-19 om 09 47 27


simon-debruijn:
I still have to move the logic of announcements and seenAnnouncements up to SideBar, to be able to show the number of unseen announcements in the SideBar

src/components/AnnouncementsModal.js
```diff
@@ -0,0 +1,305 @@
...
+            &:hover {
+              opacity: 0.85;
+            }
+          `}
+        />
+      </Link>
+    )}
+
+    <div
+      dangerouslySetInnerHTML={{ __html: body }}
```
bertramakers, line 154: Nit: Maybe we should investigate a way to render this HTML with some kind of whitelist, using e.g. https://github.com/cure53/DOMPurify ?
@geoffreydhuyvetters do you have experience with cases like this where we can't block all HTML from being rendered, but should also prevent a security issue in an external system from causing an XSS vulnerability in our app?

bertramakers: To clarify: We're fetching this data from our own CMS for our public website, publiq.be, which in theory we can trust. But it's not impossible that that gets hacked at some point and someone inserts script tags etc into the content that we're fetching and rendering here.

src/components/SideBar.js
```diff
@@ -91,8 +97,30 @@ Menu.propTypes = {
   title: PropTypes.string,
 };
 
+const fetchAnnouncements = async () => {
+  const res = await fetch(process.env.NEXT_PUBLIC_NEW_ANNOUNCEMENTS_URL);
+  const { data } = await res.json();
+  return data;
+};
```
geoffreydhuyvetters, line 104: I would prefer to use something like react-query in our codebase instead of directly calling endpoints with fetch https://react-query.tanstack.com/

src/components/AnnouncementsModal.js
```diff
@@ -0,0 +1,305 @@
...
+const getValueForAnnouncementItem = getValueFromTheme('announcementItem');
+const getValueForAnnouncementList = getValueFromTheme('announcementList');
+const getValueForAnnouncementContent = getValueFromTheme('announcementContent');
+
+const AnnouncementItem = ({
+  id,
+  title,
+  selected,
+  seen,
+  setSelectedAnnouncementId,
```
geoffreydhuyvetters, line 27: I would add an onClick here and link the handlers on the surrounding component instead of passing down the setter

src/components/AnnouncementsModal.js
```diff
@@ -0,0 +1,305 @@
...
+          background-color: ${selected
+            ? getValueForAnnouncementItem('selected.hoverBackgroundColor')
+            : getValueForAnnouncementItem('hoverBackgroundColor')};
+        }
+      `}
+      onClick={() => {
+        setSelectedAnnouncementId(id);
+      }}
+    >
+      {seen ? <Icon name={Icons.EYE_SLASH} /> : <Icon name={Icons.EYE} />}
```
geoffreydhuyvetters, line 50: I would prefer a status instead of a boolean flag

```js
status === Announcement.SEEN
```

src/components/AnnouncementsModal.js
```diff
@@ -0,0 +1,305 @@
...
+    })}
+  </List>
+);
+
+AnnouncementsList.propTypes = {
+  announcements: PropTypes.array,
+  selectedAnnouncementId: PropTypes.string,
+  seenAnnouncements: PropTypes.array,
+  setSelectedAnnouncementId: PropTypes.func,
+};
```
geoffreydhuyvetters, line 111: Is it usefull to wrap up this in another component? would simplify a lot of the 'prop drilling'

src/components/SideBar.js
```diff
@@ -44,7 +48,9 @@ const MenuItem = ({ href, iconName, children, onClick }) => {
         spacing={3}
       >
         <Icon name={iconName} />
-        <Box as="span">{children}</Box>
+        <Box as="div" css="text-align: left; width: 100%;">
```
geoffreydhuyvetters, line 51: the default of a box is a div

src/components/SideBar.js
```diff
@@ -91,8 +97,30 @@ Menu.propTypes = {
...
+    const newAnnouncements = await fetchAnnouncements();
+    setAnnouncements(newAnnouncements);
+  }, []);
+
+  useEffect(() => {
+    if (announcements.length === 0) return;
+    setCountUnseenAnnouncements(
+      announcements.length - seenAnnouncements.length,
+    );
+  }, [announcements, seenAnnouncements]);
```
geoffreydhuyvetters, line 123: Can't we derive this instead of recalculating and storing it? If it's super intensive (which it isn't) we can wrap it in a useMemo.

src/components/AnnouncementsModal.js
```diff
@@ -0,0 +1,305 @@
...
+
+    const selectedAnnouncement = announcements.find(
+      (announcement) => announcement.uid === selectedAnnouncementId,
+    );
+    setSelectedAnnouncement(selectedAnnouncement);
+
+    if (!seenAnnouncements.includes(selectedAnnouncementId)) {
+      addToSeenAnnouncements(selectedAnnouncementId);
+    }
+  }, [selectedAnnouncementId]);
```
geoffreydhuyvetters, line 255: Might make sense to reevaluate what is state and what is calculated here. I have the feeling we're going over the top with the effects here.

src/components/AnnouncementsModal.js
```diff
@@ -0,0 +1,305 @@
...
+        }
+        ul {
+          display: block;
+          list-style-type: disc;
+          margin-block-start: 1em;
+          margin-block-end: 1em;
+          margin-inline-start: 0px;
+          margin-inline-end: 0px;
+          padding-inline-start: 40px;
+        }
```
geoffreydhuyvetters, line 182: We can use our spacing system here too.

src/components/publiq-ui/Stack.js
```diff
@@ -23,6 +23,8 @@ const Stack = forwardRef(
     const clonedChildren = Children.map(children, (child, i) => {
       const isLastItem = i === children.length - 1;
 
+      if (!child) return;
```
geoffreydhuyvetters, line 26: In what situation would that happen?

src/components/SideBar.js
geoffreydhuyvetters: I can help set this up on Monday if you want.

src/components/SideBar.js
```diff
@@ -159,36 +201,47 @@ const SideBar = () => {
...
         </Stack>
       </Stack>
-    </Stack>
+      <AnnouncementsModal
+        visible={isModalVisible}
+        announcements={announcements}
+        seenAnnouncements={seenAnnouncements}
+        setSeenAnnouncements={setSeenAnnouncements}
+        onClose={() => {
+          setModalVisibility(false);
```
geoffreydhuyvetters, line 241: Try to mirror the state name with the setter. `setIsModalVisible`

src/components/SideBar.js
```diff
@@ -159,36 +201,47 @@ const SideBar = () => {
...
+            <Menu items={manageMenu} title={t('menu.management')} />
+            <Menu items={notificationMenu} />
+          </Stack>
         </Stack>
       </Stack>
-    </Stack>
+      <AnnouncementsModal
+        visible={isModalVisible}
+        announcements={announcements}
+        seenAnnouncements={seenAnnouncements}
```
geoffreydhuyvetters, line 238: Can't we fix this by manipulating the data? Changing the status of the announcement and taking that into consideration while rendering?

src/components/SideBar.js
bertramakers: Good point. Does react-query also make it easier to mock the requests/responses in tests?

src/components/AnnouncementsModal.js
```diff
@@ -0,0 +1,305 @@
...
+      const updatedSeenAnnouncements = cookies.seenAnnouncements.filter(
+        (announcement) => !announcements.includes(announcement.uid),
+      );
+      setSeenAnnouncements(updatedSeenAnnouncements);
+    }
+  }, []);
+
+  useEffect(() => {
+    if (!visible || announcements.length === 0) return;
+    setSelectedAnnouncementId(announcements[0].uid);
```
bertramakers, line 241: Nit: The advantage of an early return is not that big here. We could also write it as
```
if (visible && announcements.length > 0) {
    setSelectedAnnouncementId(announcements[0].uid);
} 
```
Early returns are mostly useful if you have more than one check.

Or if you prefer to keep using the short syntax:
```
if (visible && announcements.length > 0) setSelectedAnnouncementId(announcements[0].uid);
```

Personally less a fan of that syntax though (also in the case of the early return).

But anyway this is all very minor.

src/components/AnnouncementsModal.js
```diff
@@ -0,0 +1,305 @@
...
+          margin-block-start: 1em;
+          margin-block-end: 1em;
+          margin-inline-start: 0px;
+          margin-inline-end: 0px;
+          padding-inline-start: 40px;
+        }
+        ul {
+          display: block;
+          list-style-type: disc;
+          margin-block-start: 1em;
```
simon-debruijn, line 177:
```suggestion
          margin-block-start: ${getSpacing(3)};
```

src/components/AnnouncementsModal.js
```diff
@@ -0,0 +1,305 @@
...
+  announcements,
+  seenAnnouncements,
+  setSeenAnnouncements,
+  onClose,
+}) => {
+  const { t } = useTranslation();
+  const cookieOptions = {
+    path: '/',
+    maxAge: 60 * 60 * 24 * 30,
+  };
```
simon-debruijn, line 214: Move outside of component

src/components/SideBar.js
bertramakers: I was thinking we could maybe combine the `AnnouncementsModal` and `Inline` that serves as a link in a single component that fetches the announcements and keeps track of which are seen? And then the `AnnouncementsModal` is a stateless component that takes the data as props and has some kind of `onRead` prop so the parent component can update the state and also push that read count down to the badge?

bertramakers:
Great work so far @simon-debruijn ! 👏 

Functionally it works very well and looks great. Out of all the review comments I personally think these three would be the biggest possible improvements:

- Using a library like React Query to make HTTP requests instead of using fetch (also improving testability I believe)
- Improving the state management of the announcements, which are seen, etc
- Looking into the security implications of rendering the articles as-is without any filtering => This could also be split off since it was already an issue in Vue I believe, and it will require some analysis to figure out what elements we want to render  and which we want to filter out.

src/components/AnnouncementsModal.js
simon-debruijn: We'll make this a separate ticket and PR?

src/components/SideBar.js
geoffreydhuyvetters: It wraps fetch with a hook so I don't think it will make it easier

src/components/AnnouncementsModal.js
bertramakers: 👍

src/components/SideBar.js
bertramakers: Okay, I found a comment by the author about this and his solution seems sufficient I think https://github.com/tannerlinsley/react-query/issues/113#issuecomment-586822007

src/components/SideBar.js
geoffreydhuyvetters: Discussed a different approach with @simon-debruijn in a call, we should be able to treat announcements as the single source of truth and derive everything from there

bertramakers: Awesome 👍

src/components/AnnouncementsModal.js
```diff
@@ -0,0 +1,246 @@
...
+  onShow,
+  onClose,
+}) => {
+  const { t } = useTranslation();
+
+  const activeAnnouncement = () => {
+    return announcements.find(
+      (announcement) => announcement.status === AnnouncementItemStatus.ACTIVE,
+    );
+  };
```
geoffreydhuyvetters, line 178: Don't think this needs to be a function, no?

simon-debruijn: Yo have to filter the announcement that has status 'active' to show the content of, no?

src/provider/ContextProvider.js
```diff
@@ -0,0 +1,18 @@
...
+import { queryCache, ReactQueryCacheProvider } from 'react-query';
+
+const providers = [
+  ({ children }) => I18nextProvider({ children, i18n }),
+  ThemeProvider,
+  CookiesProvider,
+  ({ children }) => ReactQueryCacheProvider({ children, queryCache }),
+];
+
+const ContextProvider = combineComponents(...providers);
```
geoffreydhuyvetters, line 16: I think I would prefer to make it a little bit more declarative like this.

https://medium.com/@juliuskoronci/avoid-a-long-list-of-react-providers-c45a269d80c1

src/components/AnnouncementsModal.js
geoffreydhuyvetters: Sure, but not as a function?

PR #122, opened 2020-11-24
Feature/iii 3589 login flow

simon-debruijn:
### Added

-

### Changed

-

### Removed

-

### Fixed

-

---

Ticket: https://jira.uitdatabank.be/browse/III-3589

### DIFF
--- a/src/components/SideBar.js
+++ b/src/components/SideBar.js
@@ -117,10 +117,11 @@ const ProfileMenu = ({ profileImage }) => {
         removeCookie('token');
         removeCookie('user');
 
-        const baseUrl = `${window.location.protocol}//${window.location.host}`;
+        const getBaseUrl = () =>
+          `${window.location.protocol}//${window.location.host}`;
 
         const queryString = new URLSearchParams({
-          destination: baseUrl,
+          destination: getBaseUrl,
         }).toString();
 
         router.push(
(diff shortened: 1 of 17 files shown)


simon-debruijn:
@geoffreydhuyvetters Do you see the issue?

src/pages/_app.js
```diff
@@ -44,6 +47,30 @@ Layout.propTypes = {
...
+      if (!cookies?.token) {
+        router.push('/login');
+      }
+    }
+  }, [query, pathname]);
+
+  useEffect(() => {
+    if (user) {
+      setCookie('user', user);
+      // TODO: Currently after logging in it returns to the login page again after loading dashboard, this forces to go back to dashboard
```
simon-debruijn, line 69: @geoffreydhuyvetters

bertramakers:
Looks good to me aside from the workaround that needs to be fixed, but I'm also not sure what's causing the redirect to `/login` after the `token` is set.

src/components/SideBar.js
```diff
@@ -105,12 +105,21 @@ Menu.propTypes = {
...
       children: t('menu.logout'),
-      onClick: () => {},
+      onClick: () => {
+        removeCookie('token');
+        removeCookie('user');
+
+        const queryString = new URLSearchParams({
+          destination: `${window.location.protocol}//${window.location.host}`,
+        }).toString();
+        window.location.href = `${process.env.NEXT_PUBLIC_AUTH_URL}/logout?${queryString}`;
```
geoffreydhuyvetters, line 121: Just out of curiousity. Could we also use router.push to navigate to an external page?

geoffreydhuyvetters (approved):

src/components/SideBar.js
simon-debruijn: Does seem to work perfectly with router.push. I've changed it

PR #123, opened 2020-11-25
Feature/iii 3589 message handler iframe

simon-debruijn:
### Added

- event handler for iframe messages

---

### DIFF
--- a/src/pages/[...params].js
+++ b/src/pages/[...params].js
@@ -14,10 +14,11 @@ const Fallback = () => {
       router.push('/login/');
     }
 
-    delete query.params;
+    const queryParameters = { ...query };
+    delete queryParameters.params;
 
     const queryString = new URLSearchParams({
-      ...query,
+      ...queryParameters,
       jwt: cookies.token,
       lang: i18next.language,
     }).toString();
--- a/src/pages/[...params].js
+++ b/src/pages/[...params].js
@@ -14,6 +14,8 @@ const Fallback = () => {
       router.push('/login/');
     }
 
+    delete query.params;
+
     const queryString = new URLSearchParams({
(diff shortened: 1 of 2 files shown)


src/pages/[...params].js
```diff
@@ -14,6 +14,8 @@ const Fallback = () => {
       router.push('/login/');
     }
 
+    delete query.params;
```
Comment (bertramakers) [inline src/pages/[...params].js:17]: Nit: Maybe we should add a comment here to explain why we need to do this.

Comment (bertramakers) [inline src/pages/[...params].js:17]:
```suggestion
    // Remove the "params" query parameter added by Next 
    // (which holds the current path, based on `[...params]` in the filename)
    // because it causes infinite redirects as it keeps adding this query parameter to the iframe's URL
    delete query.params;
```

bertramakers (approved):

src/pages/[...params].js
```diff
@@ -14,8 +14,16 @@ const Fallback = () => {
       router.push('/login/');
     }
 
+    /* 
+    Remove the "params" query parameter added by Next 
+    (which holds the current path, based on `[...params]` in the filename)
+    because it causes infinite redirects as it keeps adding this query parameter to the iframe's URL 
+    */
+    const queryParameters = { ...query };
+    delete queryParameters.params;
```
Comment (geoffreydhuyvetters) [inline src/pages/[...params].js:23]: You could also just keep grab what you need and ignore params

```js
  const { query: {...queryWithoutParams, params} , asPath, ...router } = useRouter();
```

It's an example where using deeper restructuring would make sense, otherwise, you would need to create an unused intermediate variable

geoffreydhuyvetters (approved):

PR #124, opened 2020-11-25
Feature/iii 3590 filtering menu by permissions

simon-debruijn:
### Added

- get request for permissions
- get request for roles
- get request for eventsToModerate
- counter for eventsToModerate
- filtering of manageMenu based on the permissions of the user

---


src/hooks/api/user.js
```diff
@@ -1,26 +1,64 @@
...
+const getRoles = async (headers) => {
+  const res = await fetch(`${process.env.NEXT_PUBLIC_API_URL}/user/roles/`, {
+    headers,
+  });
+  return await res.json();
+};
+
+const useGetRoles = (config = {}) => {
+  const [cookies] = useCookiesWithOptions(['token']);
+  const headers = useHeaders();
```
bertramakers, line 57: I'm wondering if we could move this inside for example `fetchWithRedirect` or `useAuthenticatedQuery` so we don't have to repeat this for every API request hook.
(Looking at the name `useAuthenticatedQuery` I'd also assume at first glance that that would add the necessary authentication headers)

src/pages/[...params].js
```diff
@@ -14,10 +14,11 @@ const Fallback = () => {
       router.push('/login/');
     }
 
-    delete query.params;
+    const queryParameters = { ...query };
+    delete queryParameters.params;
```
Comment (bertramakers) [inline src/pages/[...params].js:18]: This change (also) belongs in #123 . Maybe you can cherry-pick the commit with this change to that PR as well?

Comment (bertramakers) [reply src/pages/[...params].js:18]: Note also my comment on #123 to add a comment here to explain why we remove the `.params`

src/hooks/api/user.js
simon-debruijn: See changes I've made in [this commit](https://github.com/cultuurnet/udb3-frontend/pull/124/commits/473e5060576d86889700a6581fe4b1c68a9e16bc)

bertramakers: Looks great!

src/hooks/api/events.js
```diff
@@ -0,0 +1,30 @@
+import { fetchWithRedirect } from '../../utils/fetchWithRedirect';
+import { useAuthenticatedQuery } from './useAuthenticatedQuery';
+
+const findToModerate = async (headers, searchQuery, start = 0, limit = 1) => {
+  const url = new URL(`${process.env.NEXT_PUBLIC_API_URL}/events/`);
+  url.search = new URLSearchParams({
+    q: searchQuery,
+    audienceType: 'everyone',
+    availableFrom: new Date().toISOString().split('.')[0] + 'Z',
```
bertramakers, line 9: Nit: Maybe we could put this in a separate utils file like `fetchWithRedirect`, because most of our backend services can't deal with the microseconds in the `toISOString()` format right now so we'll probably need to do this formatting in multiple places.

bertramakers (approved):

src/hooks/api/events.js
simon-debruijn: @bertramakers are you referring to this particular code? 
`.toISOString().split('.')[0] + 'Z'`

src/hooks/api/events.js
```diff
@@ -0,0 +1,35 @@
...
+  const res = await fetchWithRedirect(url, {
+    headers,
+  });
+  return await res.json();
+};
+
+const useFindToModerate = (searchQuery, config) => {
+  const [cookies] = useCookies(['token']);
+  return useAuthenticatedQuery(
+    'findToModerate',
```
geoffreydhuyvetters, line 26: This is where you need to look into how passing arguments and caching works in react-query

src/components/SideBar.js
```diff
@@ -147,22 +151,41 @@ ProfileMenu.defaultProps = {
...
+
 const SideBar = () => {
   const { t } = useTranslation();
 
+  const [cookies, setCookie] = useCookiesWithOptions([
+    'seenAnnouncements',
+    'userPicture',
+  ]);
+
+  const [filteredManageMenu, setFilteredManageMenu] = useState();
```
geoffreydhuyvetters, line 174: We don't need to keep this in state.

src/components/SideBar.js
```diff
@@ -228,60 +303,46 @@ const SideBar = () => {
 
   const manageMenu = [
     {
+      neededPermission: PermissionTypes.AANBOD_MODEREREN,
```
geoffreydhuyvetters, line 306: I think 'permission' is enough here.

src/components/SideBar.js
```diff
@@ -147,22 +151,41 @@ ProfileMenu.defaultProps = {
...
-  const [activeAnnouncementId, setActiveAnnouncementId] = useState();
-  const [cookies, setCookie] = useCookiesWithOptions([
-    'seenAnnouncements',
-    'userPicture',
-  ]);
+  const { data: permissions = [] } = useGetPermissions();
+  const { data: roles = [] } = useGetRoles();
+  const {
+    data: { totalItems: countEventsToModerate = 0 } = {},
+  } = useFindToModerate(searchQuery);
```
geoffreydhuyvetters, line 188: useGetEventsToModerate

src/hooks/api/events.js
```diff
@@ -0,0 +1,35 @@
...
+    (headers) => findToModerate(headers, searchQuery),
+    {
+      enabled: !!(cookies.token && searchQuery),
```
geoffreydhuyvetters, line 29: An empty object is also true, is this the behaviour we want?

src/components/SideBar.js
```diff
@@ -147,22 +151,41 @@ ProfileMenu.defaultProps = {
...
-
-  const [activeAnnouncementId, setActiveAnnouncementId] = useState();
-  const [cookies, setCookie] = useCookiesWithOptions([
-    'seenAnnouncements',
-    'userPicture',
-  ]);
+  const { data: permissions = [] } = useGetPermissions();
+  const { data: roles = [] } = useGetRoles();
+  const {
+    data: { totalItems: countEventsToModerate = 0 } = {},
```
geoffreydhuyvetters, line 187: Really not a big fan of several levels deep destructuring.

You can make good usage of ?? and ? here

```js
data: eventsToModerate
```

```js
const countEventsToModerate = data?.totalItems ?? 0
```

src/components/SideBar.js
```diff
@@ -208,6 +231,58 @@ const SideBar = () => {
...
+    if (permissions.length === 0) {
+      return;
+    }
+
+    setFilteredManageMenu(
+      manageMenu.filter((menuItem) =>
+        permissions.includes(menuItem.neededPermission),
+      ),
+    );
+  }, [permissions]);
```
geoffreydhuyvetters, line 244: Refactor this into a computed value

src/components/SideBar.js
```diff
@@ -147,22 +151,41 @@ ProfileMenu.defaultProps = {
...
+  ]);
+
+  const [filteredManageMenu, setFilteredManageMenu] = useState();
   const [isJobLoggerVisible, setJobLoggerVisibility] = useState(false);
   const [isModalVisible, setIsModalVisible] = useState(false);
+  const [searchQuery, setSearchQuery] = useState('');
+  const [activeAnnouncementId, setActiveAnnouncementId] = useState();
 
   const {
     data: { data: rawAnnouncements = [] } = {},
```
geoffreydhuyvetters, line 181: Same here, try to avoid deep destructuring

src/hooks/api/events.js
```diff
@@ -0,0 +1,35 @@
...
+    headers,
+  });
+  return await res.json();
+};
+
+const useFindToModerate = (searchQuery, config) => {
+  const [cookies] = useCookies(['token']);
+  return useAuthenticatedQuery(
+    'findToModerate',
+    (headers) => findToModerate(headers, searchQuery),
```
geoffreydhuyvetters, line 27: A bit confused on why we're passing down headers here? (and where they are coming from)

src/hooks/api/useAuthenticatedQuery.js
```diff
@@ -1,10 +1,20 @@
...
 import { Errors } from '../../utils/fetchWithRedirect';
+import { useCookiesWithOptions } from '../useCookiesWithOptions';
+import { useHeaders } from './useHeaders';
 
 const useAuthenticatedQuery = (...args) => {
   const router = useRouter();
-  const result = useReactQuery(...args);
+  const headers = useHeaders();
+  const [cookies] = useCookiesWithOptions(['token']);
+  const [key, func, config] = args;
```
geoffreydhuyvetters, line 11: can we rename func to 'queryFunction' to make it a bit clearer

src/hooks/api/useAuthenticatedQuery.js
```diff
@@ -1,10 +1,20 @@
...
+  const alteredArgs = [
+    key,
+    () => func(headers),
```
geoffreydhuyvetters, line 14: This should be solved if we use the correct way to pass down variables

src/hooks/api/events.js
geoffreydhuyvetters: Think I understand it now, nevermind.

src/hooks/api/events.js
simon-debruijn: I believe token and searchQuery are strings

src/hooks/api/events.js
simon-debruijn: I've have changed it to what I believe should work, based on react-queries docs

PR #125, opened 2020-11-27
Feature/iii 3599 showing notifications in panel

simon-debruijn:
### Added

- notifications on export

---

Ticket: https://jira.uitdatabank.be/browse/III-3599


simon-debruijn:
You can go through and review, there is still a small issue though. The icon shown dependent on jobLoggerStatus does the animation multiple times

PR #126, opened 2020-11-30
Introduce responsive props + internal refactors of components

geoffreydhuyvetters:
### Added

- Added the possibility to pass an object as styling prop, to use for hover states and breakpoints

```js

backgroundColor={{
  hover: 'green',
  s: 'yellow',
  default: 'orange'
}}

```

### Changed

- did some internal component refactors while at it

the breakpoint with their values are defined in the theme


src/components/publiq-ui/Panel.js
```diff
@@ -1,27 +1,61 @@
...
-    {...getStackProps(props)}
-    className={className}
-    marginBottom={4}
-  >
-    {children}
-  </Stack>
-);
+const Panel = ({ children, className, ...props }) => {
+  const parsedChildren =
+    Children.count(children) === 1 ? <>{children}</> : children;
```
simon-debruijn, line 10: I see this line is added in several components. But why is this needed?

simon-debruijn (approved):

src/components/publiq-ui/Panel.js
geoffreydhuyvetters: It's a fix for passing a single element to Panel and ListItem (these are Inlines now).

PR #130, opened 2020-12-03
Feature/iii 3644 display username correctly

simon-debruijn:
### Added

- Username in sidebar

### Fixed

- `Text content did not match. Server: "" Client:` error from NextJs

---

Ticket: https://jira.uitdatabank.be/browse/III-3644

### DIFF
--- a/src/components/SideBar.js
+++ b/src/components/SideBar.js
@@ -96,8 +96,14 @@ Menu.propTypes = {
 
 const ProfileMenu = ({ profileImage }) => {
   const { t } = useTranslation();
-  const [, , removeCookie] = useCookiesWithOptions();
+  const [cookies, , removeCookie] = useCookiesWithOptions();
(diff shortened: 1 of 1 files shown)


src/components/SideBar.js
```diff
@@ -96,8 +96,14 @@ Menu.propTypes = {
 
 const ProfileMenu = ({ profileImage }) => {
   const { t } = useTranslation();
-  const [, , removeCookie] = useCookiesWithOptions();
+  const [cookies, , removeCookie] = useCookiesWithOptions(['user']);
   const router = useRouter();
+  const [username, setUsername] = useState('');
```
geoffreydhuyvetters, line 101: Do we have to set it in the state? We can use the cookie value and default to an empty string, no?

geoffreydhuyvetters (approved):

PR #134, opened 2020-12-09
Feature/iii 3592 productions UI - List view

simon-debruijn:
### Added

- `/productions` page with list view

---

Ticket: https://jira.uitdatabank.be/browse/III-3600


src/hooks/api/productions.js
```diff
@@ -25,4 +27,36 @@ const useGetProductions = ({ name = '', start = 0, limit = 15 }, config = {}) =>
...
+
+const deleteEventsByIds = async ({
+  productionId = '',
+  eventIds = [],
+  headers,
+}) => {
+  const mappedEvents = eventIds.map((eventId) => {
+    return deleteEventById({ productionId, eventId, headers });
+  });
+  return await Promise.all(mappedEvents);
```
geoffreydhuyvetters, line 56: not 100% sure, but I don't think you need to await a promise.

src/pages/manage/productions/index.js
```diff
@@ -45,22 +45,25 @@ const Index = () => {
...
-    const rawProductions = productionsData?.member ?? [];
     return rawProductions.map((production) => ({
       ...production,
       id: production.production_id,
       active: production.production_id === activeProductionId,
     }));
-  }, [productionsData, activeProductionId]);
-
-  useEffect(() => setActiveProductionId(productions?.[0]?.id), [productions]);
+  }, [activeProductionId]);
```
geoffreydhuyvetters, line 62: I think we also need to pass rawProductions here then?

.eslintrc
```diff
@@ -3,7 +3,8 @@
   "plugins": ["unicorn"],
   "rules": {
     "react/react-in-jsx-scope": "off",
-    "unicorn/no-abusive-eslint-disable": "error"
+    "unicorn/no-abusive-eslint-disable": "error",
+    "no-console": "warn"
```
geoffreydhuyvetters, line 7: Good idea 🎉

src/pages/manage/productions/index.js
```diff
@@ -45,22 +45,25 @@ const Index = () => {
     start: currentPageProductions - 1,
     limit: productionsPerPage,
   });
+  const rawProductions = productionsData?.member ?? [];
+
+  useEffect(() => {
+    if (rawProductions.length === 0) return;
+    // make the first production active when new data comes in from the apiCall
+    setActiveProductionId(rawProductions[0].production_id);
+  }, [rawProductions]);
```
geoffreydhuyvetters, line 54: Was there a situation where the previous solution was breaking things?

simon-debruijn: The previous solution didn't work. The active property was being overwritten as soon as a new activeProductionId was set.

src/pages/manage/productions/index.js
simon-debruijn: I don't think so because setActiveProductionId gets called when rawProductions updates

src/pages/manage/productions/index.js
geoffreydhuyvetters: Ah yeah, think I got it. Had to read that sentence 10 times though 😆

PR #140, opened 2020-12-16
Fix env variables

simon-debruijn:
### Fixed

- env variables for public runtime

---

Ticket: https://jira.uitdatabank.be/browse/III-3603


src/utils/fetchWithRedirect.js
```diff
@@ -1,9 +1,13 @@
...
+
 const Errors = {
   UNAUTHORIZED: 'unauthorized',
 };
 
-const fetchWithRedirect = async (...args) => {
-  const response = await fetch(...args);
+const fetchWithRedirect = async (path, ...args) => {
+  const { publicRuntimeConfig } = getConfig();
+  console.log({ path, args, apiUrl: publicRuntimeConfig.apiUrl });
```
geoffreydhuyvetters, line 9:
```suggestion
```

geoffreydhuyvetters:
The build seems to fail now.

src/pages/_app.js
```diff
@@ -81,13 +81,15 @@ Layout.propTypes = {
...
+const isBrowser = () => typeof window !== 'undefined';
+
 // eslint-disable-next-line react/prop-types
 const App = ({ Component, pageProps, cookies }) => (
   <ContextProvider
     providers={[
       [I18nextProvider, { i18n }],
       ThemeProvider,
-      [CookiesProvider, { cookies }],
+      [CookiesProvider, { cookies: isBrowser() ? undefined : cookies }],
```
bertramakers, line 91: Got this solution from https://github.com/reactivestack/cookies/issues/204#issuecomment-487365880
I tried to set `cookies` to `undefined` in `getInitialProps` if there was no `req`, so we wouldn't have to use this `isBrowser()` check, but that didn't work for some reason.
But it works, and if you want we can try to better optimize it tomorrow.

bertramakers: @geoffreydhuyvetters can you take a look at this if you have some time and approve if okay or look at it together in a call? I've tried a lot of things to get it to work without a check on `window`, too many to all list here, but couldn't find any.

src/pages/_app.js
```diff
@@ -79,13 +80,15 @@ Layout.propTypes = {
 const queryCache = new QueryCache();
 const queryClient = new QueryClient({ queryCache });
 
+const isBrowser = () => typeof window !== 'undefined';
```
geoffreydhuyvetters, line 83: We already have an useIsClient hook.

bertramakers: As discussed over Slack I'm not going to use that for now because it results in errors again. (The first 1-2 renders of `App` on the client-side get `false` from `useIsClient` for some reason.)
We'll look into that later.

bertramakers (approved):

PR #144, opened 2020-12-21
Compare against actual components

geoffreydhuyvetters:
### Fixed

- The actual types are minified on the production build, so the page was gone. This fixes it

### DIFF
--- a/src/components/publiq-ui/Page.js
+++ b/src/components/publiq-ui/Page.js
@@ -9,26 +9,12 @@ import { getBoxProps } from './Box';
 
 const getValueForPage = getValueFromTheme('page');
 
-const CompoundChildren = {
-  TITLE: 'PageTitle',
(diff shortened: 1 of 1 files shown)


src/components/publiq-ui/Page.js
```diff
@@ -9,26 +9,12 @@ import { getBoxProps } from './Box';
...
-  );
-
-  const actions = children.find(
-    (child) => child.type.name === CompoundChildren.ACTIONS,
-  );
-
-  const content = children.find(
-    (child) => child.type.name === CompoundChildren.CONTENT,
-  );
+  const title = children.find((child) => child.type === PageTitle);
```
simon-debruijn, line 15: Doesn't this have to be 'PageTitle', ...?

Or does it refer to the component PageTitle itself

simon-debruijn (approved):

PR #147, opened 2020-12-21
Listen for WindowMessageTypes.HTTP_STATUS_CODE

simon-debruijn:
### Added

- listening for messages from old angular app for potential redirect

---

Ticket: https://jira.uitdatabank.be/browse/III-3589

Related PR: https://github.com/cultuurnet/udb3-angular/pull/626

### DIFF
--- a/src/hooks/useHandleWindowMessage.js
+++ b/src/hooks/useHandleWindowMessage.js
@@ -8,7 +8,7 @@ const WindowMessageSources = {
 const WindowMessageTypes = {
   URL_CHANGED: 'URL_CHANGED',
   JOB_ADDED: 'JOB_ADDED',
-  HTTP_STATUS_CODE: 'HTTP_STATUS_CODE',
+  HTTP_ERROR_CODE: 'HTTP_ERROR_CODE',
 };
 
 const useHandleWindowMessage = (eventsMap = {}) => {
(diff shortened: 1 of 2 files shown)


src/pages/_app.js
```diff
@@ -64,6 +64,14 @@ const useHandleAuthentication = () => {
     }, 5000); // checking every 5 seconds
     return cleanUp;
   }, [asPath]);
+  // redirect when the old anular app throws an error because of a missing token
```
geoffreydhuyvetters, line 67:
```suggestion
  // redirect when the old angular app throws an error because of a missing token
```

src/hooks/useHandleWindowMessage.js
```diff
@@ -8,6 +8,7 @@ const WindowMessageSources = {
 const WindowMessageTypes = {
   URL_CHANGED: 'URL_CHANGED',
   JOB_ADDED: 'JOB_ADDED',
+  HTTP_STATUS_CODE: 'HTTP_STATUS_CODE',
```
geoffreydhuyvetters, line 11: Maybe we should rename it to HTTP_ERROR_CODE since it only transmits those right?

geoffreydhuyvetters (approved):

PR #150, opened 2020-12-22
Add productions create page

geoffreydhuyvetters:
### Added

- `/productions/create` page


src/components/publiq-ui/Typeahead.js
```diff
@@ -28,28 +33,44 @@ const Typeahead = ({
           background-color: ${getValue('activeBackgroundColor')};
         }
       `}
-      onInputChange={onInputChange}
+      onSearch={onSearch}
+      onInputChange={onInput}
       onChange={onSelection}
```
geoffreydhuyvetters, line 38: I would mimic the props as much as we can, so onInputChange and onSelection

src/components/publiq-ui/Typeahead.js
```diff
@@ -28,28 +33,44 @@ const Typeahead = ({
...
+      placeholder={placeholder}
+      emptyLabel={emptyLabel}
+      delay={275}
```
geoffreydhuyvetters, line 41: So this handles debouncing internally? nice.

src/hooks/api/authenticated-query.js
```diff
@@ -129,11 +131,43 @@ const useAuthenticatedMutation = ({
...
+  const getNewMockData = () => {
+    if (mockDataArray.length === 1) return mockDataArray[0];
+    return mockDataArray[Math.floor(Math.random() * mockDataArray.length)];
+  };
+
+  const refetch = () => setRandomMockData(getNewMockData());
+
+  useEffect(() => {
+    if (
+      mockData &&
```
geoffreydhuyvetters, line 151:
```suggestion
      !!mockData &&
```

Just to be sure I guess

src/pages/manage/productions/create.js
```diff
@@ -1,28 +1,314 @@
...
+
+  const suggestedProductions = suggestedProductionsData?.member ?? [];
+  const events = suggestedEvents?.events ?? [];
+  const similarity = suggestedEvents?.similarity ?? 0;
+
+  const productions = useMemo(
+    () =>
+      events
+        .map((event) => event?.production)
+        .filter((production) => !!production),
```
geoffreydhuyvetters, line 169: No need to loop twice here.

src/pages/manage/productions/create.js
```diff
@@ -1,28 +1,314 @@
...
+      return ProductionStatus.EXISTING_SELECTED;
+    }
+    if (searchInput) {
+      return ProductionStatus.NEW;
+    }
+    return ProductionStatus.MISSING;
+  }, [selectedProductionId, searchInput]);
+
+  const handleInputSearch = (searchTerm) => {
+    debounce(() => setSearchInput(searchTerm.toString().trim()), 275)();
```
geoffreydhuyvetters, line 190: Do we still need this with the debouncing in the typeahead itself?

src/pages/manage/productions/create.js
```diff
@@ -1,28 +1,314 @@
...
+  const { data: suggestedProductionsData } = useGetProductions({
+    name: searchInput,
+    limit: 10,
+  });
+
+  const handleSuccess = async () => {
+    setSelectedProductionId('');
+    setSearchInput('');
+    await refetchSuggestedEvents();
+    await queryClient.refetchQueries(['productions']);
```
geoffreydhuyvetters, line 142: I would not use refetchQueries here, this will retrigger all caches that have 'productions' in it. I would use the refetch function that is returned from useGetProductions()

src/pages/manage/productions/create.js
```diff
@@ -1,28 +1,314 @@
...
+
+  const { mutate: addEventsByIds } = useAddEventsByIds({
+    onSuccess: handleSuccess,
+  });
+
+  const suggestedProductions = suggestedProductionsData?.member ?? [];
+  const events = suggestedEvents?.events ?? [];
+  const similarity = suggestedEvents?.similarity ?? 0;
+
+  const productions = useMemo(
```
geoffreydhuyvetters, line 165: A bit unclear what this variable contains. Wouldn't linkedProductions be a bit more clearer?

src/pages/manage/productions/create.js
```diff
@@ -1,28 +1,314 @@
...
+      </Button>
+    </Inline>
+  );
+};
+
+LinkAndSkipButtons.propTypes = {
+  shouldDisableLinkButton: PropTypes.bool,
+  onClickLink: PropTypes.func,
+  onClickSkip: PropTypes.func,
+};
```
geoffreydhuyvetters, line 114: I personally think there is little value in splitting this up in a component

src/pages/manage/productions/create.js
```diff
@@ -1,28 +1,314 @@
...
+
+TwoProductions.propTypes = {
+  productionNames: PropTypes.array,
+  onChange: PropTypes.func,
+};
+
+TwoProductions.defaultProps = {
+  productionNames: [],
+};
```
geoffreydhuyvetters, line 85: I think splitting these components up is creating more mental overhead than its actually helping us.

geoffreydhuyvetters (commented):
In general, I think we can define all components in Create instead of splitting them up, seeing the code now, it just adds mental overhead (they all contain just one or 2 components. We wouldn't need to pass around all the props and add proptypes etc..

src/pages/manage/productions/create.js
simon-debruijn: The internal debouncing was causing problems, so I moved it to be our own

src/pages/manage/productions/create.js
simon-debruijn: This is to filter out the undefined values

geoffreydhuyvetters: Ah yeah, you need the productions. Misread that one

PR #151, opened 2021-01-04
Several smaller fixes for productions

geoffreydhuyvetters:
### Fixed

- fixed some productions issues
- prefetch user, permissions and roles
-

### DIFF
--- a/src/utils/getApplicationServerSideProps.js
+++ b/src/utils/getApplicationServerSideProps.js
@@ -5,7 +5,7 @@ const getApplicationServerSideProps = (callbackFn) => async ({
   req,
   query,
 }) => {
-  if (process.env.NOD_ENV !== 'production') {
+  if (process.env.NODE_ENV !== 'production') {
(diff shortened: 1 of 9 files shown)


src/utils/getApplicationServerSideProps.js
```diff
@@ -23,8 +23,6 @@ const getApplicationServerSideProps = (callbackFn) => async ({
 
   const queryClient = new QueryClient();
 
-  // TODO: prefetch user, permission, roles
```
simon-debruijn, line 26: Wasn't it possible to prefetch the user, permission and roles?

geoffreydhuyvetters: Yeah, it's possible but it adds to the loading time, a lot. It fetches with every page navigation, not just on the first hit of the application. I think it's better to leave it for now, and maybe later investigate another approach if we really want to add it.

simon-debruijn (approved):

PR #156, opened 2021-01-05
Bug/iii 3673 redirect to dashboard when a path is invalid in angular app

simon-debruijn:
### Changed

- redirect to /dashboard to a redirect to 404

---

Ticket: https://jira.uitdatabank.be/browse/III-3673

### DIFF
--- a/src/components/publiq-ui/Alert.js
+++ b/src/components/publiq-ui/Alert.js
@@ -29,7 +29,6 @@ const Alert = ({
       hidden={!visible}
       dismissible={dismissible}
       className={className}
-      {...getBoxProps(props)}
       css={`
         margin: 0;
         &.alert {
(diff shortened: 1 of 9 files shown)


bertramakers:
Very nice @simon-debruijn ! ⭐

bertramakers (approved):

src/components/publiq-ui/Alert.js
```diff
@@ -22,21 +22,22 @@ const Alert = ({
   className,
   ...props
 }) => {
+  // TODO: box props (positioning) doesn't seem to work
```
bertramakers, line 25: Nit: It's usually best to refrain from adding todo's in code and create a ticket instead so it doesn't get lost in all the other work. But since this app is still very actively being developed and optimized I trust it will get picked up 😉

simon-debruijn: Good to know 😸

simon-debruijn: https://jira.uitdatabank.be/browse/III-3676

PR #158, opened 2021-01-06
Bug/iii-3678 - event names are empty when language fr is active

simon-debruijn:
### Fixed

- Improved translations of event in productions

---

Ticket: https://jira.uitdatabank.be/browse/III-3678


src/pages/manage/productions/create.js
```diff
@@ -163,21 +163,27 @@ const Create = () => {
...
+                        ?.addressLocality
                     }
                     terms={event?.terms}
                     flex={1}
                     imageUrl={event?.image}
-                    description={event?.description?.[i18n.language]}
+                    description={
+                      event?.description?.[i18n.language] ??
+                      event?.description?.[event.mainLanguage]
+                    }
```
geoffreydhuyvetters, line 186: I'm not sure what this code changes, looks the same to me but just more verbose?

src/i18n/fr.json
```diff
@@ -98,6 +98,51 @@
...
+    "Opendeurdag": "Journée portes ouvertes",
+    "Recreatiedomein of centrum": "Centre de loisirs ou centre de récréation",
+    "Park of tuin": "Jardin ou parc",
+    "Archeologische Site": "Site Archéologique",
+    "School of onderwijscentrum": "École ou centre éducatif",
+    "Sportcentrum": "Centre sportif",
+    "Winkel": "Magasin",
+    "Museum of galerij": "Musée ou galerie",
+    "Zaal of expohal": "Hall ou salle d'expo"
+  },
```
bertramakers, line 145: Nit: Maybe we should use the ids of the terms as keys here instead of their labels? Because the copy of the labels changes sometimes, and then we need to make sure we also change the copy here in the _key_ as well. While the ids stay the same.

bertramakers (approved):
Good spot! 🕵️‍♂️

src/i18n/fr.json
simon-debruijn: I'm going to implement it this way now 😄

src/i18n/fr.json
```diff
@@ -98,6 +98,49 @@
     "notifications": "Notifications",
     "time_ago": "Il y a {{time}}"
   },
+  "offerTypes": {
+    "0.0.0.0.0": "Exposition",
+    "0.12.0.0.0": "Journée portes ouvertes",
+    "0.14.0.0.0": "Monument",
+    "0.15.0.0.0": "Zone naturel ou parc",
```
simon-debruijn, line 105: @bertramakers I've changed the keys to the id's.

src/components/manage/productions/index/Events.js
```diff
@@ -52,6 +47,11 @@ const Event = ({
     setIsExpanded((prevValue) => !prevValue);
   };
 
+  const type = useMemo(() => {
+    const typeId = terms.find((term) => term.domain === 'eventtype')?.id ?? '';
+    return t(`offerTypes*${typeId}`, { keySeparator: '*' });
```
simon-debruijn, line 52: This was necessary  because the id's contain '.' which i18n uses as a keySeparator

bertramakers:
```suggestion
    // The custom keySeparator was necessary because the ids contain '.' which i18n uses as default keySeparator
    return t(`offerTypes*${typeId}`, { keySeparator: '*' });
```

bertramakers: Thanks for the explanation! Adding it to the code could help prevent confusion in the future.

bertramakers (approved):

src/components/manage/productions/create/Event.js
```diff
@@ -27,10 +27,10 @@ const Event = ({
 }) => {
   const { t, i18n } = useTranslation();
 
-  const type = useMemo(
-    () => terms.find((term) => term.domain === 'eventtype')?.label ?? '',
-    [terms],
-  );
+  const type = useMemo(() => {
+    const typeId = terms.find((term) => term.domain === 'eventtype')?.id ?? '';
+    return t(`offerTypes*${typeId}`, { keySeparator: '*' });
```
simon-debruijn, line 32:
```suggestion
    // The custom keySeparator was necessary because the ids contain '.' which i18n uses as default keySeparator
    return t(`offerTypes*${typeId}`, { keySeparator: '*' });
```

PR #163, opened 2021-01-18
Redirect when token is expired

geoffreydhuyvetters:
### Fixed

- Redirect to auth page when token is expired

---

Ticket: https://jira.uitdatabank.be/browse/III-3683

### DIFF
--- a/src/utils/getApplicationServerSideProps.js
+++ b/src/utils/getApplicationServerSideProps.js
@@ -1,5 +1,6 @@
 import { Cookies } from 'react-cookie';
 import { QueryClient } from 'react-query';
+import jwtDecode from 'jwt-decode';
 
 const getApplicationServerSideProps = (callbackFn) => async ({
   req,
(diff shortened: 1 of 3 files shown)


jdrieghe (approved):

src/utils/getApplicationServerSideProps.js
```diff
@@ -21,6 +22,17 @@ const getApplicationServerSideProps = (callbackFn) => async ({
...
+  const { exp } = jwtDecode(query?.jwt ?? cookies.token);
+
+  if (!exp || Date.now() >= exp * 1000) {
+    return {
+      redirect: {
+        destination: '/login',
+        permanent: false,
+      },
+    };
+  }
```
bertramakers, line 34: Does this check run on every re-render or only on the server-side render?
Because in the case of the latter, that would mean that if the token expires while the app is already open, the redirect to login won't work.
We used to do that when an API response was 401, but that doesn't work anymore since some changes to how we use react-query I think. Before this fix if you had an expired token, you could see in the network inspector that react-query kept trying to do API requests if it got a 401 instead of the app redirecting to the login page. And that caused the Auth0 login page to show in the iframe, because the Angular app did redirect when it got a 401.
I'm wondering if we shouldn't that underlying cause, in case we get a 401 for another reason than an expired token? Then we wouldn't need to manually check the expiration in the front-end either.

bertramakers: Ah we even already had an expiration check here:
https://github.com/cultuurnet/udb3-frontend/blob/main/api/api.js#L27-L37

Used here:
https://github.com/cultuurnet/udb3-frontend/blob/main/api/api.js#L52-L55

Should that one be removed then?

bertramakers: Here is where we did the redirect on 401, which stopped working: https://github.com/cultuurnet/udb3-frontend/blob/main/api/api.js#L41-L43

If we don't want to fix that, we should probably remove it?

PR #164, opened 2021-01-19
Improve handling of invalid tokens

geoffreydhuyvetters:
This handles the following situations

- invalid token when entering the page
- don't trigger a request on invalid token
- check every 5 seconds for invalid token and redirect if needed

Might be good to change the 400 to 401 on the back end when receiving a bad token. (also handled in the front-end)

### Added

- checking all parts of the token now (instead of only the payload)


src/utils/isTokenValid.js
```diff
@@ -0,0 +1,32 @@
+import base64 from 'base-64';
+
+const isTokenValid = (token) => {
+  if (!token) return false;
+  const keys = ['header', 'payload', 'signature'];
+
+  const parts = token.split('.');
+  if (parts.length !== 3) return false;
```
bertramakers, line 8: A JWT can also not have a signature in some cases, but never in our case. (So this is okay 👍) More of an FYI why the other library might not have complained about that.

bertramakers (approved):
"Update dependencies" 😜

simon-debruijn:
The code looks correct. I'm just wondering if decoding tokens on the client-side isn't overkill / bad practice.

simon-debruijn (approved):

geoffreydhuyvetters:
@simon-debruijn Why would it be a bad practice?

simon-debruijn:
> @simon-debruijn Why would it be a bad practice? It's pretty common afaik to access some data from the token in the client (that's also a decode)

That's also true. No I was just wondering if this is a common practice.

geoffreydhuyvetters:
> "Update dependencies" 😜

Oh, I just noticed it now. It changed my title again when switching branches 🙄

src/utils/isTokenValid.js
geoffreydhuyvetters: The other library only looked at the payload, but not the header. The header should also be valid JSON.

bertramakers:
> That's also true. No I was just wondering if this is a common practice.

Yeah it's pretty common. Some things to keep in mind when doing this:

- Because migrated users from uitid v1 have their old id in a different claim than `sub` (the standard claim for the id in Auth0), we have some logic how to determine the correct id. This is already implemented on the backend and when you request `/user` you don't need to pay attention to that. If you ever need to get the user id from the token, keep in mind that it's a bit more complex than just using the `sub`. https://github.com/cultuurnet/udb3-jwt/blob/master/src/Udb3Token.php#L46-L56 (`uid` is the claim in old long-lived tokens, `https://publiq.be/uitidv1id` is the id of migrated users, and `sub` is the id of new Auth0 users)
- While you can do a complete validation of the token on the front-end including the signature check using a public key, the back-end should obviously also do this validation. So my advice would be to do some basic validation on the frontend like checking the expiration etc like we do now, but there's little reason to try and validate it 100%. If someone forges a token and gets past the frontend validation the furthest they'd get is seeing a UI but no data since the backend won't return any data (for protected routes).
- If you want to be 100% sure the token is valid, in another project for example, it's better to use a 3rd party jwt package that can also validate the signature with the correct algorithm to prevent forging of tokens by other parties. You'll need to add a public key to your project's configuration to do that, but since that's public that's easy to get in Auth0 (just ask me/Kristof/... to show you how).

☝️ Just some general info, nothing blocking this PR.

geoffreydhuyvetters:
Can you validate a token without the secret key? I thought that was not possible, hence why we do it on the server.

bertramakers:
Yes, we used `RS256` before and also now on Auth0, which is an assymetric signing algorithm. That means that there's no shared secret, but instead a private key on the identity provider (Auth0 in this case) to _sign_ the tokens, and a public key on the server and optionally frontend to _validate_ the signature. You cannot sign tokens with the public key, so it's safe to expose. https://auth0.com/docs/tokens/signing-algorithms

> RS256 (RSA Signature with SHA-256): An asymmetric algorithm, which means that there are two keys: one public key and one private key that must be kept secret. Auth0 has the private key used to generate the signature, and the consumer of the JWT retrieves a public key from the metadata endpoints provided by Auth0 and uses it to validate the JWT signature.

The shared secret is when you're using `HS256` which is a symmetric algorithm which means it requires the secret both to sign and validate.

bertramakers:
For example, you can find the public key of our Auth0 tenant on production here: https://publiq.eu.auth0.com/.well-known/jwks.json

bertramakers:
Note that if you want to validate it (which is okay by me), the suggested approach is to fetch the key from that endpoint instead of bundling it in the app itself. That way we could easily do key rotation if our private key would ever become compromised and we need to generate a new private/public key pair.

bertramakers:
I'd have to check which is the correct one though, because there's two for some reason now 🤔

PR #169, opened 2021-01-22
Add error boundary

geoffreydhuyvetters:
### Added

- Add error state via boundary

Left out the link for now, we can add it back later, but I wanted to merge the mechanism first


src/pages/_app.js
```diff
@@ -123,9 +115,17 @@ ApplicationLayout.propTypes = {
 const Layout = ({ children }) => {
   const { asPath } = useRouter();
 
-  if (asPath.startsWith('/login')) return <LoginLayout>{children}</LoginLayout>;
-  if (asPath.startsWith('/404')) return children;
-  return <ApplicationLayout>{children}</ApplicationLayout>;
+  if (
+    asPath.startsWith('/login') ||
+    asPath.startsWith('/404') ||
+    asPath.startsWith('/error')
```
simon-debruijn, line 121:
```suggestion
    asPath.startsWith('/404')
```

simon-debruijn: @geoffreydhuyvetters Can you change this? the error page doesn't exist anymore so

bertramakers: FYI you can also explicitly request changes when reviewing, so merging becomes impossible until it's fixed 😄

PR #174, opened 2021-01-26
Regain focus on MouseOver

geoffreydhuyvetters:
### Fixed

- regain focus on mouse over on sidebar (need to do a setTimeout trick to make it work)

---

Ticket: https://jira.uitdatabank.be/browse/III-3727

### DIFF
--- a/src/components/SideBar.js
+++ b/src/components/SideBar.js
@@ -423,6 +423,7 @@ const SideBar = () => {
       zIndex={1998}
       padding={{ default: 2, s: 0 }}
       spacing={3}
+      onMouseOver={() => setTimeout(window.focus, 0)}
     >
       <Link
         justifyContent="center"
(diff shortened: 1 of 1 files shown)


src/components/SideBar.js
```diff
@@ -423,6 +423,7 @@ const SideBar = () => {
       zIndex={1998}
       padding={{ default: 2, s: 0 }}
       spacing={3}
+      onMouseOver={() => setTimeout(window.focus, 200)}
```
simon-debruijn, line 426: Why is there a focus of `window`? Why not `e.target` for example?

geoffreydhuyvetters: Because you don't need to put focus on the menu items, you need to focus the app window

simon-debruijn: Is that because the iframe is seen as a separate instance than window (rest of the app)?
ecause if I think about it e.target can be different elements.

So this is probably correct 😄

geoffreydhuyvetters: Jep, indeed. You're managing focus between 2 windows.

simon-debruijn (approved):

PR #176, opened 2021-01-26
Don't show giftbox menu in sidebar when not using Dutch

bertramakers:
### Added

- Added check on sidebar menu items to hide giftbox menu item if not using Dutch

***

Ticket: https://jira.uitdatabank.be/browse/III-3695

### DIFF
--- a/src/components/SideBar.js
+++ b/src/components/SideBar.js
@@ -32,7 +32,11 @@ const getValueForSideBar = getValueFromTheme('sideBar');
 const getValueForMenu = getValueFromTheme('menu');
 
 const MenuItem = memo(
-  ({ href, iconName, suffix, children: label, onClick }) => {
+  ({ href, iconName, suffix, children: label, onClick, visible = true }) => {
+    if (!visible) {
+      return null;
+    }
+
     const Component = href ? Link : Button;
 
     return (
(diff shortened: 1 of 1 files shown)


src/components/SideBar.js
```diff
@@ -200,14 +200,16 @@ const NotificationMenu = memo(
           <Badge>{countUnseenAnnouncements}</Badge>
         ),
         onClick: onClickAnnouncementsButton,
+        visible: i18n.language === 'nl',
```
bertramakers, line 203: @geoffreydhuyvetters @simon-debruijn is this okay with the `.filter()` or should we add some kind of `visible` prop on box and use that instead? I couldn't find one in Storybook so I assume it doesn't exist yet.

bertramakers: Ah, I think I can just add a prop to `MenuItem`

bertramakers: See https://github.com/cultuurnet/udb3-frontend/pull/176/commits/e977e6c5482d19d3cd424ff0b3277be6ade4b84b, but I can still revert that if the first solution was better.

simon-debruijn: I prefer the last solution

simon-debruijn (approved):

src/components/SideBar.js
bertramakers: Do you think it would be valuable to add the `visible` prop to `Box` while we're at it? cc @geoffreydhuyvetters ?

geoffreydhuyvetters: I was planning on doing that idd, makes sense to have it everywhere IMO.

geoffreydhuyvetters: We can merge this one as is

geoffreydhuyvetters (approved):

PR #177, opened 2021-01-26
III-3680 Add organizer name to details of suggested events

bertramakers:
### Added

- Added organizer name to details of suggested events

### Removed

- Removed mocking of suggested events in dev mode

---

Ticket: https://jira.uitdatabank.be/browse/III-3680


src/components/manage/productions/create/Event.js
```diff
@@ -59,6 +60,7 @@ const Event = ({
           <Text>
             {locationName} {locationCity}
           </Text>
+          {organizerName.length > 0 && <Text>{organizerName}</Text>}
```
bertramakers, line 63: If I didn't do this `organizerName.length > 0` check, extra spacing would be visible because the `Text` would still be rendered (with empty content) and it would still render the space around it.

bertramakers: Alternatively I could do this check inside `Text` itself, so it doesn't get rendered if the content is an empty string?

src/hooks/api/authenticated-query.js
```diff
@@ -134,43 +133,11 @@ const useAuthenticatedMutation = ({
   return useMutation(innerMutationFn, configuration);
 };
 
-const useAuthenticatedQuery = ({ mockData, ...options } = {}) => {
+const useAuthenticatedQuery = ({ ...options } = {}) => {
   if (!!options.req && !!options.queryClient && typeof window === 'undefined') {
     return prefetchAuthenticatedQuery(options);
   }
 
-  const [randomMockData, setRandomMockData] = useState();
```
simon-debruijn, line 142: Why remove mockData?

bertramakers: See the description in the commit:

> These can be "mocked" in the similar_events table instead so we can also test the actual API integration locally

bertramakers: I understand why it was added, but it's not necessary anymore because we're not getting those suggestions from an external API anymore. It's all in the UDB3 backend now (been for a while actually) and easy to add test data with a MySQL QUI app. And then you're testing the actual API call and not a mock.

src/components/manage/productions/create/Event.js
geoffreydhuyvetters: `{organizerName && <Text>{organizerName}</Text>}` is enough here I think

bertramakers: Okay I'll change that. We won't do the check in `Text` itself then? Because I think it's also an "issue" (minor) for other `Text`s in this component.

bertramakers: It only works with either `organizerName.length > 0` or `!!organizerName`. Not sure which is preferable. It looks like it doesn't automatically cast to boolean, so we need to either do an explicit comparison or cast to boolean ourselves.

simon-debruijn: `{!!organizerName && <Text>{organizerName}</Text>}`

This will fix the issue

src/hooks/api/authenticated-query.js
simon-debruijn: 👍🏻

simon-debruijn (approved):

src/components/manage/productions/create/Event.js
bertramakers: Yes that's what I said, but I don't know which is preferable? 😄

src/hooks/api/authenticated-query.js
bertramakers: @simon-debruijn @geoffreydhuyvetters if you want I can document how to do this exactly? Or is it clear enough?

src/components/manage/productions/create/Event.js
simon-debruijn: !!organizerName is more commonly used in React I think

PR #178, opened 2021-01-26
Buf/iii 3731 fix language iframe

simon-debruijn:
### Fixed

- ***


src/pages/[...params].js
```diff
@@ -43,7 +34,9 @@ const Fallback = () => {
     return `${publicRuntimeConfig.legacyAppUrl}${path}${queryString}`;
   }, [asPath, cookies.token, i18next.language]);
 
-  return <IFrame url={legacyPath} />;
+  return (
+    <Box as="iframe" src={legacyPath} width="100%" height="100vh" flex={1} />
+  );
```
Comment (bertramakers) [inline src/pages/[...params].js:39]: > We just found out we were constantly refreshing the angular app on navigation, should be fixed (and 10 times faster) now

Was the memo the cause of that? Curious!

src/pages/_app.js
```diff
@@ -80,13 +80,11 @@ const ApplicationLayout = ({ children }) => {
...
-      const url = new URL(
+      window.history.pushState(
+        undefined,
+        '',
         `${window.location.protocol}//${window.location.host}${path}`,
       );
-      const hasPage = url.searchParams.has('page');
-      router.push({ pathname: url.pathname, query: {} }, path, {
-        shallow: hasPage,
-      });
```
bertramakers, line 89: If we don't push this to the router, does the following scenario still work?

- Click on search in the menu
- Click on an event title to go to its detail page
- Click on search in the menu
=> Do you still go to the search page? Because that was a bug in the Vue app that we fixed by pushing to the router.

bertramakers: Ah I just tested it and you cannot even get to the detail page of a result now. In other words clicking a link in the iframe doesn't do anything anymore.

That's because we block the link from going through in the Angular app, which was to solve the flickering when we'd push the change to the (Vue) router.

bertramakers: See:

- https://github.com/cultuurnet/udb3-angular-app/pull/149
- https://github.com/cultuurnet/udb3-frontend/pull/5

bertramakers: You could remove that `e.preventDefault()` in the AngularJS app, but then you might still run into the original issue of the React app not realizing that it's not on the search page anymore, so when you click the search link it doesn't "refresh". But maybe the React router is different enough from the Vue router so that it's not an issue.

bertramakers: (I'm using the search page as an example, but it's also an issue on the dashboard, roles, labels, etc pages. Basically everywhere where you click a menu link, click a link in the iframe, then click the same menu link again.)

simon-debruijn: Yeah @bertramakers You're right. the first scenario doesn't work with this method

simon-debruijn: 🤔

src/hooks/useHandleEvent.js
```diff
@@ -0,0 +1,23 @@
+import { useEffect } from 'react';
+import { useIsClient } from './useIsClient';
+
+const EventTypes = {
+  NAVIGATE_PREVIOUS_PAGE: 'popstate',
+};
+
+const useHandleEvent = (eventsMap = {}) => {
```
simon-debruijn, line 8: This is a hook I added, we can use in the future, but isn't related to this PR specifically

bertramakers: Cool 👍

src/pages/_app.js
```diff
@@ -83,10 +83,17 @@ const ApplicationLayout = ({ children }) => {
...
       const hasPage = url.searchParams.has('page');
-      router.push({ pathname: url.pathname, query: {} }, path, {
-        shallow: hasPage,
-      });
+      if (hasPage) {
+        window.history.pushState(
+          undefined,
+          '',
+          `${window.location.protocol}//${window.location.host}${path}`,
+        );
```
bertramakers, line 93: Can you create a follow-up ticket to look into the following scenario?

- Click search in menu
- Go to page 2 of search results
- Click search in menu again => Nothing happens

I know the context from the call we did and we don't need to fix it in this PR, but it should be investigated if/how we can solve that.

bertramakers: It will also be an issue for other pages, like:

- Click roles in the menu
- Go to page 2 of roles
- Click roles in the menu again => Nothing happens

simon-debruijn: Isn't this just related to [this issue](https://jira.uitdatabank.be/browse/III-3727). Because the second click it works for me

simon-debruijn: Either way.

Here is a ticket to go over all possible routing scenario's, to then find the best possible solution: [III-3743](https://jira.uitdatabank.be/browse/III-3743)

bertramakers: Might be, let's get that fixed first then!

bertramakers (approved):

PR #179, opened 2021-01-27
Fix pull request template formatting

bertramakers:
### Fixed

- Fixed PR template formatting


bertramakers:
You can see the issue this causes in this PR by @simon-debruijn https://github.com/cultuurnet/udb3-frontend/pull/178

Screen Shot 2021-01-27 at 10 15 09

The line to separate the changelog from additional info is on the bullet point of "Fixed"

simon-debruijn (approved):

bertramakers:
Hmm ignoring the `.github` directory is harder than I thought, because we set the `--ignore-path` to `.gitignore` in `package.json`. Which makes sense since we don't want to format those files either. But I cannot add `.github` to`.gitignore` for obvious reasons. 

There are multiple issues open for this exact scenario:
- https://github.com/prettier/prettier/issues/8048
- https://github.com/prettier/prettier/issues/8506
- https://github.com/prettier/prettier/issues/8288

It seems the only option to have it ignore the `.github` directory at this point is to use `.prettierignore` and copy-paste everything from `.gitignore` into `.prettierignore`. But that is _a lot_ of duplicated info.

I'll see if I can just change the template so prettier doesn't complain about it or alter it.

simon-debruijn (approved):
🥇 🥇 🥇

PR #180, opened 2021-01-27
Remove deprecated react-query-devtools

simon-debruijn:
### Removed

- [deprecated react-query-devtools](https://github.com/tannerlinsley/react-query-devtools#readme)
---


bertramakers (approved):
Clear! 👍 
It's "deprecated" btw 😉

PR #181, opened 2021-01-27
Bug/iii 3727 only redirects after 2 clicks

simon-debruijn:
### Fixed

- only redirects after 2 clicks

---

Ticket: https://jira.uitdatabank.be/browse/III-3727

Should work on Chrome an d Firefox now 😄 !!!
We couldn't use window.focus because it depends on the browsersettings if we can use it or not.

[It may fail due to user settings and the window isn't guaranteed to be frontmost before this method returns.](https://developer.mozilla.org/en-us/docs/Web/API/Window/focus)

### DIFF
--- a/src/components/SideBar.js
+++ b/src/components/SideBar.js
@@ -418,6 +418,7 @@ const SideBar = () => {
   return [
     <Stack
       key="sidebar"
+      tabIndex={0}
       forwardedAs="nav"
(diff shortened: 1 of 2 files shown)


bertramakers (approved):
Works for me locally! 👏  Thanks for the extra explanation in the PR description! 🏅

PR #182, opened 2021-01-28
Improvement/iii 3746 add favicon

simon-debruijn:
### Added

- favicon
- title
- description
---

Ticket: https://jira.uitdatabank.be/browse/III-3746


bertramakers (approved):

src/pages/_app.js
```diff
@@ -135,6 +136,28 @@ Layout.propTypes = {
...
+        content="initial-scale=1.0, width=device-width"
+      />
+      <link
+        key="icon"
+        rel="icon"
+        type="image/png"
+        sizes="32x32"
+        href="/favicon.png"
+      />
+      <title key="title">UitDatabank</title>
```
bertramakers, line 155: Should we add a follow-up ticket / PR to add more info to this title, e.g. the title of the current page the user is on? (Might also need changes in AngularJS to push the title of the page to the parent window)

simon-debruijn: We can Improve this in the future, but I asked @corneelwille how important SEO was and we came to the conclusion to not spend too much time on it right now.

bertramakers: No of course but it's also just a nice UX improvement for users using multiple tabs (most probably?)

PR #183, opened 2021-01-28
Fix primary and secondary buttons to be like in Angular app

simon-debruijn:
### Changed

- styling primary button
- styling secondary button
---

You can easily verify by running `yarn storybook` on `udb3-frontend` and comparing it to [localhost:9999](http://localhost:9999) when running `grunt serve` on `udb3-angular-app`.

Ticket: https://jira.uitdatabank.be/browse/III-3741


bertramakers (approved):
👌

PR #185, opened 2021-02-03
Feature/III-3669 - Create routes/pages for place-status and event-status

simon-debruijn:
### Added

- api route/hook for place
- page route for /event/{eventId}/status
- page route for /place/{eventId}/status
- error state in status pages
- use SelectionTable component in event status page

### Fixed

- some typos in api hooks
- error's not being thrown in fetchWithApi

---

This PR focusses on having created the pages for place and event with their respective routes.

Screenshot event:
Schermafbeelding 2021-02-04 om 11 40 28

Screenshot place:
Schermafbeelding 2021-02-03 om 10 45 34

Screenshot error-state:
Schermafbeelding 2021-02-03 om 11 35 37

Ticket: https://jira.uitdatabank.be/browse/III-3669


src/components/EventStatus.js
```diff
@@ -0,0 +1,39 @@
+import PropTypes from 'prop-types';
+import { useTranslation } from 'react-i18next';
+import { Alert } from './publiq-ui/Alert';
+import { Button, ButtonVariants } from './publiq-ui/Button';
+import { Page } from './publiq-ui/Page';
+import { Spinner } from './publiq-ui/Spinner';
+
+const EventStatus = ({ offer, loading }) => {
```
simon-debruijn, line 8: OfferStatus would actually be more accurate right? But yeah for consistency's sake in frontend/backend I'll keep it named like this.

geoffreydhuyvetters:
Just a general remark. I think it makes sense to mimic the pages folder structure in components as much as we can

simon-debruijn:
New requirements have been added to the ticket, converting PR to DRAFT again.

src/components/EventStatus.js
bertramakers: It's not called `eventStatus` in the backend btw. We named it simply `status` for this reason. But Corneel keeps referring to it as "eventStatus" 😅 

So yes you can call this `offerStatus` (or just `status`), that would be much clearer I think!

src/pages/[...params].js
```diff
@@ -46,9 +46,7 @@ const Fallback = () => {
       '?',
     );
 
-    return `${
-      'http://localhost:9999' || publicRuntimeConfig.legacyAppUrl
-    }${path}${queryString}`;
+    return `${publicRuntimeConfig.legacyAppUrl}${path}${queryString}`;
```
Comment (bertramakers) [inline src/pages/[...params].js]: If you want to use `http://localhost:9999`, you can add `.env.local` to `source/udb3-frontend` with:
```
NEXT_PUBLIC_LEGACY_APP_URL=http://localhost:9999
```
That should override the `NEXT_PUBLIC_LEGACY_APP_URL` set in the `.env` that gets set by the initscripts.

geoffreydhuyvetters:
I would go for 2 separate pages and don't abstract it in a component.
There will always be changes later.

What do you think @bertramakers?

src/components/EventStatus.js
simon-debruijn: done 👍🏻

bertramakers:
> I would go for 2 separate page implementations without abstracting it in the components folder.
> There will always be changes later.
> 
> What do you think @bertramakers?

The biggest difference between events and places in context of status are the calendar types they can have and how that impacts the statuses.

Events can have four calendar types:

- *single*
  - has one date
  - has one subEvent, with a status 
  - has a status on the top level
  - status + reason can be updated on top level through the API, subEvent will then get the same status + reason
  - status + reason can also be updated on subEvent level through the API, top level will then get the same status + reason
- *multiple*
  - has multiple dates
  - has multiple subEvents, with a status each
  - has a status on the top level
  - status + reason can be updated on top level through the API, subEvents (all!) will then get the same status + reason
  - status + reason can also be updated on subEvent level through the API, top level will then get a status derived from the subEvents (there's some logic for this, for example if one of the subEvents is still "Available" but the other are "Unavailable", the top status will still be "Available" - Corneel has more info about this if you need to know the exact logic, but it's always determined by the backend already)
- *periodic*
  - has start- and end date for a longer period of time (spanning weeks, months, ...)
  - can have optional opening hours -> these have no status
  - does not have subEvents
  - has a status on the top level
  - status + reason can be updated on top level through the API
- *permanent* (no start/end, optionally with "opening" hours, and a status on the top level)
  - has no start- and end date
  - can have optional opening hours -> these have no status
  - does not have subEvents
  - has a status on the top level
  - status + reason can be updated on top level through the API

Places van have two calendar types:

- *periodic* (same as for events)
- *permanent* (same as for events)

I think the difference in the UI won't be the offer type (event or place), but the calendar type. The wireframes @corneelwille showed are clearly for calendar type *multiple* (i.e. only events), where you select one or more subEvents and update their status. However for *single*, *periodic*, and *permanent* the only functionality is updating the 1 status they have so the UI we went over in the call does not really fit that use case. 

A UI for those calendar types could be as simple as having the form we'd otherwise put in the modal I think:
- Choice between 3 status types (radio buttons)
- Optional reason field
- Save button

So my advice would be to split up the components per calendar type. (Although for single, periodic and permanent they would be the same, so splitting those up might also be overkill 🤔)

I'm not sure I see the need to split up the pages themselves because they will just show the relevant "calendar type component" then?

simon-debruijn:
@bertramakers 

`I'm not sure I see the need to split up the pages themselves because they will just show the relevant "calendar type component" then?`

If you mean why we have 2 different status pages, one for `place` and one for `event`. That is because of the way Next.js's [dynamic routing](https://nextjs.org/docs/routing/dynamic-routes) works.

bertramakers:
> @bertramakers
> 
> `I'm not sure I see the need to split up the pages themselves because they will just show the relevant "calendar type component" then?`
> 
> If you mean why we have 2 different status pages, one for `place` and one for `event`. That is because of the way Next.js's [dynamic routing](https://nextjs.org/docs/routing/dynamic-routes) works.

That was a reply to `I would go for 2 separate page implementations without abstracting it in the components folder.`

bertramakers:
Ah, another difference between events & places will be the copy.
For example the status `TemporarilyUnavailable` on event is "Uitgesteld", and on place it's "Temporarily closed".
Not sure how easy/hard it would be to have different copy in the components depending on the offer type.

simon-debruijn:
Can we merge this one? I am already working on a [follow-up PR](https://github.com/cultuurnet/udb3-frontend/pull/186/files). 
There I add a SelectionTable component and I have removed the OfferStatus component and there only remain both pages.

src/pages/event/[eventId]/status.js
```diff
@@ -0,0 +1,108 @@
...
+          <Alert variant={AlertVariants.WARNING}>{errorMessage}</Alert>
+        ) : (
+          [
+            <Alert key="alert">{t('offerStatus.info')}</Alert>,
+            <SelectionTable
+              key="table"
+              columns={columns}
+              data={data}
+              onSelectionChanged={setSelectedOffers}
+            />,
```
Comment (bertramakers) [inline src/pages/event/[eventId]/status.js:91]: Only events with calendar type `multiple` will have this `SelectionTable`. While we can of course keep this and add a check later to show the simplified form for `single`, `periodic` and `permanent`, I'm wondering if it wouldn't be better to leave the table out for now if this PR is just about routing anyway?

src/pages/event/[eventId]/status.js
```diff
@@ -0,0 +1,108 @@
...
+    ],
+    [],
+  );
+
+  const data = useMemo(
+    () =>
+      subEvents.map((subEvent) => ({
+        time: 'Dinsdag 13 Februari 2021',
+        status:
+          OfferStatusMap[offerType]?.[camelCase(subEvent.status.type)] ?? '',
```
Comment (bertramakers) [inline src/pages/event/[eventId]/status.js:70]: This shows the top status, but in the table we should be showing the status of the subEvents (for calendar type multiple). So continuing on my other comment, I'd suggest to just leave this table out of this PR and add a new PR later for the UI for events with calendar type multiple specifically.

src/pages/event/[eventId]/status.js
Comment (simon-debruijn) [reply src/pages/event/[eventId]/status.js:91]: Okay I'll leave the use of the SelectionTable out of this PR

src/pages/event/[eventId]/status.js
Comment (simon-debruijn) [reply src/pages/event/[eventId]/status.js:70]: Okay I'll leave the use of the SelectionTable out of this PR

bertramakers (approved):
Looks good! 👍  I think we can even merge this into `main` now.

PR #191, opened 2021-02-05
Feature/III-3576 place status page

simon-debruijn:
### Added

- form for place status
- api call /places/uuid/status
- TextArea component

### Fixed

- configs not being passed to useAuthenticatedMutation
- fetchFromApi throwing error on 204 (No content)

---

Screenshots:

Schermafbeelding 2021-02-08 om 13 59 42
Schermafbeelding 2021-02-08 om 13 59 59

Ticket: https://jira.uitdatabank.be/browse/III-3576


src/pages/place/[placeId]/status.js
```diff
@@ -1,48 +1,148 @@
...
-  const { data: place = {}, status } = useGetPlaceById(
+  const handleError = (error) => setErrorMessage(error.message);
+  const { data: place = {}, status: queryStatus } = useGetPlaceById(
     { id: placeId },
     { onError: handleError },
   );
+  const handleSuccessChangeStatus = () => {
+    if (isClient) {
+      router.push(`/place/${placeId}/preview`);
+    }
```
Comment (geoffreydhuyvetters) [inline src/pages/place/[placeId]/status.js:47]: Is the isClient check needed here? You will never trigger this handler from the server

geoffreydhuyvetters (commented):
The only thing that's missing atm is the prefetching on the server, have a look at productions on how to implement that.

src/pages/place/[placeId]/status.js
```diff
@@ -1,48 +1,148 @@
...
+  const [reason, setReason] = useState('');
 
-  const handleError = (error) => {
-    setErrorMessage(error.message);
-  };
-  const { data: place = {}, status } = useGetPlaceById(
+  const handleError = (error) => setErrorMessage(error.message);
+  const { data: place = {}, status: queryStatus } = useGetPlaceById(
     { id: placeId },
     { onError: handleError },
```
Comment (geoffreydhuyvetters) [inline src/pages/place/[placeId]/status.js:42]: no need to use an onError handler here, you can switch the UI into error mode by using status, and the actual error is returned

src/pages/place/[placeId]/status.js
```diff
@@ -1,48 +1,148 @@
...
   );
+  const handleSuccessChangeStatus = () => {
+    if (isClient) {
+      router.push(`/place/${placeId}/preview`);
+    }
+  };
+  const { mutate: changeStatus } = useChangeStatus({
+    onSuccess: handleSuccessChangeStatus,
+    onError: handleError,
```
Comment (geoffreydhuyvetters) [inline src/pages/place/[placeId]/status.js:51]: Same here, use status and error

src/components/publiq-ui/RadioButtonWithLabel.js
```diff
@@ -7,29 +7,48 @@ import {
 import { Inline } from './Inline';
 import { Label } from './Label';
 import { getBoxProps, boxPropTypes } from './Box';
+import { Stack } from './Stack';
+import { Text } from './Text';
+import { getValueFromTheme } from '../../components/publiq-ui/theme';
```
LucWollants, line 12: I think here it's also possible to do `from './theme`

src/components/publiq-ui/TextArea.js
```diff
@@ -0,0 +1,34 @@
+import PropTypes from 'prop-types';
+import { Box, boxPropTypes, getBoxProps } from './Box';
+import { getValueFromTheme } from './theme';
+
+const getValue = getValueFromTheme('textArea');
+
+const TextArea = ({ id, children, className, onInput, value, ...props }) => {
```
LucWollants, line 7: The prop `children` is not used so maybe it's not needed?

LucWollants:
There is a small typo inside `authenticated-query.js` on [line 71 `perparedArguments`](https://github.com/cultuurnet/udb3-frontend/pull/191/files#diff-73178666f3f200c749ac486b737bdb3d0b99450ebea802cde9b96a3231a5281aR71) instead of `preparedArguments`

src/pages/place/[placeId]/status.js
```diff
@@ -1,55 +1,173 @@
...
+import { useChangeStatus, useGetPlaceById } from '../../../hooks/api/places';
+import { Text } from '../../../components/publiq-ui/Text';
+import { getValueFromTheme } from '../../../components/publiq-ui/theme';
+import { Stack } from '../../../components/publiq-ui/Stack';
+import { Inline } from '../../../components/publiq-ui/Inline';
+import { dehydrate } from 'react-query/hydration';
+
+const getValue = getValueFromTheme('statusPage');
+
+const OfferStatus = {
```
Comment (LucWollants) [inline src/pages/place/[placeId]/status.js:21]: Some constants to put in a more general file and to reuse?

Comment (simon-debruijn) [reply src/pages/place/[placeId]/status.js:21]: In the following PR I have moved OfferStatus and maxReasonLength to EventStatusModal and import them from there. But I'm also not sure about that approach. The thing is we've tried so far to define constants in logical places and tried to avoid putting them in a constants directory.

Comment (LucWollants) [reply src/pages/place/[placeId]/status.js:21]: The logical place is fine, as long as it gets reused.

LucWollants:
When I visit the status page and run the frontend in development mode with `yarn dev`, I get 2 JS warnings:
- Warning: Invalid value for prop `color` on <nav> tag. Either remove it from the element, or pass a string or number value to keep it in the DOM. For details, see https://reactjs.org/link/attribute-behavior
- Warning: Invalid value for prop `_css` on <div> tag. Either remove it from the element, or pass a string or number value to keep it in the DOM. For details, see https://reactjs.org/link/attribute-behavior

Most likely this is not related to a change in the PR but I like an empty console at all times. This makes it much easier to spot warnings or errors introduced with new code or changes.

simon-debruijn:
> When I visit the status page and run the frontend in development mode with `yarn dev`, I get 2 JS warnings:
> 
>     * Warning: Invalid value for prop `color` on  tag. Either remove it from the element, or pass a string or number value to keep it in the DOM. For details, see https://reactjs.org/link/attribute-behavior
> 
>     * Warning: Invalid value for prop `_css` on  tag. Either remove it from the element, or pass a string or number value to keep it in the DOM. For details, see https://reactjs.org/link/attribute-behavior
> 
> 
> Most likely this is not related to a change in the PR but I like an empty console at all times. This makes it much easier to spot warnings or errors introduced with new code or changes.

This is due to our spacing/styling system in place where we add custom props like color to our components and they get set on the dom element itself. I'm afraid we weren't able to fix these warnings. But we could look into it again some time.
@geoffreydhuyvetters right?

LucWollants:
I also noticed an issue when clicking "Bewaar" and the reason was not filled in.
The body that gets send to the status endpoint is
```
{"type":"Available","reason":{"nl":""}}
```
The reason property should not be passed in this case

simon-debruijn:
> I also noticed an issue when clicking "Bewaar" and the reason was not filled in.
> The body that gets send to the status endpoint is
> 
> ```
> {"type":"Available","reason":{"nl":""}}
> ```
> 
> The reason property should not be passed in this case

I also noticed this and fixed this in the following PR, but I've pushed the change here now

LucWollants:
I am not sure what the policy is for frontend tests, but I think this can use some tests?

src/pages/place/[placeId]/status.js
```diff
@@ -1,55 +1,176 @@
...
 
-  const handleError = (error) => {
-    setErrorMessage(error.message);
-  };
-  const { data: place = {}, status } = useGetPlaceById(
-    { id: placeId },
-    { onError: handleError },
-  );
   const name = place.name?.[i18n.language] ?? place.name?.[place.mainLanguage];
+  const rawStatusType = place?.status?.type;
```
Comment (LucWollants) [inline src/pages/place/[placeId]/status.js:55]: What do you mean by `raw`?

Comment (simon-debruijn) [reply src/pages/place/[placeId]/status.js:55]: as in `raw data from the api query`, these values get set as state values in the component.
e.g. type gets set to the value of rawStatusType

simon-debruijn:
> I am not sure what the policy is for frontend tests, but I think this can use some tests?

Yes I will add a test covering the place/placeid/status page and it's scenario's

src/pages/place/[placeId]/status.js
Comment (LucWollants) [reply src/pages/place/[placeId]/status.js:55]: Ok, clear. Thanks for explaining. Is `original` then a better prefix?

Comment (simon-debruijn) [reply src/pages/place/[placeId]/status.js:56]: I'm going to keep the raw naming for now because we use this in several other places

LucWollants:
> Yes I will add a test covering the place/placeid/status page and it's scenario's

As discussed through Slack this will be done in another PR.

LucWollants (approved):

PR #193, opened 2021-02-09
Feature/iii 3758 modal offer status

simon-debruijn:
This PR is about the EventStatusModal component, not the use of the component nor the api calls made for the subEvents

### Added

- EventStatusModal

### Changed

- Several components to add functionality

---

Ticket: https://jira.uitdatabank.be/browse/III-3758

Screenshots:
Schermafbeelding 2021-02-22 om 09 26 45
Schermafbeelding 2021-02-22 om 09 26 58
Schermafbeelding 2021-02-22 om 09 27 22


src/components/event/eventId/status/EventStatusModal.js
```diff
@@ -0,0 +1,104 @@
...
+  TEMPORARILY_UNAVAILABLE: 'TemporarilyUnavailable',
+  UNAVAILABLE: 'Unavailable',
+};
+
+const maxLengthReason = 200;
+
+const getValue = getValueFromTheme('statusModal');
+
+const EventStatusModal = ({
+  subEvents,
```
LucWollants, line 23: For now `subEvents` are not used

src/components/event/eventId/status/EventStatusModal.js
```diff
@@ -0,0 +1,104 @@
...
+      variant={ModalVariants.QUESTION}
+      size="xl"
+      className={className}
+      confirmTitle={t('offerStatus.actions.save')}
+      cancelTitle={t('offerStatus.actions.close')}
+      onConfirm={() => changeStatuses()}
+      onClose={onClose}
+      confirmButtonDisabled={reason.length > maxLengthReason}
+    >
+      <Stack padding={4} spacing={5}>
```
LucWollants, line 47: Can this become a separate component? So that it can be reused for places and events?

simon-debruijn: In places this won't be used, because there it is a seperate page and not a modal

src/components/event/eventId/status/EventStatusModal.js
simon-debruijn: Yep, this modal will be implemented in the following PR, where all the logic will be implemented to change the satus of multiple subEvents

simon-debruijn:
We're continuing development on the OfferStatus development without tests for now.

simon-debruijn:
If you want to see the modal locally, follow these steps:

You can go to this page /event/[eventId]/status
under `<Alert key="alert">{t('offerStatus.info')}</Alert>`,
Add this code:
```
<EventStatusModal
  key="statusmodal"
  visible
  subEvents={[]}
  onClose={() => {}}
  onConfirm={() => {}}
/>, 
```

src/components/event/eventId/status/EventStatusModal.js
bertramakers: But the form inside the modal is the same as the form we show on the page for places, no?

simon-debruijn: Yes that's true, apart from the buttons, that live in the modal itself here and are part of the page form itself.

But in one of the previous PR's where I started the work on offerStatus I first had 1 component `OfferStatus`. But then the advice was to split them up because:
1) they would possibly be a bit different (now seems to be quite similar for that specific form that's reused)
2) like this we have the path for the component 1 on 1 equal to the path of the page it's used on, whereas if you have a seperate component it can't live under a specific path like `src/components/event/eventId/status/EventStatusModal.js `

But I agree that this creates duplicate code that does the same thing

bertramakers: Yep, I remember that PR and the 1 component. I personally didn't see an issue in having 1 component and tried to explain that I think. The argument was that it would be different for events and places but that's not really correct. There's no difference between events and places in this case, only some differences between calendar types. On the one hand there are `single`, `periodic`, and `permanent`. Those can be on events or places, and will all have the form on the page. Then there's `multiple`, which is only on events and will use the form in the modal with the selection table. So instead of differentiating per offer (having a page/component for event, and different a page/component for place), I would advise differentiating per calendar type instead.

bertramakers: So to give an example in pseudo-code:

```
switch (offer.calendarType) {
    case `single`:
    case `periodic`:
    case `permanent`:
        return <ComponentWithStatusFormOnPage />

    case `multiple`:
        return <ComponentWithSelectionTableAndStatusFormInModal />
}
```

And both use a shared `StatusForm` either on the page or in the modal (name is just an example)

simon-debruijn: I'll try to split them up by calendar type then, and the components folder will look something like this then:

```
> components
    > status
       > Single
       > Periodic
       > Permanent
       > Multiple
```

Does this mach what you're thinking @bertramakers?

bertramakers: Yes but no 😄 `single`, `periodic` and `permanent` will be the same because they have no subEvents so you always set the status on the "top level" of the offer. While for `multiple` there are subEvents, so you need the selection table + modal and set the status on specific subEvents only.

So a better differentiation would probably be `OfferTopStatus` and `OfferSubEventStatus` or something like that now that I think about it in terms of how the code will be stored in directories etc.

simon-debruijn: I'll give it a crack 😄

simon-debruijn: @bertramakers The forms are different though! The translations used for the radiobuttons are different. Also the api calls used are different.

The component will have more logic inside to show the correct information and do the correct api calls (place vs event)

bertramakers: Yeah, there will be some differences but there are ways to make components generic of course. For example for the API callbacks, you can use a callback prop like an `onSave()` instead of having the form do the API call. It's hard to make a decision on whether it's best to make a shared component or make separate components without diving into the actual code though. The translations also seem easy to tackle, for example also passing them as props. If there's more separated logic it might be less interesting to have a generic component, but it's hard to say without more concrete information. It will always be a trade-off between trying to fit too many differences into a generic component vs accidentally having 2 different behaviors for events and places where there should be none because it's two separate components and they have slight accidental differences (now or later after a number of changes), causing UX issues.

simon-debruijn: Refactored 👍🏻

src/components/offerStatus/StatusFormOnPage.js
```diff
@@ -0,0 +1,120 @@
...
+                    reason:
+                      type === OfferStatus.AVAILABLE
+                        ? undefined
+                        : {
+                            ...offer.status.reason,
+                            [i18n.language]: reason,
+                          },
+                  });
+                }}
+              >
```
bertramakers, line 104: Maybe best to check with @corneelwille , but I have a feeling the save button should go first in this case.

Screen Shot 2021-02-22 at 17 06 58

In modals the primary button is on the right side, but the buttons are also aligned to the right there. Here they are aligned to the left. I don't think we have an explicit rule about this, but looking at other apps it seems like the primary button is always the first when aligned left and last when aligned right.

src/components/offerStatus/StatusFormOnPage.js
```diff
@@ -0,0 +1,120 @@
...
+                {t('offerStatus.actions.cancel')}
+              </Button>
+              <Button
+                variant={ButtonVariants.PRIMARY}
+                disabled={
+                  !offer ||
+                  reason.length === 0 ||
+                  reason.length > MaxLengthReason
+                }
```
bertramakers, line 90: I think this is incorrect, the reason is always optional. So you should be able to set the status to `Unavailable` or `TemporarilyUnavailable` without a reason.

src/components/offerStatus/StatusFormOnPage.js
bertramakers: Also, can we set a max-width on the form? The optimal line length of text is usually between 50-75 characters per line. (For example https://baymard.com/blog/line-length-readability)

As you can see in the screenshot, the text gets very wide on bigger screens and harder to read.

src/components/offerStatus/ReasonAndTypeForm.js
```diff
@@ -0,0 +1,102 @@
...
+}) => {
+  const { t } = useTranslation();
+
+  const radioButtonItems = useMemo(
+    () => [
+      {
+        label:
+          offerType === OfferType.PLACE
+            ? t('offerStatus.status.open')
+            : t('offerStatus.status.scheduled'),
```
bertramakers, line 30: You could remove these `OfferType.PLACE ? t('...') : t('offerStatus.status.scheduled')` by using translation keys like:

```
{
    offerStatus: {
        event: {
            available: "Gaat door",
            temporarilyUnavailable: "Uitgesteld",
            unavailable: "Geannuleerd"
        },
        place: {
            available: "Open",
            temporarilyUnavailable: "Tijdelijk gesloten",
            unavailable: "Permanent gesloten"
        }
    }
}
```

And then calling it like `t('offerStatus.' + offerType + '.available')`

src/components/offerStatus/StatusFormOnPage.js
bertramakers: Additionally, it should also be enabled when you pick "Available" (Gaat door/Open), because you need to be able to change the status back to that from Unavailable or TemporarilyUnavailable.

So I think a better would be something like the selected type not being the current type on the offer.

src/components/offerStatus/StatusModal.js
```diff
@@ -0,0 +1,50 @@
...
+      variant={ModalVariants.QUESTION}
+      size="xl"
+      className={className}
+      confirmTitle={t('offerStatus.actions.save')}
+      cancelTitle={t('offerStatus.actions.close')}
+      onConfirm={() => {}}
+      onClose={onClose}
+      confirmButtonDisabled={
+        type !== OfferStatus.AVAILABLE &&
+        (reason.length === 0 || reason.length > MaxLengthReason)
```
bertramakers, line 28: Same comment here about the disabled state not being correct:
- Reason is always optional
- You also need to be able to set the status back to "Available" if it's "Unavailable" or "TemporarilyUnavailable"

src/components/publiq-ui/Modal/ContentModal.js
```diff
@@ -53,11 +54,13 @@ ContentModal.propTypes = {
   onShow: PropTypes.func,
   onClose: PropTypes.func,
   children: PropTypes.node,
+  size: PropTypes.string,
 };
 
 ContentModal.defaultProps = {
   visible: false,
   title: '',
+  size: 'xl',
```
bertramakers, line 63: Is it possible to use the `ModalSizes.XL` constant here instead?

src/components/publiq-ui/Modal/QuestionModal.js
```diff
@@ -83,9 +84,7 @@ QuestionModal.defaultProps = {
   title: '',
   confirmTitle: 'Ok',
   cancelTitle: 'Cancel',
-  onShow: () => {},
-  onClose: () => {},
-  onConfirm: () => {},
+  size: 'sm',
```
bertramakers, line 87: Same question if we can use `ModalSizes.SM` here

src/pages/event/[eventId]/status.js
```diff
@@ -51,6 +52,18 @@ const Status = () => {
...
+    await useGetEventById({ req, queryClient, id: eventId });
+
+    return {
+      props: {
+        dehydratedState: dehydrate(queryClient),
+        cookies,
+      },
+    };
+  },
+);
```
Comment (bertramakers) [inline src/pages/event/[eventId]/status.js:67]: Can't comment on the `<Page>` component in this file because it wasn't changed, but: Events can also have the `StatusFormOnPage` scenario if their calendar type is `single`, `periodic` or `permanent`.

bertramakers (changes requested):

src/components/publiq-ui/Modal/ContentModal.js
simon-debruijn: Because the ModalSize lives in Another component `Modal`, It would throw error's that the constant was used before it was defined

src/components/publiq-ui/Modal/QuestionModal.js
simon-debruijn: Because the ModalSize lives in Another component `Modal`, It would throw error's that the constant was used before it was defined

src/components/offerStatus/StatusFormOnPage.js
simon-debruijn: in ticket: [III-3576](https://jira.uitdatabank.be/browse/III-3576)
this is written though `when a user chooses to update the status of his event to `available`, no reason can be added`

src/components/offerStatus/StatusFormOnPage.js
simon-debruijn: done

src/components/offerStatus/StatusFormOnPage.js
bertramakers: Yes, that means the reason field should be disabled which is implemented correctly. My comment is about the save button.

bertramakers: I meant that the save button should not be disabled just because the reason is empty.

simon-debruijn: I think the behaviour now is as it should be then

bertramakers: No because now the save button is disabled if you don't enter a reason? The reason should always be optional

bertramakers: Ah you updated the disabled check, looks good now 👍

src/pages/event/[eventId]/status.js
Comment (bertramakers) [reply src/pages/event/[eventId]/status.js:67]: Maybe better to do this in a follow-up PR as part of III-3576

bertramakers (approved):
Nice! 👍

PR #194, opened 2021-02-16
Feature/iii 3576 write tests

simon-debruijn:
### Added

- [Added ...]

### Changed

- [Changed ...]

### Removed

- [Removed ...]

### Fixed

- [Fixed ...]

---


src/tests/pages/place/[placeId]/status.test.js
```diff
@@ -0,0 +1,76 @@
+import { place as mockedPlace, place } from '../../../data/place';
```
Comment (simon-debruijn) [inline src/tests/pages/place/[placeId]/status.test.js:1]:
```suggestion
import { place as mockedPlace } from '../../../data/place';
```

src/tests/pages/place/[placeId]/status.test.js
```diff
@@ -0,0 +1,76 @@
...
+
+import { renderPageWithWrapper } from '../../../utils/renderPageWithWrapper';
+import { waitForFetch } from '../../../utils/waitForFetch';
+
+describe('Status page place', () => {
+  describe('When the status "Available" is loaded from place object', () => {
+    let router;
+
+    beforeEach(async () => {
+      router = setupPage({
```
Comment (simon-debruijn) [inline src/tests/pages/place/[placeId]/status.test.js:19]: maybe we should separate setupPage into setupRouter which returns the router object and setupPage or a different name that handles the rest

Comment (simon-debruijn) [reply src/tests/pages/place/[placeId]/status.test.js:19]: It would be more logical when the router object is returned from a function called `setupRouter` or `createMockRouter`

Comment (geoffreydhuyvetters) [reply src/tests/pages/place/[placeId]/status.test.js:19]: Yeah but setupPage does more than that. 
If you split it up, you need to remember to do all the steps every time.

src/tests/pages/place/[placeId]/status.test.js
Comment (geoffreydhuyvetters) [reply src/tests/pages/place/[placeId]/status.test.js:1]: I'm fixing this in a separate commit since it's not mockedPlace we're using everywhere

src/tests/pages/place/[placeId]/status.test.js
Comment (geoffreydhuyvetters) [reply src/tests/pages/place/[placeId]/status.test.js:19]: I could return an object that contains the router?

Comment (simon-debruijn) [reply src/tests/pages/place/[placeId]/status.test.js:19]: and doesn't `createMockRouter` fit the function `setupPage`

Comment (geoffreydhuyvetters) [reply src/tests/pages/place/[placeId]/status.test.js]: it also sets the mockresponses

Comment (simon-debruijn) [reply src/tests/pages/place/[placeId]/status.test.js]: Yep like this approach 👍🏻

geoffreydhuyvetters (approved):

PR #200, opened 2021-03-03
Feature/iii 3577 multiple type 2

simon-debruijn:
### Added

- layout for multiple type
- api call for changing status of subEvents

### Changed

- SelectionTable
- StatusModal

---

Ticket: https://jira.uitdatabank.be/browse/III-3577

Screenshot:
Schermafbeelding 2021-02-25 om 16 01 31

### DIFF
--- a/src/components/StatusPageMultiple.js
+++ b/src/components/StatusPageMultiple.js
@@ -125,7 +125,7 @@ const StatusPageMultiple = ({ event, refetchEvent }) => {
           onSelectionChanged={setSelectedRows}
           actions={[
             {
-              iconName: Icons.PENCIL_ALT,
+              iconName: Icons.PENCIL,
               title: t('offerStatus.changeStatus'),
               onClick: () => setIsModalVisible(true),
               disabled: selectedRows.length === 0,
--- a/src/components/StatusPageMultiple.js
+++ b/src/components/StatusPageMultiple.js
@@ -132,7 +132,11 @@ const StatusPageMultiple = ({ event, refetchEvent }) => {
             },
           ]}
         />
-        <Link href="/dashboard" variant={LinkVariants.UNSTYLED} customChildren>
+        <Link
+          href={`/event/${eventId}/preview`}
+          variant={LinkVariants.UNSTYLED}
+          customChildren
+        >
           <Button variant={ButtonVariants.SUCCESS}>
             {t('offerStatus.modificationReady')}
           </Button>
--- a/src/components/StatusPageMultiple.js
+++ b/src/components/StatusPageMultiple.js
@@ -131,6 +131,11 @@ const StatusPageMultiple = ({ event, refetchEvent }) => {
               disabled: selectedRows.length === 0,
             },
           ]}
+          translateSelectedRowCount={(count) =>
+            t('selectionTable.rowsSelectedCount', {
+              count,
+            })
+          }
         />
         <Link
           href={`/event/${eventId}/preview`}
--- a/src/components/StatusPageMultiple.js
+++ b/src/components/StatusPageMultiple.js
@@ -1,0 +1,156 @@
+import PropTypes from 'prop-types';
+import { useMemo, useState } from 'react';
+import { useTranslation } from 'react-i18next';
+import { useChangeStatusSubEvents } from '@/hooks/api/events';
+import { formatPeriod } from '@/utils/formatPeriod';
+import { parseOfferId } from '@/utils/parseOfferId';
+import { Alert } from '@/ui/Alert';
+import { Page } from '@/ui/Page';
+import { SelectionTable } from '@/ui/SelectionTable';
+import { Stack } from '@/ui/Stack';
+import { Text } from '@/ui/Text';
+import { getValueFromTheme } from '@/ui/theme';
+import { StatusModal } from './StatusModal';
+import { QueryStatus } from '@/hooks/api/authenticated-query';
+import { camelCase } from 'lodash';
+import { Icons } from '@/ui/Icon';
+import { Button, ButtonVariants } from '@/ui/Button';
+import { Link, LinkVariants } from '@/ui/Link';
+import { OfferStatus } from '@/constants/OfferStatus';
+
+const getValue = getValueFromTheme('statusPage');
+
+const Status = ({ type, reason }) => {
+  const { i18n } = useTranslation();
+  return (
+    <Stack>
+      <Text>{type}</Text>
+      {!!reason?.[i18n.language] && (
+        <Text color={getValue('infoTextColor')}>{reason[i18n.language]}</Text>
+      )}
+    </Stack>
+  );
+};
+
+Status.propTypes = {
+  type: PropTypes.string.isRequired,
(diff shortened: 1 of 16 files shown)


src/components/StatusPageMultiple.js
```diff
@@ -0,0 +1,156 @@
...
+              `offerStatus.status.event.${camelCase(subEvent.status.type)}`,
+            )}
+            reason={subEvent.status.reason}
+          />
+        ),
+      })),
+    [subEvents],
+  );
+
+  return [
```
bertramakers, line 117: Is this square brackets syntax new? I always remember it being `return (...);` when returning JSX.

src/components/StatusPageNotMultiple.js
```diff
@@ -14,7 +14,7 @@ import { ReasonAndTypeForm } from '@/components/ReasonAndTypeForm';
 
 import { OfferStatus } from '@/constants/OfferStatus';
 
-const StatusFormOnPage = ({ offer, error, useChangeStatus }) => {
+const StatusPageNotMultiple = ({ offer, error, useChangeStatus }) => {
```
bertramakers, line 17: So... single? 😛 

I think I get the reasoning, because it's also the page for `periodic` and `permanent`. But in this context we can maybe just use `single` since it's a "single status"?

src/ui/Icon.js
```diff
@@ -80,6 +82,7 @@ const IconsMap = {
   [Icons.TRASH]: faTrash,
   [Icons.BINOCULARS]: faBinoculars,
   [Icons.EXCLAMATION_TRIANGLE]: faExclamationTriangle,
+  [Icons.PENCIL_ALT]: faPencilAlt,
```
bertramakers, line 85: Nit: I'm wondering if we should really keep the `_ALT` in the name? On the one hand it makes it predictable as long as we keep using FontAwesome, but if we ever switch to another icon set later it might not make sense anymore 🤔 And it would be cool if we could just switch the icons here then and not have to search & replace all usages. Because then it kind of defeats the purpose of abstracting it away in our own icon component.

bertramakers: If we change it we might also want to change it for `SIGN_OUT_ALT`

src/ui/SelectionTable.js
```diff
@@ -61,69 +72,120 @@ const SelectionTable = ({ columns, data, onSelectionChanged, ...props }) => {
     ]);
   });
 
+  const { t } = useTranslation();
```
bertramakers, line 75: I agree with @geoffreydhuyvetters 's comment on the original PR that we should not use the translations hook in the component library, but instead pass them down as props. Can either be strings in the case of simple translations, or functions in the case of translations like here with the dynamic count. So for example a `renderSelectedRowCountText` function prop that you can call like `renderSelectedRowCountText(count)` (not 100% sure if that's the best name though, just an example). Maybe @geoffreydhuyvetters also has some more input on this.

src/utils/formatPeriod.js
```diff
@@ -0,0 +1,45 @@
...
+
+const formatPeriod = (startDate, endDate, locale, t) => {
+  const start = new Date(startDate);
+  const end = new Date(endDate);
+
+  const formatTypeDate = 'EEEE d MMMM yyyy';
+  const formatTypeTime = 'HH:mm';
+
+  const formattedDateStart = format(start, formatTypeDate, {
+    locale: locales[locale],
```
bertramakers, line 19: Why don't we just pass the `locale` here? `locales[nl]` always evaluates to `nl`, no? And same for `fr`?

src/components/StatusPageMultiple.js
```diff
@@ -0,0 +1,156 @@
...
+          actions={[
+            {
+              iconName: Icons.PENCIL_ALT,
+              title: t('offerStatus.changeStatus'),
+              onClick: () => setIsModalVisible(true),
+              disabled: selectedRows.length === 0,
+            },
+          ]}
+        />
+        <Link href="/dashboard" variant={LinkVariants.UNSTYLED} customChildren>
```
bertramakers, line 135: This should link to the details page of the place/event as discussed on Slack. (Corneel didn't update the ticket after you discussed with him to make it redirect to the details page for the other calendar types)

bertramakers (commented):
Very clean PR @simon-debruijn ! Thanks for taking the time to split it up in smaller commits, makes it a lot easier to review! 

I didn't spot any big issues, just added some minor comments.

I also love how we can build pages like this with practically almost no extra CSS now :)

src/components/StatusPageMultiple.js
simon-debruijn: The square bracket syntax is needed when we want to apply margins passed from the parent component. Because if we use a React Fragment to wrap these, properties like spacing={5} on the parent component won't work

bertramakers: Okay, weird but good to know. Thanks for the clarification! 👍

src/components/StatusPageNotMultiple.js
bertramakers: Suggestion: `SingleStatusPage` and `MultipleStatusPage`?

src/ui/Icon.js
simon-debruijn: Yes for now we've mapped it 1 to 1 with fontawesome naming. I think we're not going to change away from fontawesome any time soon, or are we? 😄

src/utils/formatPeriod.js
simon-debruijn: The locales used here are the translations from the `date-fns` library.

```
import nl from 'date-fns/locale/nl-BE';
import fr from 'date-fns/locale/fr';
```

src/components/StatusPageMultiple.js
```diff
@@ -0,0 +1,156 @@
...
+  type: PropTypes.string.isRequired,
+  reason: PropTypes.object,
+};
+
+const StatusPageMultiple = ({ event, refetchEvent }) => {
+  const { t, i18n } = useTranslation();
+
+  const eventId = parseOfferId(event['@id']);
+  const name =
+    event?.name?.[i18n.language] ?? event?.name?.[event.mainLanguage];
```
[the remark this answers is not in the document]
simon-debruijn: Jep this is something we could add. I'll leave it out the PR for now

src/ui/Icon.js
bertramakers: No, not immediately, but I would keep open the possibility at least. The goal of our UI components was to abstract away the implementation, like we also do with the Bootstrap components by not exposing all props etc, and using our own names. So I was thinking we should do the same for the icons?

simon-debruijn: @bertramakers Maybe create a ticket for this and we'll pick that up later?

bertramakers: Why? It's only added in this PR so might as well change it right now no?

simon-debruijn: oke done

bertramakers (approved):
🚢 it!

PR #201, opened 2021-03-04
III-3825 Use lodash flatten() instead of Array.prototype.flat()

bertramakers:
### Changed

- Use lodash flatten() instead of Array.prototype.flat() for node 10 compatibility. (The servers and Vagrant box use node 10 right now, and while we should look into updating that it might take some work to get udb3-websockets etc compatible with newer versions, so we should make sure udb3-frontend can be deployed in the meantime.)

---

Ticket: https://jira.uitdatabank.be/browse/III-3825

### DIFF
--- a/src/hooks/api/authenticated-query.js
+++ b/src/hooks/api/authenticated-query.js
@@ -1,3 +1,4 @@
+import flatten from 'lodash/flatten';
 import { useRouter } from 'next/router';
 import { useCallback } from 'react';
 import { Cookies } from 'react-cookie';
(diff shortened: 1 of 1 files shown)


simon-debruijn (approved):

LucWollants (approved):

geoffreydhuyvetters:
Another approach could have been to use a polyfill 

https://www.npmjs.com/package/array-flat-polyfill

bertramakers:
True, we can still do that of course? This is now getting deployed to deblock Corneel on acc but that doesn't mean we can change the fix and include it in a next deploy?

bertramakers:
Because I see the value on using a polyfill so we can already use .flat() now and don't have to refactor usages of lodash' flatten later.

geoffreydhuyvetters:
Jep, no problem, I just like to think forward.

bertramakers:
Yes I was agreeing with you actually, just didn't think of the polyfill and this is now merged already but we can do it in another PR 😄

PR #202, opened 2021-03-08
Add tests for status pages

geoffreydhuyvetters:
### Added

- tests for `place/.../status` and `event/..../status`


src/ui/Link.js
```diff
@@ -82,7 +82,11 @@ const Link = ({
 
   if (isInternalLink) {
     return (
-      <NextLink href={href} passHref prefetch={!process.env.STORYBOOK}>
+      <NextLink
+        href={href}
+        passHref
+        {...(process.env.STORYBOOK ? { prefetch: false } : {})}
```
geoffreydhuyvetters, line 88: changed it like this because I got tons of warning in my Next setup when adding the prop with value true

geoffreydhuyvetters:
We can definitely merge this one too if we want, some tests are better then no tests I would say.

simon-debruijn (approved):
Let's merge it 😄

PR #203, opened 2021-03-12
Add loading state and disabled state when linking or when skipping

simon-debruijn:
### Added

- loading state for link button
- disabled state for skip button

### Changed

- disabled state for skip button

---

Ticket: https://jira.uitdatabank.be/browse/III-3793

Screenshot:
Schermafbeelding 2021-03-12 om 11 46 57

### DIFF
--- a/src/components/pages/manage/productions/create/index.js
+++ b/src/components/pages/manage/productions/create/index.js
@@ -50,19 +50,31 @@ const Create = () => {
     await refetchSuggestedEvents();
   };
 
-  const { mutate: skipSuggestedEvents } = useSkipSuggestedEvents({
+  const {
+    mutate: skipSuggestedEvents,
+    ...skipSuggestedEventsMutation
+  } = useSkipSuggestedEvents({
     onSuccess: handleSuccess,
   });
 
-  const { mutate: createProductionWithEvents } = useCreateWithEvents({
+  const {
+    mutate: createProductionWithEvents,
+    ...createProductionWithEventsMutation
+  } = useCreateWithEvents({
     onSuccess: handleSuccess,
   });
 
-  const { mutate: mergeProductions } = useMergeProductions({
+  const {
+    mutate: mergeProductions,
+    ...mergeProductionsMutation
+  } = useMergeProductions({
     onSuccess: handleSuccess,
   });
 
-  const { mutate: addEventsByIds } = useAddEventsByIds({
+  const {
+    mutate: addEventsByIds,
+    ...addEventsByIdsMutation
+  } = useAddEventsByIds({
     onSuccess: handleSuccess,
   });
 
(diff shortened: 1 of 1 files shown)


src/components/pages/manage/productions/create/index.js
```diff
@@ -50,19 +50,31 @@ const Create = () => {
     await refetchSuggestedEvents();
   };
 
-  const { mutate: skipSuggestedEvents } = useSkipSuggestedEvents({
+  const {
+    mutate: skipSuggestedEvents,
+    ...skipSuggestedEventsMutation
+  } = useSkipSuggestedEvents({
```
geoffreydhuyvetters, line 56: Can't we just use skipSuggestedEventsMutation.mutate everywhere? The mix of restructuring and renaming is a bit confusing imo.

src/components/pages/manage/productions/create/index.js
```diff
@@ -89,6 +101,20 @@ const Create = () => {
...
+        createProductionWithEventsMutation.status,
+        mergeProductionsMutation.status,
+        addEventsByIdsMutation.status,
+      ].some((status) => status === QueryStatus.LOADING),
+    [
+      createProductionWithEventsMutation.status,
+      mergeProductionsMutation.status,
+      addEventsByIdsMutation.status,
+    ],
+  );
```
geoffreydhuyvetters, line 116: We could also use something like this here.

https://react-query.tanstack.com/reference/useIsFetching

geoffreydhuyvetters (approved):

src/components/pages/manage/productions/create/index.js
simon-debruijn: I don't think we can use this one. Because it only counts the queries being executed, since mutations don't contain a key

src/components/pages/manage/productions/create/index.js
simon-debruijn: Changed it

src/components/pages/manage/productions/create/index.js
geoffreydhuyvetters: True!

geoffreydhuyvetters (approved):

src/components/pages/manage/productions/create/index.js
```diff
@@ -50,19 +50,19 @@ const Create = () => {
     await refetchSuggestedEvents();
   };
 
-  const { mutate: skipSuggestedEvents } = useSkipSuggestedEvents({
+  const skipSuggestedEventsMutation = useSkipSuggestedEvents({
```
geoffreydhuyvetters, line 53: To go a step further, we can just call this `skipSuggestedEvents` too, no need to suffix to it with 'Mutation'.

But we can do that in a different PR, let's merge this one :)

geoffreydhuyvetters (approved):

PR #208, opened 2021-03-22
Bug/III-3885 production name remains

simon-debruijn:
### Fixed

- Clear typeahead when a link or skip action is performed successfully

---

Ticket: https://jira.uitdatabank.be/browse/III-3885

### DIFF
--- a/src/components/pages/manage/productions/create/index.js
+++ b/src/components/pages/manage/productions/create/index.js
@@ -7,7 +7,7 @@ import { Button, ButtonVariants } from '@/ui/Button';
 import { Event } from './Event';
 import { getApplicationServerSideProps } from '@/utils/getApplicationServerSideProps';
 import { parseOfferId } from '@/utils/parseOfferId';
-import { useState, useMemo, useEffect, useCallback } from 'react';
+import { useState, useMemo, useEffect, useCallback, useRef } from 'react';
 import {
   useGetProductions,
   useGetSuggestedEvents,
@@ -32,6 +32,7 @@ const Create = () => {
   const { t, i18n } = useTranslation();
   const [searchInput, setSearchInput] = useState('');
   const [selectedProductionId, setSelectedProductionId] = useState('');
+  const typeaheadComponent = useRef();
 
   const {
     data: suggestedEvents,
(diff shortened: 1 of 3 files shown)


LucWollants:
I wonder if it's also an option to not expose the `ref` but work with a boolean property `clear`?
When `clear` is set to true the internal `ref` in  `TypeAhead` is used to call the `clear()` method.

simon-debruijn:
> I wonder if it's also an option to not expose the `ref` but work with a boolean property `clear`?
> When `clear` is set to true the internal `ref` in `TypeAhead` is used to call the `clear()` method.

When using a boolean named `clear`, we won't be able to know when or how many times `clear()` should be called. It could be once, or as long as clear is `true`. So that's not really something we can do

bertramakers:
I also don't think we can use a `clear` prop, because from the outside we'd need to change it back to `false` after setting it to `true` once. Otherwise, it would keep clearing on every re-render. Seems like a hard thing to get right.

What might be a possibility is that every time the value of the typeahead changes, you call some kind of `onTextValueChange` prop that gets passed from outside. And in that callback, the parent component (`Create`) sets the value in its internal state and back as a prop on `TypeaheadWithLabel` which sets the value on the input field. And when the value needs to be cleared, the internal state of the value can be set to an empty string inside `Create`.

However, I'm not sure if that's necessarily a better approach. Maybe @geoffreydhuyvetters can explain what the best practice is for this scenario?

LucWollants (approved):

geoffreydhuyvetters:
Maybe we could use defaultInputValue and force a re-render (by changing the key of the component)

bertramakers:
> Maybe we could use defaultInputValue and force a re-render (by changing the key of the component)

Ah yes, that's clever!

LucWollants:
After some info from Simon, the `forwardRef` approach is how it's explained on the website of the package http://ericgio.github.io/react-bootstrap-typeahead/#public-methods

geoffreydhuyvetters:
Jep, I think it's easy enough to understand.

geoffreydhuyvetters (approved):

PR #211, opened 2021-03-25
Implement sentry

simon-debruijn:
### Added

- sentry

### Changed

- email on errorpage to be clickable emailto link

Linked to [this PR](https://github.com/cultuurnet/appconfig/pull/58)

Screenshot:
Schermafbeelding 2021-03-25 om 17 18 16


src/utils/sentry.js
```diff
@@ -0,0 +1,19 @@
+import * as Sentry from '@sentry/react';
+import { Integrations } from '@sentry/tracing';
+import { publicRuntimeConfig } from '../../next.config';
+
+const initializeSentry = () => {
+  Sentry.init({
+    environment: publicRuntimeConfig.environment,
+    dsn:
+      'https://0aa9d6e3306c46fc90eade0e05b955f8@o424400.ingest.sentry.io/5692505',
```
bertramakers, line 9: Can you also make this configurable through an ENV variable? So that we can set it to an empty string locally. Otherwise Sentry will get flooded with errors caused by local WIP :)

simon-debruijn: Yep of course

simon-debruijn: Done

bertramakers (approved):
Very nice! 👌

src/components/layouts/index.js
```diff
@@ -36,7 +38,9 @@ const useHandleAuthentication = () => {
 
   useEffect(() => {
     if (!getUserQuery.data) return;
+    const { id, email, username } = getUserQuery.data;
     setCookie('user', getUserQuery.data);
+    setSentryUser({ id, email, username });
```
brampauwelyn, line 43: @bertramakers  @simon-debruijn  are we allowed to store email addresses in Sentry?
I mean is this GDPR compliant?

simon-debruijn: Good question, I'm googleing about it. But @bertramakers do you use this feature too on e.g. udb3-silex?

simon-debruijn: When following GDPR, aren't we also obligated to add a cookie-banner to be able to save user data to the cookies?

geoffreydhuyvetters: When you use cookies, you need a cookie banner. So yes.

simon-debruijn: A cookie banner is something to add to the backlog then. But setting the user here for Sentry, idk, I haven't found a clear answer on if we can just do this.

LucWollants: These tags are added for UDB3 backend, either with custom tags are default Sentry tags:
Screenshot 2021-03-26 at 09 27 47

simon-debruijn: Maybe we just save the id? That doesn't reveal to much information

LucWollants: I think we can avoid the cookie banner inside the app because we have a signup process and during the signup process you give a consent for cookies. But I'm not an expert in this field, so better check with someone from "legal" 😅

brampauwelyn: Jup, just the `id` would be better I guess

simon-debruijn: See [this commit](https://github.com/cultuurnet/udb3-frontend/pull/211/commits/982ce8c8fd947cd8bdb3c5dcb73a4d69b06818d1)

bertramakers: Wow there's a lot going on in this thread 😅 

> no email address is stored in Sentry

@LucWollants we do currently store the email address from the backend Sentry calls, see for example https://sentry.io/organizations/publiq-vzw/issues/2287500812/?project=5435528&query=is%3Aunresolved (it's not a tag, it's user info)

> are we allowed to store email addresses in Sentry?
> I mean is this GDPR compliant?

@brampauwelyn Technically I think yes it's compliant. In the sense that we already have that info when the user signs up and we are allowed to process it. Sentry itself is also GDPR compliant. However, I agree that storing only the id might be better because while Sentry is a closed system and only accessible with a login, it is possible to make issues public to share them and then we might accidentally expose personally identifiable information. And having the email in Sentry is probably not that necessary now that I think about it. In the case that we really need to identify a particular user we can look up their id in Auth0.

> When following GDPR, aren't we also obligated to add a cookie-banner to be able to save user data to the cookies?

@simon-debruijn 
"When following GDPR" -> We always have to follow GDPR.
Storing the email address has nothing to do with cookies though.

But about a cookie banner in general: I agree with @LucWollants that a cookie banner is probably not necessary since you need to login first. However, I don't see a cookie consent on Auth0/CP right now. Shouldn't we add one there instead? (@brampauwelyn ?)

The only other cookie I know of that we store before logging in is the selected language. So maybe we should just find a solution for that. A cookie banner won't necessarily solve that because right now we set that language cookie immediately once you visit the landing page, while even with a cookie banner we shouldn't do that but wait for the consent instead.

Btw maybe we should discuss privacy- and security-sensitive questions like this on Slack in the future 😉

PR #212, opened 2021-03-29
Improvement - Make all Lodash imports the same

simon-debruijn:
### Changed

- Lodash imports

Reference: [lodash-import-done-right](https://dev.to/icokie/lodash-import-done-right-290n)

### DIFF
--- a/src/components/pages/manage/productions/index/index.js
+++ b/src/components/pages/manage/productions/index/index.js
@@ -14,7 +14,7 @@ import { parseOfferId } from '@/utils/parseOfferId';
 import { QueryStatus } from '@/hooks/api/authenticated-query';
 
 import { Text } from '@/ui/Text';
-import { throttle } from 'lodash';
+import throttle from 'lodash/throttle';
 import { useQueryClient } from 'react-query';
 import { DeleteModal } from './DeleteModal';
 import { Events } from './Events';
--- a/src/components/StatusPageMultiple.js
+++ b/src/components/StatusPageMultiple.js
@@ -12,7 +12,7 @@ import { Text } from '@/ui/Text';
 import { getValueFromTheme } from '@/ui/theme';
 import { StatusModal } from './StatusModal';
 import { QueryStatus } from '@/hooks/api/authenticated-query';
-import { camelCase } from 'lodash';
+import camelCase from 'lodash/camelCase';
 import { Icons } from '@/ui/Icon';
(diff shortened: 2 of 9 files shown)


brampauwelyn (approved):

src/components/pages/manage/productions/index/index.js
```diff
@@ -14,7 +14,7 @@ import { parseOfferId } from '@/utils/parseOfferId';
 import { QueryStatus } from '@/hooks/api/authenticated-query';
 
 import { Text } from '@/ui/Text';
-import { throttle } from 'lodash';
+import throttle from 'lodash/throttle';
```
bertramakers, line 17: If this is for the autocomplete, maybe we should use `debounce` instead? https://css-tricks.com/the-difference-between-throttling-and-debouncing/

With throttle we will send one request every `275ms` even if the user is still typing, while with debounce we will only send a request once the user has stopped typing for `275ms`

bertramakers: (Same for the throttle on the create page)

simon-debruijn: From what I remember we specifically chose throttle, for a more fluid user experience, because we first had debounce

bertramakers: Won't that result in the same issue as in https://jira.uitdatabank.be/browse/III-3824 then?

bertramakers: Since if the backend response time takes ~1s but you do a request every 275ms, the wait time will increase a lot?

bertramakers: Ah but the labels issue doesn't even use throttle at all so maybe throttle is fine. This is an interesting read though https://www.peterbe.com/plog/how-to-throttle-and-debounce-an-autocomplete-input-in-react

simon-debruijn: I'm not sure if it will cause an issue. `debounce` and `throttle` both only look at the result of the last call. 

But the difference lies in the result already being fetched while typing vs typing and then waiting for the result.

Both have their advantages, if we want less calls we could use debounce

bertramakers: We could do both as mentioned in the article above. But could just be a nice improvement down the line if needed, not necessary now then.

PR #217, opened 2021-04-07
III-3893 Trigger production search after entering 3 characters

LucWollants:
### Changed
- Trigger production search after entering at least 3 characters
---
Ticket: https://jira.uitdatabank.be/browse/III-3893

### DIFF
--- a/src/components/pages/manage/productions/create/index.js
+++ b/src/components/pages/manage/productions/create/index.js
@@ -29,6 +29,7 @@ const ProductionStatus = {
 };
 
 const Create = () => {
+  const minSearchLength = 3;
   const { t, i18n } = useTranslation();
   const [searchInput, setSearchInput] = useState('');
   const [selectedProductionId, setSelectedProductionId] = useState('');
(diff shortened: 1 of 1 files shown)


src/components/pages/manage/productions/create/index.js
```diff
@@ -235,6 +242,7 @@ const Create = () => {
                   maxWidth="43rem"
                   label={t('productions.create.production_name')}
                   emptyLabel={t('productions.create.no_productions')}
+                  minLength={minSearchLength}
                   onInputChange={throttle(handleInputSearch, 275)}
```
simon-debruijn, line 246: Looks good, the only improvement I see is that throttle is still called a few times while we don't want to.

So maybe we could write it like this:

```javascript
onInputChange={handleInputSearch}
```

And 

```javascript
const handleInputSearch = useCallback((searchTerm) => {
  const trimmedSearchTerm = searchTerm.toString().trim();
  if (trimmedSearchTerm.length < minSearchLength) return;
  return throttle(() => {
    setSelectedProductionId(undefined);
    setSearchInput(trimmedSearchTerm);
  }, 275);
})
```

LucWollants: Is it not more clear if the throttle stays in the same place?

simon-debruijn: Yes it's clearer. Just leave it as it is. I'll approve

simon-debruijn (approved):

PR #219, opened 2021-04-12
Feature flags setup

geoffreydhuyvetters:
### Added

- Simple hook to check if a feature flag is enabled and set it
- constants to use in code
- helper for the browser console
- custom 500 page

```js
setFeatureFlag(FeatureFlags.REACT_DASHBOARD, false)
```

### Updated

- Next to 10.1 and using Webpack 5 now

### DIFF
--- a/src/constants/FeatureFlags.js
+++ b/src/constants/FeatureFlags.js
@@ -1,6 +1,0 @@
-const FeatureFlags = {
-  REACT_DASHBOARD: 'react_dashboard',
-  REACT_CREATE: 'react_create',
-};
-
-export { FeatureFlags };
--- a/src/constants/FeatureFlags.js
+++ b/src/constants/FeatureFlags.js
@@ -1,0 +1,6 @@
+const FeatureFlags = {
+  REACT_DASHBOARD: 'react_dashboard',
+  REACT_CREATE: 'react_create',
+};
+
+export { FeatureFlags };
--- a/src/hooks/useFeatureFlag.js
+++ b/src/hooks/useFeatureFlag.js
@@ -1,29 +1,18 @@
-import { FeatureFlags } from '@/constants/FeatureFlags';
-import Cookies from 'universal-cookie';
 import { useCookiesWithOptions } from './useCookiesWithOptions';
 
-const cookies = new Cookies();
+const createCookieName = (identifier) => `ff_${identifier}`;
 
-const createFullIdentifier = (identifier) => `ff_${identifier}`;
-
-const useFeatureFlag = (identifier) => {
-  if (!identifier) return [false, () => {}];
+const useFeatureFlag = (featureFlagName) => {
+  if (!featureFlagName) return [false, () => {}];
 
   const { cookies, setCookie } = useCookiesWithOptions();
 
-  const fullIdentifier = createFullIdentifier(identifier);
+  const cookieName = createCookieName(featureFlagName);
 
-  const set = (value) => setCookie(fullIdentifier, value);
-  const value = cookies?.[fullIdentifier] ?? false;
+  const set = (value) => setCookie(cookieName, value);
+  const value = cookies?.[cookieName] ?? false;
 
   return [value, set];
 };
 
-if (typeof window !== 'undefined') {
-  window.FeatureFlags = FeatureFlags;
-  window.setFeatureFlag = (identifier, value) => {
-    cookies.set(createFullIdentifier(identifier), value);
-  };
-}
-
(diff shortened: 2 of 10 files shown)


src/constants/FeatureFlags.js
```diff
@@ -0,0 +1,6 @@
+const FeatureFlags = {
+  REACT_DASHBOARD: 'react_dashboard',
+  REACT_CREATE: 'react_create',
+};
+
+export { FeatureFlags };
```
bertramakers, line 6: Small nit pick: Do you think we'll use this constant somewhere without using the `useFeatureFlag` hook? If not I think we can put it together with the hook in the same file.

I think this `constants` folder was just added for constants that are used in multiple contexts, not related to one specific hook/component.

src/hooks/useFeatureFlag.js
```diff
@@ -1,16 +1,29 @@
...
 
   return [value, set];
 };
 
+if (typeof window !== 'undefined') {
+  window.FeatureFlags = FeatureFlags;
+  window.setFeatureFlag = (identifier, value) => {
+    cookies.set(createFullIdentifier(identifier), value);
+  };
+}
```
bertramakers, line 27: No comment but just a question: I'm curious why we need these on the `window` object? To use them manually in the `console` or something?

src/constants/FeatureFlags.js
simon-debruijn: Yep I think this can also live in useFeatureFlags

bertramakers (approved):

src/hooks/useFeatureFlag.js
bertramakers: Nevermind I saw your explanation on Slack now 👍

simon-debruijn (approved):
Would move `FeatureFlags` constant to `usefeatureFlag.js`.

The rest looks good 👍🏻

src/constants/FeatureFlags.js
geoffreydhuyvetters: Jep, good idea

jdrieghe (approved):
image

src/hooks/useFeatureFlag.js
geoffreydhuyvetters: idd, exposing a function on window makes them 'global' to use in the console

bertramakers:
The Chromatic build seems to be failing because it cannot read the apiKey config (which is undefined at build time). I think this will also be a problem when we build it on Jenkins, or at the latest when we deploy it and run it with runtime config on the servers. Because Next seems to be reading the config at build-time again. Is this something that changed in the Next update maybe?

simon-debruijn:
> The Chromatic build seems to be failing because the Next app cannot read the apiKey config (which is undefined at build time). I think this will also be a problem when we build the app on Jenkins, or at the latest when we deploy it and run it with runtime config on the servers. Because Next seems to be reading the config at build-time again. Is this something that changed in the Next update maybe?

Yep I think it has been changed.

We can just call `process.env.NEXT_PUBLIC_MY_VAR` it seems. In stead of having to use `next/config`
https://nextjs.org/docs/basic-features/environment-variables#exposing-environment-variables-to-the-browser

geoffreydhuyvetters:
that's not new since 10.1 @simon-debruijn, we tried it like that before (on 10.0.x) but for some reason that didn't work in our setup.

Not sure what's happening now to be fair 🤔 Let me have a look.

simon-debruijn:
> that's not new since 10.1 @simon-debruijn, we tried it like that before (on 10.0.x) but for some reason that didn't work in our setup. I think @bertramakers can provide a bit of context there.
> 
> Not sure what's happening now to be fair 🤔 Let me have a look.

Hmm ok strange

bertramakers:
I think `process.env.NEXT_PUBLIC_MY_VAR` is for exposing environment variables to the browser, but their values will still be "hardcoded" into the build files. Both `process.env.NEXT_PUBLIC_...` and `publicRuntimeConfig` are public, but we also need it to be runtime config specifically.

geoffreydhuyvetters:
Exactly, that was it. It's causing issues on the 500 page, which is new to 10.1. At least I have a direction to look into.

bertramakers:
> Exactly, that was it. It's causing issues on the 500 page, which is new to 10.1. At least I have a direction to look into.

I seem to remember we also had that issue with the 404 page. Next is trying to render them as static pages. So we forced the 404 page to be rendered server-side instead (vs at build time) I think.

PR #223, opened 2021-04-13
Use consistent plural naming for routes

LucWollants:
### Changed
- Use consistent plural naming for routes

---

Spotted this together with @bertramakers while working on the ReactJS dashboard


bertramakers (approved):

bertramakers:
I'm going to go ahead and merge this, if you spot a mistake or problem @geoffreydhuyvetters or @simon-debruijn feel free to ping us and we'll fix it in a follow-up PR

next.config.js
```diff
@@ -15,6 +15,16 @@ module.exports = {
         destination: '/dashboard',
         permanent: false,
       },
+      {
+        source: '/event/:eventId/status',
+        destination: '/events/:eventId/status',
+        permanent: true,
```
geoffreydhuyvetters, line 21: Permanent acts a bit weird when you're working on multiple projects, that's why we're using false. We should probably change that.

simon-debruijn: Yep the rest looks good, but the permanent boolean should be false

bertramakers: Hmm, the thing is that in production mode ideally it would be permanent though. There's no need for the browser to not remember it when the app is running on a domain that is only used by our app like acc.uitdatabank.be, www.uitdatabank.be, ...

Is there a way to check here if we're running in dev mode or not and set the boolean based on that so that it doesn't give problems when running on localhost?

simon-debruijn: You probably can do this:

`permanent: (process.env.NEXT_PUBLIC_ENVIRONMENT !== 'development')`

I added `NEXT_PUBLIC_ENVIRONMENT` also to use for Sentry. So locally you should name this to `development`, and the other environments have `acceptance`, `testing` an `production`

bertramakers: I was looking into it and thinking of using `process.env.NODE_ENV === 'production'` actually. Testing it as we speak 😄  What would be the advantage of using the custom environment variable? Also, I think it's missing in the local config: https://github.com/cultuurnet/udb3-vagrant/blob/main/config/udb3-frontend/env

simon-debruijn: There is a difference, since from my understanding `process.env.NODE_ENV` only tells us if we're running a production build ('production') or running in dev mode

bertramakers: Which is what we need to know, no?

simon-debruijn: Yes I think we can set `NODE_ENV` itself in our env files

bertramakers: No I mean that `NEXT_PUBLIC_ENVIRONMENT` is missing.

But regardless of that, `process.env.NODE_ENV` seems to be what we need here, since it would also make the redirects permanent for `www.uitdatabank.be` (= `production` build, in a local `development` environment)

geoffreydhuyvetters: NODE_ENV is set through Webpack (via Next.js), just not sure what NODE_ENV returns when you create a production build and run it locally. (yarn build && yarn start)

simon-debruijn: I think if you use NODE_ENV now as it is and you create a local production build and run it. The permanent will be set to true, because he sees it as `production`

simon-debruijn: And this will still run on localhost, thus creating the same issue we had no?

bertramakers: Hmm I overlooked that because I usually test production builds using the initscripts and `www.uitdatabank.dev`. Do you often test production builds on `localhost`?

simon-debruijn: Sometimes 😉

geoffreydhuyvetters: That happens, yes. Mostly for SSR issues

simon-debruijn: But I think if I use `NODE_ENV` on all envirmonments, and set it to the correct values for each environment. It will work as expected

bertramakers: I wouldn't overwrite something like `NODE_ENV` though, that will probably give unexpected results sometime down the line. I'll just use `NEXT_PUBLIC_ENVIRONMENT` for now. It just means that the redirects will not permanent on `www.uitdatabank.dev` either but that's not a big issue.

simon-debruijn: so in `local.env` set `NODE_ENV=development` `acceptance.env` set `NODE_ENV=acceptance` and so on

simon-debruijn: But it's not like I've noticed big draw backs from not using permanent, do you feel like this makes a huge difference?

bertramakers: If it's permanent the redirect is instant because it gets cached. If not it takes a HTTP request to the server to check if the redirect is still needed. For one redirect it's not a big deal but if it's many that starts to add up and it's easily fixed using `NEXT_PUBLIC_ENVIRONMENT`? Just not for `.dev` but that's okay.

bertramakers: Resolved in https://github.com/cultuurnet/udb3-frontend/pull/224

LucWollants: If we locally hit an issue with a permanent redirect we can clear the cache of our browser or use an incognito window?

bertramakers: Yes a cache clear would fix it but I can understand that that would get annoying if you often need to switch between projects, especially since we have a redirect on `/` (to `/dashboard` now, `/events` later)

PR #224, opened 2021-04-14
Don't make redirects permanent on localhost

bertramakers:
### Changed

- Made redirects only permanent when not in the `development` environment, to prevent issues with other projects running on `localhost`


bertramakers:
Note that you will need the env variable added in https://github.com/cultuurnet/udb3-vagrant/pull/62

geoffreydhuyvetters (approved):

PR #225, opened 2021-04-14
III-3954 Add Tabs component

LucWollants:
### Added

- Added Tabs component

---

Ticket: https://jira.uitdatabank.be/browse/III-3954


src/ui/Tabs.js
```diff
@@ -30,6 +30,8 @@ const Tabs = ({ activeKey, onSelect, children: rawChildren }) => {
         .nav-item {
           color: ${getValue('color')};
           border-radius: ${getValue('borderRadius')};
+          padding: 10px 15px;
+          margin-right: 2px;
```
bertramakers, line 33: @geoffreydhuyvetters @simon-debruijn this padding and margin now gets added to a nested bootstrap element (the individual tabs that you can switch between). Is this okay or should we try to find a way to use the `Box` component for those?

I _think_ we might be able to make the individual `Tab` components start from `Box`, but I'm not sure how to do that because they now use `BootstrapTab` underneath.

I tried looking at `ContentModal` and `Button`, but those don't use `Box` either for Bootstrap components.

bertramakers: This is the `Tab` component that we might be able to change to make it use `Box`, but again I don't know how exactly: https://github.com/cultuurnet/udb3-frontend/pull/225/files#diff-1366f718fdcc66938f8cc29d6f2a19a0314527631b16db8d7cdbe6e016cc700aR63-R70

bertramakers: I tried removing the `padding` and `margin-right` from the CSS and to render the `Tab` like this:

```
const Tab = ({ eventKey, title }) => {
  return <Inline forwardedAs={BootstrapTab} eventKey={eventKey} title={title} marginRight={2} paddingY={2} paddingX={10} />;
};
```
(With an exagerated `paddingX` to see if it worked)

But the tabs just had the default Bootstrap margin & padding.
I think because Bootstrap's Tabs component doesn't directly render the `Tab` component inside.

bertramakers: Resolved in dde0c2a

src/ui/Tabs.js
```diff
@@ -0,0 +1,83 @@
+import { Tab as BootstrapTab, Tabs as BootstrapTabs } from 'react-bootstrap';
+import { getValueFromTheme } from './theme';
+import * as PropTypes from 'prop-types';
```
simon-debruijn, line 3: I think the import should just be 

`import PropTypes from 'prop-types'`

simon-debruijn (approved):

simon-debruijn (approved):

src/ui/Tabs.js
bertramakers: Fixed in 3528f9e

PR #227, opened 2021-04-14
Bug with display: flex on links

bertramakers:
This PR illustrates the issue with `display: flex` on links.

You can view the Story that illustrates it on https://5f69c0eb52e6b60022561d99-jyexhyrsvo.chromatic.com/?path=/story/components-link--internal

Screen Shot 2021-04-14 at 17 27 50

On the Story for the unstyled variant with text, there is a similar/related issue: https://5f69c0eb52e6b60022561d99-jyexhyrsvo.chromatic.com/?path=/story/components-link--unstyled-external

The unstyled variant has `width: min-content`, which was probably meant to undo the 100% width set by `display: flex`, but makes the text inside the link wrap to multiple lines while it could fit on a single line.

Screen Shot 2021-04-14 at 17 28 41

### DIFF
--- a/src/ui/Link.js
+++ b/src/ui/Link.js
@@ -18,6 +18,7 @@ const BaseLink = forwardRef(({ variant, ...props }, ref) => {
       <Inline
         ref={ref}
         forwardedAs="a"
+        width="min-content"
         color={{ default: 'inherit', hover: 'inherit' }}
         alignItems="center"
         css={`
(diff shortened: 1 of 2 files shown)


bertramakers:
I added two commits that fixed it for the stories in Storybook, but then the sidebar links looked like this so I reverted them:

Screen Shot 2021-04-14 at 18 54 16

simon-debruijn (approved):

PR #230, opened 2021-04-16
Fix client side redirects

bertramakers:
### Added

- Added "client-side" redirects (you can test this by going to the detail page of an event and clicking the "Status wijzigen" button -> This should redirect from `/event/:eventId/status` to `/events/:eventId/status`)
- Added possibility to put redirects handled by `[...params].js` behind feature flags (so only works for paths that don't exist in React, which should be fine for the porting of AngularJS pages to React and making the AngularJS pages redirect when the React page is finished)

### Changed

- Changed 404 logic to only listen for `URL_UNKNOWN` messages in the `Fallback` component, because if it was registered sooner in the app itself the 404 logic would conflict with the redirect logic and sometimes a page that should redirect would show 404 instead because the iframe would be rendered and the AngularJS app would send a message that the page was not found before the redirect logic had kicked in
- 404 pages are now rendered on the current path instead of redirecting to `/404`

---

**Note!** This points to #231 because that fix made it easier to test that the client-side and server-side rendering still works as expected with these redirects in place.

Screenshot of how the 404 now looks (notice the unchanged URL)

Screen Shot 2021-04-16 at 15 38 31

### DIFF
--- a/next.config.js
+++ b/next.config.js
@@ -8,6 +8,17 @@ module.exports = {
     '@storybook/addon-essentials',
     '@storybook/addon-a11y',
   ],
+  async redirects() {
+    // Redirects to fix non-existing paths should go in `src/redirects.js`!!!
+    const env = process.env.NEXT_PUBLIC_ENVIRONMENT;
+    return [
+      {
+        source: '/',
+        destination: '/dashboard',
+        permanent: env !== 'development',
+      },
+    ];
+  },
   publicRuntimeConfig: {
     environment: process.env.NEXT_PUBLIC_ENVIRONMENT,
     apiKey: process.env.NEXT_PUBLIC_API_KEY,
--- a/next.config.js
+++ b/next.config.js
@@ -1,3 +1,5 @@
+const { getRedirects } = require('./src/redirects');
+
 module.exports = {
   future: {
     webpack5: true,
@@ -9,27 +11,8 @@ module.exports = {
     '@storybook/addon-a11y',
   ],
   async redirects() {
-    // Only make the permanent redirects really permanent in environments other
-    // than development, so we don't get permanent redirects on localhost which
-    // may conflict with other projects.
     const env = process.env.NEXT_PUBLIC_ENVIRONMENT;
-    return [
-      {
-        source: '/',
-        destination: '/dashboard',
-        permanent: env !== 'development',
-      },
-      {
-        source: '/event/:eventId/status',
-        destination: '/events/:eventId/status',
-        permanent: env !== 'development',
-      },
-      {
-        source: '/place/:placeId/status',
-        destination: '/places/:placeId/status',
-        permanent: env !== 'development',
-      },
-    ];
+    return getRedirects(env);
   },
   publicRuntimeConfig: {
     environment: process.env.NEXT_PUBLIC_ENVIRONMENT,
--- a/next.config.js
+++ b/next.config.js
@@ -1,5 +1,3 @@
-const { getRedirects } = require('./src/redirects');
-
 module.exports = {
   future: {
     webpack5: true,
@@ -10,10 +8,6 @@ module.exports = {
     '@storybook/addon-essentials',
     '@storybook/addon-a11y',
   ],
-  async redirects() {
-    const env = process.env.NEXT_PUBLIC_ENVIRONMENT;
-    return getRedirects(env);
-  },
   publicRuntimeConfig: {
     environment: process.env.NEXT_PUBLIC_ENVIRONMENT,
     apiKey: process.env.NEXT_PUBLIC_API_KEY,
--- a/package.json
+++ b/package.json
@@ -46,7 +46,7 @@
     "react-dom": "^17.0.1",
     "react-i18next": "^11.7.3",
     "react-query": "^3.2.0",
-    "react-router-dom": "^5.2.0",
+    "react-router": "^5.2.0",
     "react-table": "^7.6.3",
     "sass": "^1.27.0",
     "sass-loader": "^10.0.4",
--- a/package.json
+++ b/package.json
@@ -46,6 +46,7 @@
     "react-dom": "^17.0.1",
     "react-i18next": "^11.7.3",
     "react-query": "^3.2.0",
+    "react-router-dom": "^5.2.0",
     "react-table": "^7.6.3",
     "sass": "^1.27.0",
     "sass-loader": "^10.0.4",
--- a/src/components/pages/[...params].js
+++ b/src/components/pages/[...params].js
@@ -13,11 +13,12 @@ import {
 } from '@/hooks/useHandleWindowMessage';
 import PageNotFound from '@/pages/404';
 import { useIsClient } from '@/hooks/useIsClient';
+import { Cookies } from 'react-cookie';
 
 const prefixWhenNotEmpty = (value, prefix) =>
   value ? `${prefix}${value}` : value;
 
-const getRedirect = (originalPath, environment) => {
+const getRedirect = (originalPath, environment, cookies) => {
   const matches = getRedirects(environment)
     .map(({ source, destination, permanent }) => {
       const match = matchPath(originalPath, { path: source, exact: true });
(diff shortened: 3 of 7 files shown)


bertramakers:
Converting this back to draft because I want to see if I can use `getServerSideProps` to redirect server-side, instead of the redirects in `next.config.js`, so we can use feature flags for redirects (because we can check the cookies both client- and server-side in `[...params].js`)

geoffreydhuyvetters:
Not sure about the 404 change.

Shouldn't it redirect to a 404 page? 

https://nextjs.org/docs/advanced-features/custom-error-page#customizing-the-404-page

bertramakers:
> Not sure about the 404 change.
> 
> Shouldn't it redirect to a 404 page?
> 
> https://nextjs.org/docs/advanced-features/custom-error-page#customizing-the-404-page

I think that's if you don't use the iframe fallback that we do. I assume it then shows that 404 page on the original path that was not found.

However, in our case we're doing a hard redirect to `/404` which has 2 problems:
- The user can't copy paste the url to demonstrate where the 404 actually occurs (in Zendesk etc)
- The back button goes back to the not found url which then redirects back to `/404` so you get stuck in a loop

Aslo looking at other websites/apps, almost none redirect to an explicit `/404` path and instead just show the error on the requested path?

geoffreydhuyvetters:
@bertramakers See my edit.

bertramakers:
> @bertramakers See my edit.

Missed that, we probably cross-posted. 👍

geoffreydhuyvetters:
Makes sense with the iframe setup though, there is no **"real"** Next.js 404 situation here

bertramakers:
> Makes sense with the iframe setup though, there is no **"real"** Next.js 404 situation here

Exactly! If we didn't have that we wouldn't even need to handle that 404 logic ourselves and we could just add a 404 page and Next would probably render that for every path that doesn't exist.

bertramakers:
> Converting this back to draft because I want to see if I can use `getServerSideProps` to redirect server-side, instead of the redirects in `next.config.js`, so we can use feature flags for redirects (because we can check the cookies both client- and server-side in `[...params].js`)

Was able to do this + add a check for feature flags! 🎉

package.json
```diff
@@ -46,6 +46,7 @@
     "react-dom": "^17.0.1",
     "react-i18next": "^11.7.3",
     "react-query": "^3.2.0",
+    "react-router-dom": "^5.2.0",
```
geoffreydhuyvetters, line 49: Can't we just add `react-router` as a dependency instead of `react-router-dom`?

src/components/pages/[...params].js
```diff
@@ -4,11 +4,50 @@ import { useRouter } from 'next/router';
...
+      const match = matchPath(originalPath, { path: source, exact: true });
+      if (match) {
+        return {
+          destination: generatePath(destination, match.params),
+          permanent: featureFlag === undefined && permanent,
+        };
+      }
+      return false;
+    })
+    .filter((match) => !!match);
```
Comment (geoffreydhuyvetters) [inline src/components/pages/[...params].js:43]: We could also use a find here

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find

package.json
bertramakers: Didn't know that was possible, I just followed https://reactrouter.com/web/guides/quick-start
Thanks for the tip

src/components/pages/[...params].js
Comment (bertramakers) [reply src/components/pages/[...params].js]: Done! I wasn't able to remove the `map` though because `find` only filters the array, it doesn't allow us to change the values or return a new value.

next.config.js
```diff
@@ -9,26 +9,14 @@ module.exports = {
...
   async redirects() {
-    // Only make the permanent redirects really permanent in environments other
-    // than development, so we don't get permanent redirects on localhost which
-    // may conflict with other projects.
+    // Redirects to fix non-existing paths should go in `src/redirects.js`!!!
     const env = process.env.NEXT_PUBLIC_ENVIRONMENT;
     return [
       {
         source: '/',
         destination: '/dashboard',
```
simon-debruijn, line 17: Is this kept on purpose for now? since we were goinf to change it to just the root for dashboard

bertramakers: The new dashboard is going to be `/events` technically. (= The events tab on the dashboard)
But we have to keep this redirect here for `/` because `/` doesn't get handled by `[...params].js`, so our own redirect logic doesn't kick in.

src/components/pages/[...params].js
```diff
@@ -4,11 +4,45 @@ import { useRouter } from 'next/router';
...
 
+const getRedirect = (originalPath, environment, cookies) => {
+  return getRedirects(environment)
+    .map(({ source, destination, permanent, featureFlag }) => {
+      // Don't follow redirects that are behind a feature flag
+      if (featureFlag) {
+        const enabled = isFeatureFlagEnabledInCookies(featureFlag, cookies);
+        if (!enabled) {
+          return false;
+        }
```
Comment (simon-debruijn) [inline src/components/pages/[...params].js:30]:
```suggestion
      if (featureFlag) {
        const isEnabled = isFeatureFlagEnabledInCookies(featureFlag, cookies);
        if (!isEnabled) {
          return false;
        }
```

src/components/pages/[...params].js
```diff
@@ -4,11 +4,45 @@ import { useRouter } from 'next/router';
...
+      }
+
+      // Check if the redirect source matches the current path
+      const match = matchPath(originalPath, { path: source, exact: true });
+      if (match) {
```
Comment (simon-debruijn) [inline src/components/pages/[...params].js:35]:
```suggestion
      const isMatch = matchPath(originalPath, { path: source, exact: true });
      if (isMatch) {
```

Comment (bertramakers) [reply src/components/pages/[...params].js:35]: This is not a boolean though.

Comment (bertramakers) [reply src/components/pages/[...params].js:35]: See for example https://github.com/cultuurnet/udb3-frontend/pull/230/files/d7cf50cf95f215a21fb32adbe1298476204794f6#diff-2940335828d78d99c2e896068a755ca522bc64c04b2e68a778b32a9ec0559669R37 where it's used as an object.

src/components/pages/[...params].js
```diff
@@ -42,9 +91,53 @@ const Fallback = () => {
...
+      const queryParameters = new URLSearchParams(query);
+
+      // Return the redirect as-is if there are no additional query parameters
+      // to append.
+      if (queryParameters.toString().length === 0) {
+        return { redirect };
+      }
+
+      // Append query parameters to the redirect destination.
+      const glue = redirect.destination.indexOf('?') !== -1 ? '&' : '?';
```
Comment (simon-debruijn) [inline src/components/pages/[...params].js:133]:
```suggestion
      const glue = redirect.destination.includes('?') ? '&' : '?';
```

src/components/pages/[...params].js
Comment (bertramakers) [reply src/components/pages/[...params].js]: See f3c70d1

src/hooks/useFeatureFlag.js
```diff
@@ -15,9 +15,19 @@ const useFeatureFlag = (featureFlagName) => {
   const cookieName = createCookieName(featureFlagName);
 
   const set = (value) => setCookie(cookieName, value);
-  const value = cookies?.[cookieName] ?? false;
+  const enabled = isFeatureFlagEnabledInCookies(featureFlagName, cookies);
```
simon-debruijn, line 18:
```suggestion
  const isEnabled = isFeatureFlagEnabledInCookies(featureFlagName, cookies);
```

src/hooks/useFeatureFlag.js
```diff
@@ -15,9 +15,19 @@ const useFeatureFlag = (featureFlagName) => {
...
 
-  return [value === 'true', set];
+  return [enabled, set];
```
simon-debruijn, line 20:
```suggestion
  return [isEnabled, set];
```

simon-debruijn (approved):

src/hooks/useFeatureFlag.js
bertramakers: See e60e112 (combined your two suggestions into one commit)

src/hooks/useFeatureFlag.js
bertramakers: See e60e112 (combined your two suggestions into one commit)

PR #234, opened 2021-04-20
III-3965 Dropdown button

bertramakers:
### Added

- Added `Dropdown` component based on [Bootstrap's Dropdown component](https://react-bootstrap.github.io/components/dropdowns/#overview)

---

Ticket: https://jira.uitdatabank.be/browse/III-3965

The Dropdown component is based on the `SplitButton` example, but without the `SplitButton` convenience component: https://react-bootstrap.github.io/components/dropdowns/#split-button-dropdowns

If we'd add a `SplitButton` convenience component, we'd have to make to accept `href` and `onClick`, mixing the concept of links and buttons which we want to avoid.

For the menu items, the `Dropdown.Item` component currently accepts both `onClick` and `href` because I didn't see much added value in splitting that up in `Dropdown.Link` and `Dropdown.Button`, but it can be done of course. Both would return Bootstrap's `Dropdown.Item` then, but one with `href` and the other with `onClick`.

For the styling I tried to use `as={Button}` on the `BootstrapDropdown.Toggle` first, but that resulted in a "squished" toggle button, the menu not appearing when toggling, and the `aria` attributes (for accessibility) missing from the toggle button. So instead I set the custom Button CSS on the toggle button.

I also tried using `<Box as={...}>` for the root component, but the custom CSS didn't work then. So I ended up putting the `BootstrapDropdown` as a child inside the `Box` component (see commit db56d59) 

**Example in primary variant**

Screen Shot 2021-04-20 at 17 34 15

**Example in secondary variant**

Screen Shot 2021-04-20 at 17 34 22

~There are some minor CSS changes that still need to be done to get this 100% polished:~

- [x] Fix styling of the toggle button when the dropdown menu is open
- [x] Fix styling of the toggle button when the dropdown menu is open and you hover over the toggle button
- [x] Fix styling of the border between the primary button/link and the toggle button in `success` and `danger` variants
- [x] Remove border-radius from menu


LucWollants (approved):
I had a look at the code and did some testing in Storybook.  Everything is working as expected.

PR #236, opened 2021-04-21
III-3981 - Fix safari bug

simon-debruijn:
### Added

- hook

### Fixed

- Safari bug

---

Ticket: https://jira.uitdatabank.be/browse/III-3981


LucWollants:
Do we also need to have the same fallback for the `removeEventListener`?

simon-debruijn:
> Do we also need to have the same fallback for the `removeEventListener`?
> See https://developer.mozilla.org/en-US/docs/Web/API/MediaQueryList/removeListener

Yep true

simon-debruijn:
> > Do we also need to have the same fallback for the `removeEventListener`?
> > See https://developer.mozilla.org/en-US/docs/Web/API/MediaQueryList/removeListener
> 
> Yep true

Since we only use it in one place, I've removed the hook

LucWollants (approved):

PR #238, opened 2021-04-27
III-3974 Get events, places and organizers by creator

LucWollants:
### Added
- Get events by creatorId and email
- Get places by creatorId and email
- Get organizer by id
- Get organizer by creatorId
- Needed typings for api calls

---

Ticket: https://jira.uitdatabank.be/browse/III-3974


src/hooks/api/events.ts
```diff
@@ -0,0 +1,304 @@
...
+    },
+  });
+  if (isErrorObject(res)) {
+    // eslint-disable-next-line no-console
+    return console.error(res);
+  }
+  return await res.json();
+};
+
+type UseGetEventsByCreatorArguments = {
```
LucWollants, line 139: I wonder if we need to define a type `UseGetEventsByCreatorArguments` for the arguments of a method. I think parameters with types is perfectly fine?

simon-debruijn: The properties creatorid, start, limit, ... live in a object, not as seperate arguments. That's why I can't just say:

```javascript
const useGetEventsByCreator = (
  {
    creatorId: string,
    start: number = 0,
    limit: number = 50,
    sortOptions: SortOptions = { field: 'modified', order: 'desc' },
  }: UseGetEventsByCreatorArguments,
  configuration: UseQueryOptions = {},
) =>
```

In javascript this is seen as use this property by another name

LucWollants: Ok, I think this is indeed very easy in JavaScript because calling this method can be done by creating the parameter object on the fly like this:
```
useGetEventsByCreator({
  creatorId: 1,
  start: 0,
  limit: 50,
  sortOptions: {
    field: 'modified',
    order: 'desc'
  }
})
```

bertramakers (approved):
My knowledge of Typescript is very limited but as far as I can see this just added types without changing the "interface" of the functions/objects so I think that should be fine for now. If the structure of the types isn't perfect yet than that means we should refactor those models in later PRs.

The queries for the dashboard items also look correct to me. 👌

brampauwelyn (approved):

PR #239, opened 2021-04-28
Initial TypeScript setup

geoffreydhuyvetters:
### Added

- Typescript support (JS and TS[X] mixed

### Changed
- simplified babel setup
- ESLint support for TS files


LucWollants (approved):
I tried creating a simple `.tsx` page and import a simple `.ts` type inside that file. That all worked so the setup seems fine.

I got one error
```
Error: Failed to load config "standard-with-typescript" to extend from.
```
But I think that is something related to PHPStorm

geoffreydhuyvetters:
@LucWollants Did you update the dependencies before you ran it?

LucWollants:
> Did you update the dependencies before you ran it?
Yes indeed, the first thing I did

PR #241, opened 2021-05-03
Add muted text variant

bertramakers:
### Added

- Added `MUTED` variant for `Text` component

---

Ticket: https://jira.uitdatabank.be/browse/III-3997

Will be needed on the dashboard for past events, for example on the current one in AngularJS:
Screen Shot 2021-05-03 at 13 27 54

### DIFF
--- a/src/ui/Text.js
+++ b/src/ui/Text.js
@@ -1,9 +1,23 @@
 import PropTypes from 'prop-types';
+import { css } from 'styled-components';
 import { Box, boxPropTypes, getBoxProps } from './Box';
+import { getValueFromTheme } from './theme';
 
-const Text = ({ as, children, className, ...props }) => {
+const getValue = getValueFromTheme('text');
+
+const TextVariants = {
+  REGULAR: 'regular',
+  MUTED: 'muted',
+};
+
+const mutedCSS = css`
+  color: ${getValue('muted.color')};
+`;
+
+const Text = ({ as, children, className, variant, ...props }) => {
+  const css = variant === TextVariants.MUTED ? mutedCSS : ``;
   return (
-    <Box as={as} className={className} {...getBoxProps(props)}>
+    <Box as={as} className={className} css={css} {...getBoxProps(props)}>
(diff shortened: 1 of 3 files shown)


src/ui/Text.js
```diff
@@ -1,9 +1,23 @@
 import PropTypes from 'prop-types';
```
simon-debruijn, line 1: I don't know for sure if this requires a new variant for `Text` since this code can be reduced to 
`<Text color={getValue('muted.color')} />`.

bertramakers: Then we have to do that everywhere where we want to have muted text?

src/ui/Text.js
```diff
@@ -1,9 +1,23 @@
...
+  color: ${getValue('muted.color')};
+`;
+
+const Text = ({ as, children, className, variant, ...props }) => {
+  const css = variant === TextVariants.MUTED ? mutedCSS : ``;
   return (
-    <Box as={as} className={className} {...getBoxProps(props)}>
+    <Box as={as} className={className} css={css} {...getBoxProps(props)}>
       {children}
     </Box>
```
simon-debruijn, line 22:
```suggestion
    <Box as={as} className={className} color={variant === TextVariants.MUTED && getValue('muted.color')} {...getBoxProps(props)}>
      {children}
    </Box>
```

src/ui/Text.js
bertramakers: Plus it opens up a lot of flexibility, instead of making sure that text is always one of two colors every color becomes possible?

simon-debruijn: Yes if muted text is used a lot a variant can be handy. But you should use the color prop over the css props, it is exposed by Box

src/ui/Text.js
bertramakers: See 1ed6b7c

simon-debruijn (approved):
🎖️ 🎖️ 🎖️ Noice

PR #244, opened 2021-05-05
Remove incorrect boxPropTypes usage

simon-debruijn:
### Fixed

- boxPropTypes, stackPropTypes and inlinePropTypes aren't objects with proptypes anymore but an array of strings

```
Warning: Failed prop type: Checkbox: prop type `35` is invalid; it must be a function, usually from the `prop-types` package, but received `string`.This often happens because of typos such as `PropTypes.function` instead of `PropTypes.func`.
```

### DIFF
--- a/src/components/pages/login/[language].js
+++ b/src/components/pages/login/[language].js
@@ -6,7 +6,7 @@ import { Trans, useTranslation } from 'react-i18next';
 import { css, keyframes } from 'styled-components';
 
 import { useCookiesWithOptions } from '@/hooks/useCookiesWithOptions';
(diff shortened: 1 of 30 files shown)


bertramakers:
Will we re-add these prop-types then when we convert these components to Typescript or how do you see this working in the future?

simon-debruijn:
> Will we re-add these prop-types then when we convert these components to Typescript or how do you see this working in the future?

All `prop-types` will be removed in favor of `TS types`.

bertramakers (approved):

PR #247, opened 2021-05-18
III-3944 - Use referer to know what url to return to after login

simon-debruijn:
### Changed

- Use referer to know what url to return to after login

---

Ticket: https://jira.uitdatabank.be/browse/III-3944

### DIFF
--- a/src/components/pages/login/index.js
+++ b/src/components/pages/login/index.js
@@ -13,7 +13,7 @@ export const getServerSideProps = ({ req }) => {
   referer.searchParams.delete('jwt');
 
   req.headers.referer =
-    referer.toString() ?? req.headers.referer ?? url.toString();
+    referer.toString() || req.headers.referer || url.toString();
 
   return {
     redirect: {
--- a/src/components/pages/login/index.js
+++ b/src/components/pages/login/index.js
@@ -7,15 +7,13 @@ export const getServerSideProps = ({ req }) => {
 
   const language = cookies.get('udb-language') ?? 'nl';
 
-  const url = new URL('http://' + req.headers.host + req.url);
+  const url = new URL(`http://${req.headers.host}${req.url}`);
 
   const referer = new URL(url.searchParams.get('referer') ?? '');
   referer.searchParams.delete('jwt');
 
   req.headers.referer =
-    referer.toString() ??
-    req.headers.referer ??
-    `http://${req.headers.host}${req.url}`;
+    referer.toString() ?? req.headers.referer ?? url.toString();
 
   return {
     redirect: {
--- a/src/components/pages/login/index.js
+++ b/src/components/pages/login/index.js
@@ -2,15 +2,25 @@ import { Cookies } from 'react-cookie';
 
 const Index = () => null;
 
-export const getServerSideProps = ({ req, params }) => {
+export const getServerSideProps = ({ req }) => {
   const cookies = new Cookies(req?.headers?.cookie);
 
   const language = cookies.get('udb-language') ?? 'nl';
 
+  const url = new URL('http://' + req.headers.host + req.url);
+
+  const referer = new URL(url.searchParams.get('referer') ?? '');
+  referer.searchParams.delete('jwt');
+
+  req.headers.referer =
+    referer.toString() ??
+    req.headers.referer ??
+    `http://${req.headers.host}${req.url}`;
+
   return {
     redirect: {
-      destination: `/login/${language}`,
-      permanent: true,
+      destination: `/login/${language}?referer=${req.headers.referer}`,
+      permanent: false,
     },
   };
 };
--- a/src/components/pages/login/index.js
+++ b/src/components/pages/login/index.js
@@ -9,11 +9,14 @@ export const getServerSideProps = ({ req }) => {
 
   const url = new URL(`http://${req.headers.host}${req.url}`);
 
-  const referer = new URL(url.searchParams.get('referer') ?? '');
-  referer.searchParams.delete('jwt');
+  const referer = url.searchParams.get('referer')
+    ? new URL(url.searchParams.get('referer'))
+    : undefined;
+
+  referer?.searchParams?.delete('jwt');
 
   req.headers.referer =
-    referer.toString() || req.headers.referer || url.toString();
+    referer?.toString() ?? (req.headers.referer || url.toString());
 
   return {
     redirect: {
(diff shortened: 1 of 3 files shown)


bertramakers (approved):
Nice improvement! 👍

src/utils/getApplicationServerSideProps.js
```diff
@@ -21,9 +21,12 @@ const getApplicationServerSideProps = (callbackFn) => async ({
     cookies.remove('user');
     cookies.remove('token');
 
+    req.headers.referer =
+      req.headers.referer ?? `http://${req.headers.host}${req.url}`;
```
brampauwelyn, line 25: Shouldn't this be https:// ?

```suggestion
      req.headers.referer ?? `https://${req.headers.host}${req.url}`;
```

src/components/pages/login/index.js
```diff
@@ -2,15 +2,23 @@ import { Cookies } from 'react-cookie';
 
 const Index = () => null;
 
-export const getServerSideProps = ({ req, params }) => {
+export const getServerSideProps = ({ req }) => {
   const cookies = new Cookies(req?.headers?.cookie);
 
   const language = cookies.get('udb-language') ?? 'nl';
 
+  const url = new URL(`http://${req.headers.host}${req.url}`);
```
brampauwelyn, line 10: Shouldn't this be https:// ?
```suggestion
  const url = new URL(`https://${req.headers.host}${req.url}`);
```

brampauwelyn (approved):

src/utils/getApplicationServerSideProps.js
simon-debruijn: Https won't work locally.

I wonder if this is correct: https://github.com/vercel/next.js/issues/2469#issuecomment-313194091
@bertramakers @willaerk

willaerk: I would allow both methods if possible:
* If "locally" means you just building and serving the frontend and accessing it through http://localhost:3000, then the referrer will contain http://
* If "locally" means building the frontend and serving it from your vagrant box, the referrer will contain https://

simon-debruijn: He is the creator of Next.js so I'm pretty sure he knows some things 😄

simon-debruijn: Would checking `process.env.NODE_ENV === 'production'` be enough here?

When it's a production build use `https` else `http`

willaerk: You could use the X-Forwarded-Proto header to get the protocol I think.

simon-debruijn: Locally I don't get that header:

```
headers {
  host: 'localhost:3000',
  'user-agent': 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10.15; rv:89.0) Gecko/20100101 Firefox/89.0',
  accept: 'text/html,application/xhtml+xml,application/xml;q=0.9,image/webp,*/*;q=0.8',
  'accept-language': 'en-US,nl;q=0.7,en;q=0.3',
  'accept-encoding': 'gzip, deflate',
  connection: 'keep-alive',
  'upgrade-insecure-requests': '1',
  'cache-control': 'max-age=0'
}
```

I'll check if i'ts there with production build

willaerk: If it is not there, it means there was no forwarding, and we can assume http:// was used I guess.

bertramakers: > When it's a production build use https else http

I remember from the permanent redirects discussion that you and @geoffreydhuyvetters  sometimes also run production mode on localhost so production mode does not necessarily mean https.

Aside from that, is the referer even relevant to this discussion? Isn't Bram's question about the URL to use when there's _no_ referer? So the default URL to redirect to after login?

The way I see it, we always redirect http:// requests to https:// anyway on the box & servers so http:// would work for both localhost and the Vagrant box and servers?

bertramakers: For example browse to http://www.uitdatabank.be/ and you automatically land on the https version.

bertramakers: > Locally I don't get that header

Do you mean "locally" on localhost or "locally" in the Vagrant box?

I'm pretty sure it will only be included when you run it in the Vagrant box. The nginx server in the box will set that header.

simon-debruijn: I meant on localhost by `locally`

simon-debruijn: Either way on the production build something is broken, so I'll have to fix it first before we merge

bertramakers: That would probably explain why the header is not there then, since there's no forwarding happening by a reverse proxy that handles nginx. If the header is not there I think you can safely assume the protocol is http like mentioned in the link you shared. If the header is there you can use the protocol specified in it.

But I'm still not sure if it even matters since we redirect everything from http:// to https:// anyway. @willaerk am I overlooking something in that assumption?

simon-debruijn: I've implemented a check for the **X-Forwarded-Proto** header:

https://github.com/cultuurnet/udb3-frontend/pull/247/commits/1cdefbb428aa828b89ec061eaaa74a562e68f585

willaerk: We do redirect http:// to https:// for all requests, but maybe it is not the most "elegant" solution to always rely on that being available. After all, the X-Forwarded-Proto header is used specifically for determining the protocol of the original request.

bertramakers: Alright is this logic correct then?

- If there's no x-forwarded-proto header the request was not forwarded by e.g. nginx and the app is _probably_* running on localhost so we can use `http://` in the link
- If the header is there, we can use the protocol defined in it

*But what if nginx stops sending this header for some reason?

This is also how @simon-debruijn has implemented it now (and also what I suggested above)

simon-debruijn: If the header isn't sent by nginx the redirect from http to https will still catch this no?

bertramakers: Yes definitely, I just think that the `X-Forwarded-Proto` in itself is also not the most elegant solution. 😄  But at least if one of the two measures breaks we can still rely on the other one.

PR #248, opened 2021-05-21
III-3944 redirect to original url (follow up)

simon-debruijn:
### Fixed

- Host url not being correct

---

Ticket: https://jira.uitdatabank.be/browse/III-3944


bertramakers (approved):
Nice, this also solves the discussion about the protocol 😄

PR #249, opened 2021-05-25
III-4016 - Referer bug (from google)

simon-debruijn:
### Fixed

- User cannot log on in UiTdatabank because of referer points to different webstie

---

Ticket: https://jira.uitdatabank.be/browse/III-4016

### DIFF
--- a/src/utils/getApplicationServerSideProps.js
+++ b/src/utils/getApplicationServerSideProps.js
@@ -25,11 +25,11 @@ const getApplicationServerSideProps = (callbackFn) => async ({
 
     const { origin } = absoluteUrl(req);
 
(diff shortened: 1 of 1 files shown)


bertramakers:
Won't this completely remove the redirect-to-original-page-after-login functionality again?

As opposed to the ticket's proposed optimization:
> The referer can only contain a url on the UiTdatabank domain and not a url from a different domain.

simon-debruijn:
> Won't this completely remove the redirect-to-original-page-after-login functionality again?
> 
> As opposed to the ticket's proposed optimization:
> 
> > The referer can only contain a url on the UiTdatabank domain and not a url from a different domain.

No this should still work. We're only not using the referer header anymore.

bertramakers (approved):
I see, thanks for the clarification!

PR #253, opened 2021-06-03
III-4028 Fix sentry config

simon-debruijn:
### Changed
- Switched to next.js SDK for Sentry

Note: This uses the sentry-cli which will generate source map files on a production build. This cli tool requires some env variables which need to be added to Vagrant, GitHub and Jenkins

---

Ticket: https://jira.uitdatabank.be/browse/III-4028


bertramakers:
Can you clarify for me why we still need the `sentry-cli` to upload source maps to Sentry, when we're building public source maps that get deployed along with the code in Next now?

My concern is that the source maps used by Sentry will often be incorrect now because we do builds for Storybook for every push in this repo, and apparently those also upload source maps to Sentry (seeing how they require the Sentry ENV variables). Also local builds in Vagrant would upload source maps to Sentry, polluting the release artifacts in Sentry even more with source maps that don't really exist in the actual acc/test/prod environments.

bertramakers:
The source map uploads so far can be found here by the way: https://sentry.io/organizations/publiq-vzw/releases/?project=5692505

I see now that they have an "adoption rate" so maybe (hopefully?) Sentry has some way to see what source map is relevant for which error.

But the questions remains: Do we need both source map uploads from `sentry-cli` and public source maps from Next?

LucWollants:
> Can you clarify for me why we still need the sentry-cli to upload source maps to Sentry, when we're building public source maps that get deployed along with the code in Next now?

We don't need them, but I can't find a way to disable it. I also can't find another way to init Sentry then by modifying Webpack which then trigger the sentry-cli.
We also include the source maps publicly now and they should take precedence over the uploaded source files.

@simon-debruijn first tried to use the react.js SDK but that seems to have issues with the server side code.

bertramakers:
Okay let's do it like this and see if it works or gives problems. Someone will need to add the ENV variables to Jenkins though, and @willaerk is not available today and me neither in the afternoon. So we should probably hold off with merging this until Monday.

LucWollants:
There is an option `dryRun` which disables the source map uploads, see https://github.com/getsentry/sentry-javascript/issues/3674

src/components/layouts/ErrorBoundary.js
```diff
@@ -16,6 +17,7 @@ class ErrorBoundaryComponent extends Component {
   render() {
     const { error } = this.state;
     if (error) {
+      Sentry.captureException(error);
```
bertramakers, line 20: I'm not sure if `render()` is the best place for this. We might be sending duplicate errors since this will get called whenever the component re-renders. @geoffreydhuyvetters what do you think?

next.config.js
```diff
@@ -47,6 +47,7 @@ const SentryWebpackPluginOptions = {
   //   urlPrefix, include, ignore
   // For all available options, see:
   // https://github.com/getsentry/sentry-webpack-plugin#options.
+  dryRun: true,
```
bertramakers, line 50: Nice 🤩

.github/workflows/chromatic.yml
```diff
@@ -4,6 +4,11 @@ on: push
 jobs:
   publish:
     runs-on: ubuntu-latest
+    env:
+      SENTRY_AUTH_TOKEN: ${{ secrets.SENTRY_AUTH_TOKEN }}
+      SENTRY_ORG: ${{ secrets.SENTRY_ORG }}
+      SENTRY_PROJECT: ${{ secrets.SENTRY_PROJECT }}
+      SENTRY_URL: ${{ secrets.SENTRY_URL }}
```
bertramakers, line 11: @LucWollants these can be removed now I think right?

.gitignore
```diff
@@ -54,4 +54,7 @@ node_modules/
 .env.local
 
 # next.js build output
-.next
\ No newline at end of file
+.next
+
+# sentry
+sentry.properties
```
bertramakers, line 60: This is also no longer needed I think?

next.config.js
```diff
@@ -30,3 +38,18 @@ module.exports = {
...
 };
+
+const SentryWebpackPluginOptions = {
+  // Additional config options for the Sentry Webpack plugin. Keep in mind that
+  // the following options are set automatically, and overriding them is not
+  // recommended:
+  //   release, url, org, project, authToken, configFile, stripPrefix,
+  //   urlPrefix, include, ignore
+  // For all available options, see:
+  // https://github.com/getsentry/sentry-webpack-plugin#options.
```
bertramakers, line 49: I'd remove the comments in this file (above as well), they are just copied from the documentation

src/components/layouts/index.js
```diff
@@ -133,12 +133,9 @@ const LayoutWrapper = ({ children }) => {
...
 
   return (
-    // eslint-disable-next-line node/handle-callback-err
-    <Sentry.ErrorBoundary
-      fallback={({ error }) => <ErrorFallback error={error} />}
-    >
+    <ErrorBoundary>
       <Layout>{children}</Layout>
-    </Sentry.ErrorBoundary>
+    </ErrorBoundary>
```
bertramakers, line 138: @simon-debruijn did you intentionally not wrap the `/login`, `/404`, and `/500` pages?

src/components/layouts/ErrorBoundary.js
LucWollants: There is indeed a better place and that is `ErrorFallback`

src/components/layouts/index.js
simon-debruijn: The **404** and **500** pages don't seem that logical to wrap with a ErrorBoundary necessarily, since they both appear when an error occurs. The login page seems like a smart thing to wrap with the ErrorBoundary indeed.

If we als add a ErrorBoundary around login, make sure to test if this doesn't cause any issues though 🙂 The ErrorFallback layout will be shown without sidebar if I'm correct.

geoffreydhuyvetters (approved):

src/components/layouts/index.js
LucWollants: I gonna first get this one merged. We can test that behaviour in another PR

PR #258, opened 2021-06-18
Use Promise.all

geoffreydhuyvetters:
### Fixed

- Use Promise.all instead of Promise.allSettled

We're currently using node 10 in the vragant box, Promise.allSettled is only supported in Node 12.19 and upwards

---

### DIFF
--- a/src/components/pages/dashboard.tsx
+++ b/src/components/pages/dashboard.tsx
@@ -471,7 +471,8 @@ const getServerSideProps = getApplicationServerSideProps(
       };
     }
 
-    await Promise.allSettled(
+    await Promise.all(
+      // @ts-expect-error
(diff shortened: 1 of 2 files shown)


src/components/pages/dashboard.tsx
```diff
@@ -471,7 +471,8 @@ const getServerSideProps = getApplicationServerSideProps(
       };
     }
 
-    await Promise.allSettled(
+    await Promise.all(
+      // @ts-expect-error
```
simon-debruijn, line 475: why was this expect-error needed?

geoffreydhuyvetters: Because it was throwing a TS error? 😅 (prefetch / normal fetch)

jdrieghe (approved):

PR #260, opened 2021-06-22
III-3961 - organizer tab last improvements

simon-debruijn:
### Added

- delete action for organizer
- no_items translations

### Changed

- dropdown items to use Button and Link behind the scenes

### Deleted

- unused file createDashboard

---

Ticket: https://jira.uitdatabank.be/browse/III-3961

### DIFF
--- a/src/components/pages/dashboard.tsx
+++ b/src/components/pages/dashboard.tsx
@@ -418,7 +418,8 @@ const Dashboard = (): any => {
         {t('dashboard.welcome')}, {user?.username}
       </Page.Title>
       <Page.Actions>
-        <Link href="/create" css="text-transform: lowercase;">
+        {/* TODO: Should become /create */}
+        <Link href="/event" css="text-transform: lowercase;">
           {t('dashboard.create')}
         </Link>
       </Page.Actions>
--- a/src/components/pages/dashboard.tsx
+++ b/src/components/pages/dashboard.tsx
@@ -245,6 +245,9 @@ const OrganizerRow = ({
         <Link href={editUrl} variant={LinkVariants.BUTTON_SECONDARY} key="edit">
           {t('dashboard.actions.edit')}
         </Link>,
+        <Dropdown.Item onClick={() => onDelete(organizer)} key="delete">
+          {t('dashboard.actions.delete')}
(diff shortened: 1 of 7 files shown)


src/ui/Dropdown.tsx
```diff
@@ -68,11 +68,40 @@ type ItemProps = {
...
+        padding={0}
+        css={`
+          .btn {
+            flex: 1;
+            border: none;
+          }
+        `}
+      >
+        {children}
+      </BootstrapDropdown.Item>
```
bertramakers, line 99: What is the reason of doing it like this exactly? I don't really see the advantage, only some workarounds to make it look the same and a lot more code 😅

src/components/pages/dashboard.tsx
```diff
@@ -418,7 +418,8 @@ const Dashboard = (): any => {
         {t('dashboard.welcome')}, {user?.username}
       </Page.Title>
       <Page.Actions>
-        <Link href="/create" css="text-transform: lowercase;">
+        {/* TODO: Should become /create */}
```
bertramakers, line 421: You could add a redirect from `/create` to `/event` for now so you can already start using the new path instead of adding a `TODO` (which are often never fixed later)

src/ui/Dropdown.tsx
bertramakers: Plus we should test if keyboard navigation inside the dropdown still works as expected now.

simon-debruijn: For links this is needed, we want to use nuxt-link behind the scenes, so thath the page doesn't have to refresh when you click the url.

src/components/pages/dashboard.tsx
simon-debruijn: Good point

src/ui/Dropdown.tsx
geoffreydhuyvetters: the normal behaviour is an anchor tag = full page refresh

bertramakers: Okay I see!

geoffreydhuyvetters (approved):

PR #261, opened 2021-06-22
Fix prefetching on serverside (followup)

simon-debruijn:
### Fixed

- token being unknown in the request


src/utils/getApplicationServerSideProps.js
```diff
@@ -88,6 +88,10 @@ const getApplicationServerSideProps = (callbackFn) => async ({
     };
   }
 
+  // set token in req.headers.cookie so that the token is known when prefetching a request
+  cookies.set('token', query.jwt);
```
geoffreydhuyvetters, line 92: Shouldn't we add some checks here? if there is a token in the URL and it's different than the one we currently have in the cookie?

geoffreydhuyvetters (approved):

PR #264, opened 2021-06-24
Add footer to dashboard page

geoffreydhuyvetters:
### Added

- Extracted Footer from login page and reused it on the dashboard page
- added a boolean to disable the language switcher on the dashboard page

Screenshot 2021-06-24 at 10 52 56

### DIFF
--- a/src/components/pages/dashboard/NewsletterSingupForm.tsx
+++ b/src/components/pages/dashboard/NewsletterSingupForm.tsx
@@ -20,7 +20,7 @@ import { Title } from '@/ui/Title';
 const isEmail = (value: string) =>
   yup.string().required().email().isValidSync(value);
 
-const NewsletterSignupForm = () => {
+const NewsletterSignupForm = (props) => {
   const formRef = useRef<HTMLFormElement>();
   const [email, setEmail] = useState('');
   const [isValid, setIsValid] = useState(true);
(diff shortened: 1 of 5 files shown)


src/components/pages/dashboard/NewsletterSingupForm.tsx
```diff
@@ -20,7 +20,7 @@ import { Title } from '@/ui/Title';
 const isEmail = (value: string) =>
   yup.string().required().email().isValidSync(value);
 
-const NewsletterSignupForm = () => {
+const NewsletterSignupForm = (props) => {
```
simon-debruijn, line 23: Add type of StackProps

simon-debruijn (approved):

bertramakers:
Very small nit, but the font in the SVG seems to be wrong (Times New Roman?) compared to the icon on the current dashboard.

Old:
Screen Shot 2021-06-24 at 14 25 10

New:
Screen Shot 2021-06-24 at 14 25 15

Can also be fixed in a follow-up PR if it's too complicated to do it in this one of course.

bertramakers:
Apart from that it looks very good 👍

bertramakers:
Oh sorry apparently the email signup form isn't even part of this PR, it's just the footer below. All good for this PR then 👍

bertramakers (approved):

PR #269, opened 2021-06-28
Remove delete button for organizer

simon-debruijn:
### Removed

- Delete button for organizer
---

Ticket: https://jira.uitdatabank.be/browse/III-3961

Screenshot:
Schermafbeelding 2021-06-28 om 10 07 15


simon-debruijn:
When we rewrite the api hooks to be all separate instead of using useMemo we can delete the delete organizer hook

geoffreydhuyvetters (approved):

PR #270, opened 2021-06-28
Upgrade next to 11.0.1

simon-debruijn:
### Changed

- Upgrade next to 11.0.1

Guide followed: https://nextjs.org/docs/upgrading#upgrading-from-version-10-to-11


geoffreydhuyvetters:
There seems to be an issue with Storybook when building now

simon-debruijn:
> There seems to be an issue with Storybook when building now

I'm looking onto it, it's the typescript version I think

brampauwelyn (approved):

PR #274, opened 2021-06-29
III-3789 - fix bootstrap styles serverside

simon-debruijn:
### Fixed

- Server Side render not including all styles (bootstrap)

We're now also making use of https://nextjs.org/docs/basic-features/font-optimization

---

Ticket: https://jira.uitdatabank.be/browse/III-3789

Screenshot BEFORE:
Schermafbeelding 2021-06-29 om 16 21 03

Screenshot AFTER:
Schermafbeelding 2021-06-29 om 16 24 50

### DIFF
--- a/src/components/pages/_document.js
+++ b/src/components/pages/_document.js
@@ -42,7 +42,7 @@ class Document extends NextDocument {
           {/* eslint-disable-next-line @next/next/no-page-custom-font */}
           <link
             key="fonts"
-            href="https://fonts.googleapis.com/css?family=Open+Sans:400italic,600italic,700italic,300,400,700&display=optional"
+            href="https://fonts.googleapis.com/css?family=Open+Sans:300,400,700&display=optional"
             rel="stylesheet"
           />
           <link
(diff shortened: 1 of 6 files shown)


src/components/pages/_document.js
```diff
@@ -27,6 +27,39 @@ class Document extends NextDocument {
...
+            key="icon"
+            rel="icon"
+            type="image/png"
+            sizes="32x32"
+            href="/favicon.png"
+          />
+          {/* eslint-disable-next-line @next/next/no-page-custom-font */}
+          <link
+            key="fonts"
+            href="https://fonts.googleapis.com/css?family=Open+Sans:400italic,600italic,700italic,300,400,700&display=optional"
```
brampauwelyn, line 45: Do you need all these different italic styles?

brampauwelyn (approved):

geoffreydhuyvetters (approved):

src/components/pages/_document.js
simon-debruijn: Good question, I'll take a look. These were used in the old app that's why we import them, but I think for now we don't use them

PR #282, opened 2021-07-02
Add redirects for dashboard

geoffreydhuyvetters:
### Added

- redirects for /events /places and /organizers

### DIFF
--- a/src/pages/events/index.js
+++ b/src/pages/events/index.js
@@ -1,0 +1,1 @@
+export { default, getServerSideProps } from '@/pages/redirect';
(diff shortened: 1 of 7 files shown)


geoffreydhuyvetters:
Closes because getServerSideProps on a 404 doesn't work when building a production build. Need to add indexes at the root

src/pages/events/index.js
```diff
@@ -0,0 +1 @@
+export { default, getServerSideProps } from '@/pages/redirect';
```
simon-debruijn, line 1: funky 😄

simon-debruijn (approved):

src/pages/events/index.js
geoffreydhuyvetters: Why? We're doing this on all our pages :)

PR #286, opened 2021-07-08
III-4095 - Add sidebar link filminvoer

simon-debruijn:
### Added

- path to `/manage/movies/create`
- Page with page title
- redirect
- menu item for movies

### Changed

- transformed multiple components to typescript

### Fixed

- badge in mobile view

---

Ticket: https://jira.uitdatabank.be/browse/III-4095

Screenshot:
Schermafbeelding 2021-07-08 om 15 33 21


src/components/layouts/Sidebar.tsx
```diff
@@ -32,8 +34,40 @@ const getValueForMenuItem = getValueFromTheme('menuItem');
 const getValueForSidebar = getValueFromTheme('sidebar');
 const getValueForMenu = getValueFromTheme('menu');
 
+const PermissionTypes = {
+  AANBOD_BEWERKEN: 'AANBOD_BEWERKEN',
```
brampauwelyn, line 38: Did you consider using english names for the keys?

src/redirects.js
```diff
@@ -33,7 +33,7 @@ const createDashboardRedirects = (environment) => {
   ];
 };
 
-const getRedirects = (environment) => [
+const getRedirects = (environment, language) => [
```
brampauwelyn, line 36: Would it be an idea to add a default value?

brampauwelyn (approved):

src/redirects.js
brampauwelyn: + no types here? 😃

src/components/layouts/Sidebar.tsx
simon-debruijn: I would also prefer english names, but here we use the same ones comming from the udb3-backend also in dutch

src/redirects.js
simon-debruijn: Haha I'l typescriptify it for you

And good idea

PR #289, opened 2021-07-09
Use local strorage for seen announcements

simon-debruijn:
Because all cookies get sent to the server, this is not needed for seenAnnouncements

### Added

- useLocalStorage hook

### Changed

- use local storage instead of cookies for seenAnnouncements

### Removed

- profilePicture via cookies


src/hooks/useLocalStorage.ts
```diff
@@ -0,0 +1,37 @@
...
+};
+
+const useLocalStorage = () => {
+  const [localStorageApi] = useState({
+    /**
+     * Returns the current value associated with the given key, or null if the given key does not exist in the list associated with the object.
+     */
+    getItem: (key: Key) => {
+      const value = JSON.parse(localStorage.getItem(key));
+      if (isSeenAnnouncements(key, value)) {
```
brampauwelyn, line 21: I wonder: is the 'announcement' code in this hook not too specific?

brampauwelyn (approved):

src/hooks/useLocalStorage.ts
simon-debruijn: The 'announcement' code is there to be able to automaticly infer the type when using getItem. When there are multiple things stored in the localStorage, this hook will be able to tell the type based on the key your passing to getItem 🙂 

[Typeguards are used for this](https://www.typescriptlang.org/docs/handbook/advanced-types.html#type-guards-and-differentiating-types)

PR #292, opened 2021-07-23
III-3540 - Custom calendar

simon-debruijn:
### Added

- TimeTable component
- DatePicker component
- DatePeriodPicker component
- Step 2 movies/create page

---

Ticket: https://jira.uitdatabank.be/browse/III-3540

Schermafbeelding 2021-08-11 om 10 00 38


src/components/pages/manage/movies/create.tsx
```diff
@@ -1,91 +1,156 @@
...
-          </Text>
-          <Button variant={ButtonVariants.LINK} onClick={handleClearTheme}>
-            {t('movies.create.actions.change_theme')}
-          </Button>
-        </Inline>
-      )}
-    </Inline>
+    <Stack spacing={5} {...getStackProps(props)}>
+      <Inline spacing={3} flexWrap="wrap" maxWidth="70rem">
+        {movieState.context.theme === null ? (
```
geoffreydhuyvetters, line 77: not sure if this is the preferred way to do this.
seems a bit weird to create designated states and then write ternaries based on values in the context

simon-debruijn: It's probably better to base theis logic on the current state it's in, but right now we only have the idle state since we simplified the movie state machine

geoffreydhuyvetters (approved):

PR #296, opened 2021-08-06
Add new plugin and rule to auto fix unused imports

simon-debruijn:
### Added

- new plugin and rule to auto fix unused imports


brampauwelyn (changes requested):
👍

brampauwelyn (approved):

PR #302, opened 2021-08-27
Fix Link story

simon-debruijn:
### Fixed

- Link story


LucWollants (approved):
The storybook works now

PR #306, opened 2021-09-02
Change steps structure on filminvoer

geoffreydhuyvetters:
### Changed

- create a Step wrapper to use on every step
- move state machine to create page


src/components/pages/manage/movies/create.tsx
```diff
@@ -1,67 +1,127 @@
 import { useMachine } from '@xstate/react';
-import type { MovieContext, MovieEvent } from 'machines/movie';
-import { movieMachine } from 'machines/movie';
```
bertramakers, line 3: What is the reasoning that the state machine had to be moved into this file instead of being in a separate file?

src/components/pages/manage/movies/create.tsx
```diff
@@ -70,22 +130,22 @@ const Create = () => {
...
-                sendMovieEvent={sendMovieEvent}
-              />
-            </Step>
-          ),
-        )}
+        {steps.map((Step, index) => (
+          <Step
+            movieState={movieState}
+            sendMovieEvent={sendMovieEvent}
+            stepNumber={index + 1}
```
bertramakers, line 145: Why do we need this `stepNumber` prop if each Step is its own component that has the number in the name? Can't we hardcode the step number of each step in their own component then?

bertramakers:
Good improvement overall I think. It did take me a while to understand what was changed in this PR (and why) because everything was in a single big commit. I believe it would have been easier to review if you had split this up in multiple commits that followed the steps of the refactor 😉

geoffreydhuyvetters:
@bertramakers My bad, I got a bit carried away 😊

src/components/pages/manage/movies/create.tsx
geoffreydhuyvetters: Yeah, absolutely. Makes a lot of sense.

src/components/pages/manage/movies/create.tsx
geoffreydhuyvetters: We're only using it on this page, so to me, it made a lot more sense to define it there.

bertramakers: Okay was just wondering!

bertramakers:
No worries, it happens to me as well 😉 
I'll hold off on approving until you've removed the `stepNumber` property as agreed on.

bertramakers (approved):
Nice work! 👌

PR #311, opened 2021-09-08
III-4045 Add booking availability to single calendar status page

LucWollants:
### Added
- Added booking availability to single calendar status page

### Changed
- Improve naming for the BookingAvailabilityType prop of BookingAvailabilityForm

---

Ticket: https://jira.uitdatabank.be/browse/III-4045


src/components/AvailabilityPageSingle.js
```diff
@@ -24,34 +26,71 @@ const AvailabilityPageSingle = ({ offer, error, useChangeStatus }) => {
...
+  const createMutationPayload = () => {
+    if (offer.calendarType === CalendarType.SINGLE) {
+      return {
+        eventId: offerId,
+        subEventIds: [0],
+        subEvents: offer?.subEvent,
+        type,
+        reason: {
+          ...(offer.status.type === type && offer.status.reason),
+          [i18n.language]: reasonMainLanguage || undefined,
```
simon-debruijn, line 72: The variable name `reasonMainLanguage` isn't correct I think. Because the reason gets saved separately for the language that is used at the moment.

src/components/AvailabilityPageSingle.js
```diff
@@ -24,34 +26,71 @@ const AvailabilityPageSingle = ({ offer, error, useChangeStatus }) => {
...
-  const handleSuccessChangeStatus = () =>
-    router.push(`/${offerType}/${offerId}/preview`);
+  const handleSuccess = () => router.push(`/${offerType}/${offerId}/preview`);
 
   const changeStatusMutation = useChangeStatus({
-    onSuccess: handleSuccessChangeStatus,
+    onSuccess: handleSuccess,
   });
 
+  const createMutationPayload = () => {
```
simon-debruijn, line 63: Nice idea to put these in a function 🙂

src/components/AvailabilityPageSingle.js
```diff
@@ -64,34 +103,29 @@ const AvailabilityPageSingle = ({ offer, error, useChangeStatus }) => {
...
-                    changeStatusMutation.mutate({
-                      id: offerId,
-                      type,
-                      reason: {
-                        ...(offer.status.type === type && offer.status.reason),
-                        [i18n.language]: reason || undefined,
-                      },
-                    });
-                  }
+                  changeStatusMutation.mutate(createMutationPayload());
```
simon-debruijn, line 128: When I test this, the reason doesn't seem to get properly set

simon-debruijn (changes requested):
The reason isn't saved properly when I test this

src/components/AvailabilityPageSingle.js
LucWollants: It's indeed not the main language of the offer, maybe `reasonUserLanguage` is a better name?

simon-debruijn: I mean, it is just the reasing you're saving in state. With those variable names it looks like it is a language like `nl` or `fr` that you're saving. I would revert it to `reason` or if that is not clear to you change it to something like `reasonInCurrentLanguage`

LucWollants: Ah yes, good point. I will change it to `reasonInCurrentLanguage`

src/components/AvailabilityPageSingle.js
LucWollants: There is an edge case when setting the status with a reason on an event with single calendar type. In that case the reason is only present on the sub event and not on the top level. The UI only uses the top level to get the status from a single calendar type.

Proposed solution: inside the backend also apply the status reason from the sub event on the top level.

LucWollants: This PR fixes it https://github.com/cultuurnet/udb3-silex/pull/733

LucWollants: @simon-debruijn I guess it's fine to already merge this now that the edge case is clear?

simon-debruijn (approved):

src/components/AvailabilityPageSingle.js
simon-debruijn: Ok ✅

PR #313, opened 2021-09-10
III-4045 UI improvements to match mockups for availability of single calendar type

LucWollants:
### Changed
- Make radio button group label optional
- Add title component for status and booking on the single calendar page

---

Ticket: https://jira.uitdatabank.be/browse/III-4045

### DIFF
--- a/src/ui/RadioButtonGroup.js
+++ b/src/ui/RadioButtonGroup.js
@@ -15,7 +15,9 @@ const RadioButtonGroup = ({
 }) => {
   return (
     <Stack className={className} as="div" spacing={3} {...getStackProps(props)}>
-      {groupLabel && <Label variant={LabelVariants.BOLD}>{groupLabel}</Label>}
+      {groupLabel !== '' && (
+        <Label variant={LabelVariants.BOLD}>{groupLabel}</Label>
+      )}
       <Stack role="radiogroup" as="ul" spacing={2}>
         {items.map((item) => (
           <RadioButtonWithLabel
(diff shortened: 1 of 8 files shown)


simon-debruijn (commented):
These changes look good. But first we should fix https://github.com/cultuurnet/udb3-frontend/pull/311

src/ui/RadioButtonGroup.js
```diff
@@ -15,7 +15,9 @@ const RadioButtonGroup = ({
 }) => {
   return (
     <Stack className={className} as="div" spacing={3} {...getStackProps(props)}>
-      {groupLabel && <Label variant={LabelVariants.BOLD}>{groupLabel}</Label>}
+      {groupLabel ? (
+        <Label variant={LabelVariants.BOLD}>{groupLabel}</Label>
+      ) : null}
```
simon-debruijn, line 20:
```suggestion
      {groupLabel && <Label variant={LabelVariants.BOLD}>{groupLabel}</Label>}
```
Btw this code would also work, as it was. Conditions resulting in false are not rendered by React

LucWollants: I tried that approach first but then I got this error
```
Error: Element type is invalid: expected a string (for built-in components) or a class/function (for composite components) but got: undefined. You likely forgot to export your component from the file it's defined in, or you might have mixed up default and named imports.
```

geoffreydhuyvetters: It's because it can return `false`, which a Stack/Inline can't handle

geoffreydhuyvetters (approved):
Nice one 💪

LucWollants:
> Nice one 💪
That's because of the nice help during the pair programming 🙏

PR #316, opened 2021-09-30
Feature/iii 4233 refactor button

brampauwelyn:
### Added

- [Added ...]

### Changed

- [Changed ...]

### Removed

- [Removed ...]

### Fixed

- [Fixed ...]

---

Ticket: https://jira.uitdatabank.be/browse/III-4233

### DIFF
--- a/src/ui/Button.tsx
+++ b/src/ui/Button.tsx
@@ -9,36 +9,14 @@ import type { Icons } from './Icon';
 import { Icon } from './Icon';
 import type { InlineProps } from './Inline';
 import { getInlineProps, Inline } from './Inline';
-import { Link } from './Link';
+import { linkCSS } from './Link';
 import { Spinner, SpinnerSizes, SpinnerVariants } from './Spinner';
 import { Text } from './Text';
 import { getValueFromTheme } from './theme';
 
-const BootStrapVariants = {
-  PRIMARY: 'primary',
-  SECONDARY: 'secondary',
-  SUCCESS: 'success',
-  DANGER: 'danger',
-} as const;
-
-const ButtonVariants = {
-  ...BootStrapVariants,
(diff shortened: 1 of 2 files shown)


simon-debruijn (approved):
Looks good to me 🚀 A big improvement imo.

Feel free to add other reviewers 🙂

bertramakers:
I'm a bit out of the loop of the front developments lately, so my JS-specific feedback might not be that useful. But as general feedback I'd suggest to add some more context to this PR. Neither the ticket or the PR description describe the exact goal of the refactor. If it's just about removing those props I think you can just remove the "Refactor Button component" from the changelog since those removals are already mentioned. But if it's about something more it would help if you could clarify what the goal was and how that is now reached with the changes in this PR. Not necessarily for me but also for @LucWollants or @geoffreydhuyvetters so they have an easier time reviewing. Not saying that the changes are not good or useful by the way, that's hard for me to judge lately. 😃

simon-debruijn:
> I'm a bit out of the loop of the front developments lately, so my JS-specific feedback might not be that useful. But as general feedback I'd suggest to add some more context to this PR. Neither the ticket or the PR description describe the exact goal of the refactor. If it's just about removing those props I think you can just remove the "Refactor Button component" from the changelog since those removals are already mentioned. But if it's about something more it would help if you could clarify what the goal was and how that is now reached with the changes in this PR. Not necessarily for me but also for @LucWollants or @geoffreydhuyvetters so they have an easier time reviewing. Not saying that the changes are not good or useful by the way, that's hard for me to judge lately. 😃

The logic inside the Button component was more complex than it needed to be. So this PR is meant as a simplification and a streamlining of the Button component.

Geoffrey and I thought it would be a good challenge for Bram to get used to the udb3 setup

PR #325, opened 2021-10-29
III-4254 - login in german

simon-debruijn:
### Added

- German as a supported language

### Changed

- German to be supported when login in

---

Ticket: https://jira.uitdatabank.be/browse/III-4254


src/i18n/index.ts
```diff
@@ -15,6 +16,7 @@ i18n
     resources: {
       fr: { translation: fr },
       nl: { translation: nl },
+      de: { translation: de },
```
geoffreydhuyvetters, line 19: I guess we could also use `supportedLanguages` here

src/components/Footer.tsx
```diff
@@ -127,6 +127,12 @@ const Footer = ({
           >
             Français
           </LanguageSwitcherButton>
+          <LanguageSwitcherButton
+            variant={ButtonVariants.UNSTYLED}
+            onClick={handleChangeLanguage('de')}
+          >
+            Deutch
```
brampauwelyn, line 134: Could we store these in a constant or enum?

brampauwelyn (approved):

src/components/Footer.tsx
simon-debruijn: DONE

PR #349, opened 2022-02-10
III-4490 - Only show sorting select on event tab

brampauwelyn:
### Changed
- Only show sorting select on event tab

---

Ticket: https://jira.uitdatabank.be/browse/III-4490


bertramakers (approved):
Did a quick manual test and this works as expected! The sort URL parameter is also not included when you sorted the event tab and then switch to another tab, nice attention to detail 👌

brampauwelyn:
> Did a quick manual test and this works as expected! The sort URL parameter is also not included when you sorted the event tab and then switch to another tab, nice attention to detail 👌

Thx @bertramakers ;)

PR #353, opened 2022-02-23
III-4583 - Styling opimizations dashboard

simon-debruijn:
### Added

- Select component
- Select component story

### Changed

- hide extra info on smaller screens

### Fixed

- alignment dashboard
- Select forwardedAs

---

Ticket: https://jira.uitdatabank.be/browse/III-4583

Screenshots:

Schermafbeelding 2022-02-23 om 17 12 04
Schermafbeelding 2022-02-23 om 17 11 57


src/components/pages/dashboard/index.tsx
```diff
@@ -460,6 +462,8 @@ const Dashboard = (): any => {
     'availableTo_asc',
   ];
 
+  const shouldShowExtraInfo = useMatchBreakpoint(Breakpoints.L);
```
geoffreydhuyvetters, line 465: No way to achieve this with CSS? This will only be triggered on the client-side now, causing a difference between the server and client render.

simon-debruijn: I think we could achieve similar results using something like 
`display={{
  default: 'flex',
  l: 'hidden'
}}`

geoffreydhuyvetters (approved):

PR #360, opened 2022-03-03
Setup Create page

geoffreydhuyvetters:
### Added

- create page with type / theme and calendar step placeholders

### DIFF
--- a/src/pages/create/EventForm.tsx
+++ b/src/pages/create/EventForm.tsx
@@ -13,7 +13,7 @@ import { CalendarStep } from './CalendarStep';
 import { ThemeStep } from './ThemeStep';
 import { TypeStep } from './TypeStep';
 
-type EventType = 'event' | 'place';
+type OfferType = 'event' | 'place';
 
 type FormData = {
   type: EventType;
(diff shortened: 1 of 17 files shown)


src/pages/create/EventForm.tsx
```diff
@@ -0,0 +1,70 @@
...
+import type { Calendar } from '@/hooks/api/events';
+import type { StepsConfiguration } from '@/pages/Steps';
+import { Steps } from '@/pages/Steps';
+import { Page } from '@/ui/Page';
+
+import { CalendarStep } from './CalendarStep';
+import { ThemeStep } from './ThemeStep';
+import { TypeStep } from './TypeStep';
+
+type EventType = 'event' | 'place';
```
simon-debruijn, line 16:
```suggestion
type OfferType = 'event' | 'place';
```

Also the file should better be called `OfferForm`, since this is the correct term we use for something that is either a event or place

geoffreydhuyvetters: It's still in draft, I'm changing a lot of things as we speak

brampauwelyn (approved):

PR #363, opened 2022-03-10
Feature/create place event toggle

brampauwelyn:
### Added
- ToggleBox component 
- Integrate ToggleBox component in step1 of new entry form
Screenshot 2022-03-10 at 11 39 11

Screenshot 2022-03-10 at 13 57 36

- [ ] To check with Corneel if we are going to use the same Icons (building / calendar) as on the current entry form or if we can use font awesome icons...

### Changed
- renamed translation label `OfferTypes` to `OfferCategories` since this is the correct terminology.

### DIFF
--- a/src/i18n/de.json
+++ b/src/i18n/de.json
@@ -246,7 +246,7 @@
     "rowsSelectedCount": "{{count}} Reihe ausgewählt",
     "rowsSelectedCount_plural": "{{count}} Reihen ausgewählt"
   },
-  "offerTypes": {
+  "offerCategories": {
     "0.0.0.0.0": "Ausstellung",
     "0.12.0.0.0": "Tag der offenen Tür",
     "0.14.0.0.0": "Monument",
--- a/src/i18n/de.json
+++ b/src/i18n/de.json
@@ -246,6 +246,10 @@
     "rowsSelectedCount": "{{count}} Reihe ausgewählt",
     "rowsSelectedCount_plural": "{{count}} Reihen ausgewählt"
   },
+  "offerTypes": {
+    "event": "Veranstaltung",
+    "place": "Ort"
+  },
   "offerCategories": {
     "0.0.0.0.0": "Ausstellung",
     "0.12.0.0.0": "Tag der offenen Tür",
--- a/src/pages/create/TypeStep.tsx
+++ b/src/pages/create/TypeStep.tsx
@@ -1,15 +1,41 @@
 import { Controller } from 'react-hook-form';
+import { useTranslation } from 'react-i18next';
 
+import { OfferType } from '@/constants/OfferType';
 import type { StepProps } from '@/pages/Steps';
-import { Text } from '@/ui/Text';
+import { Icons } from '@/ui/Icon';
+import { Inline } from '@/ui/Inline';
+import { ToggleBox } from '@/ui/ToggleBox';
 
 const TypeStep = ({ control, field }: StepProps<FormData>) => {
+  const { t } = useTranslation();
+
   return (
     <Controller
       control={control}
       name={field}
       render={({ field }) => {
-        return <Text>{field.name}</Text>;
+        return (
+          <Inline
+            spacing={5}
+            alignItems="center"
+            justifyContent="space-between"
+          >
+            <ToggleBox
+              onClick={() => field.onChange(OfferType.EVENT)}
+              active={field.value === OfferType.EVENT}
+              icon={Icons.CALENDAR_ALT}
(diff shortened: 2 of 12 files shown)


src/pages/create/TypeStep.tsx
```diff
@@ -1,15 +1,41 @@
...
+            alignItems="center"
+            justifyContent="space-between"
+          >
+            <ToggleBox
+              onClick={() => field.onChange(OfferType.EVENT)}
+              active={field.value === OfferType.EVENT}
+              icon={Icons.CALENDAR_ALT}
+              text={t('offerTypes.event')}
+            />
+            <span css="font-style: italic;">{t('steps.offerTypeStep.or')}</span>
```
geoffreydhuyvetters, line 30: I would use a Text component here, if we don't have a good variant we can add some css on the element itself

src/pages/dashboard/index.page.tsx
```diff
@@ -134,7 +134,7 @@ const EventRow = ({ item: event, onDelete, ...props }: EventRowProps) => {
   const previewUrl = `/event/${parseOfferId(event['@id'])}/preview`;
   const typeId = event.terms.find((term) => term.domain === 'eventtype')?.id;
   // The custom keySeparator was necessary because the ids contain '.' which i18n uses as default keySeparator
-  const eventType = t(`offerTypes*${typeId}`, { keySeparator: '*' });
+  const eventType = t(`offerCategories*${typeId}`, { keySeparator: '*' });
```
geoffreydhuyvetters, line 137: The correct terminology is eventType here

src/pages/create/TypeStep.tsx
```diff
@@ -1,15 +1,41 @@
...
+          <Inline
+            spacing={5}
+            alignItems="center"
+            justifyContent="space-between"
+          >
+            <ToggleBox
+              onClick={() => field.onChange(OfferType.EVENT)}
+              active={field.value === OfferType.EVENT}
+              icon={Icons.CALENDAR_ALT}
+              text={t('offerTypes.event')}
```
geoffreydhuyvetters, line 28: I would all of the translations into the step object.

src/i18n/de.json
```diff
@@ -303,6 +303,15 @@
     "1.7.14.0.0": "Mehrere Filmgenres",
     "1.7.15.0.0": "Thriller"
   },
+  "steps": {
+    "offerTypes": {
```
geoffreydhuyvetters, line 307: I meant in the actual step itself. Sorry for the confusion 😅

brampauwelyn: No worries! I added it on the root level with the idea that we'll probably need it outside of the step as well. 
But adding it here also makes sense...

simon-debruijn: I would opt for offerTypes on the root level since the translation won't change depending on the location where we use them

geoffreydhuyvetters: I'm just a big fan of keeping all translations for 1 component under 1 key, yes there might be repetition, but this makes everything so much easier to move around etc

geoffreydhuyvetters: Although I understand that this opens up questions like 'what do we do with the eventype labels?" etc

simon-debruijn: It doesn't matter to me that much either way, so just merge it 😛

geoffreydhuyvetters: I agree. Bigger fish to fry 😅

brampauwelyn: 🐟

src/ui/ToggleBox.tsx
```diff
@@ -0,0 +1,85 @@
...
+};
+
+const ToggleBox = ({ children, onClick, active, icon, text }: Props) => {
+  return (
+    <Box
+      onClick={onClick}
+      padding={5}
+      display="flex"
+      flexWrap="wrap"
+      flexDirection="column"
```
geoffreydhuyvetters, line 23: If you want to give Box a flexDirection you should use Stack or Inline

src/ui/ToggleBox.tsx
```diff
@@ -0,0 +1,85 @@
...
+      {icon && (
+        <Icon
+          name={icon}
+          color={getValue('iconColor')}
+          width={30}
+          height={30}
+        />
+      )}
+      {text && (
+        <span
```
geoffreydhuyvetters, line 59: Try to avoid using plain HTML elements. Use Text and see if we can reuse an existing variant. We want to limit the amount of arbitrary CSS.

PR #367, opened 2022-03-16
III-4478 - video upload

simon-debruijn:
### Added

- [Added ...]

### Changed

- [Changed ...]

### Removed

- [Removed ...]

### Fixed

- [Fixed ...]

---

### DIFF
--- a/src/hooks/api/events.ts
+++ b/src/hooks/api/events.ts
@@ -552,11 +552,30 @@ const usePublish = (configuration = {}) =>
     ...configuration,
   });
 
+const addVideoToEvent = async ({ headers, eventId, url, language }) =>
+  fetchFromApi({
+    path: `/events/${eventId}/videos`,
+    options: {
+      method: 'POST',
+      headers,
+      body: JSON.stringify({
+        url,
+        language,
+      }),
+    },
+  });
+
+const useAddVideoToEvent = (configuration = {}) =>
+  useAuthenticatedMutation({
+    mutationFn: addVideoToEvent,
+    ...configuration,
+  });
 export {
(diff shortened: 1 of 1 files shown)


src/pages/PictureUploadBox.tsx
```diff
@@ -0,0 +1,145 @@
...
+      >
+        <Stack
+          spacing={2}
+          maxHeight={380}
+          css={`
+            overflow: auto;
+          `}
+        >
+          {images.map((image, index, imagesArr) => {
+            const thumbnailSize = 80;
```
brampauwelyn, line 64: I think you can define this `const`outside the `.map`,  since it's the same for all the thumbnails.

src/pages/VideoUploadBox.tsx
```diff
@@ -0,0 +1,126 @@
...
+      >
+        <Stack
+          spacing={2}
+          maxHeight={380}
+          css={`
+            overflow: auto;
+          `}
+        >
+          {videos.map(({ id, url, thumbnailUrl }, index, imagesArr) => {
+            const thumbnailSize = 80;
```
brampauwelyn, line 72: Same here, I think you can define this as  a global `const`?

brampauwelyn (approved):

src/pages/PictureUploadBox.tsx
simon-debruijn: Yep nice catch

src/pages/VideoUploadBox.tsx
simon-debruijn: Yep nice catch

PR #386, opened 2022-03-31
III-4481  - show price info on step5

brampauwelyn:
### Added
- Show priceInfo in step 5
Screenshot 2022-04-01 at 09 35 53

Screenshot 2022-04-04 at 09 30 13

### Fixed
- Hide `PriceInformation` in **Minimal**  version of additionalInformationStep

---
Ticket: https://jira.uitdatabank.be/browse/III-4481


src/pages/steps/PriceInformation.tsx
```diff
@@ -0,0 +1,97 @@
...
+const PriceInformation = ({
+  priceInfo,
+  onClickAddPriceInfo,
+  onClickAddFreePriceInfo,
+}: Props) => {
+  const { t, i18n } = useTranslation();
+
+  const hasPriceInfo = useMemo(() => {
+    return priceInfo.length > 0;
+  }, [priceInfo]);
```
simon-debruijn, line 31: I wouldn't use `useMemo` here because getting `priceInfo.length` is a really fast access with no computation required. If you want to keep the `hasPriceInfo` it can just be a normal function

simon-debruijn (approved):

PR #388, opened 2022-04-06
Feature/add new events route

brampauwelyn:
### Added
- Create edit route

### DIFF
--- a/src/i18n/de.json
+++ b/src/i18n/de.json
@@ -219,6 +219,7 @@
       "price_info": {
         "euro": "euro",
         "title": "Preise hinzufügen",
+        "change": "Ändern",
         "close": "Schließen",
         "save": "Speichern",
         "free": "Kostenlos",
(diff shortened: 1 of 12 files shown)


src/pages/events/[eventId]/edit/index.page.tsx
```diff
@@ -0,0 +1,7 @@
+import { getApplicationServerSideProps } from '@/utils/getApplicationServerSideProps';
+
+import { EventForm } from '../../../create/EventForm';
```
Comment (brampauwelyn) [inline src/pages/events/[eventId]/edit/index.page.tsx:3]: Not sure about the approach here.
Do use the existing EventForm from `create` or do we create a new `EventForm` inside the `edit` folder? 
I feel like they are going to use a lot of the same functionalities?

Comment (simon-debruijn) [reply src/pages/events/[eventId]/edit/index.page.tsx:3]: I think this way is good

simon-debruijn (approved):

PR #393, opened 2022-04-14
Type authenticated query correctly

simon-debruijn:
### Changed

- Type authenticated query correctly

---

Ticket: https://jira.uitdatabank.be/browse/III-5778


src/hooks/api/authenticated-query-v2.ts
```diff
@@ -0,0 +1,162 @@
...
+    queryKey,
+    queryArguments,
+  });
+
+  const queryFunctionWithHeaders = async (
+    context: QueryFunctionContext<GeneratedQueryKey>,
+  ) => {
+    return await queryFn(
+      context,
+      // @ts-expect-error
```
brampauwelyn, line 87: Can we remove this `//@ts-expect-error` ?

simon-debruijn: I think we won't be able to remove this one, because react-query doesn't expect another argument it only known `context` which contains `queryKey` and `pageParam`

simon-debruijn:
This was addressed in https://github.com/cultuurnet/udb3-frontend/pull/990 with a different approach

PR #397, opened 2022-06-02
III-4480 - organizer create modal

simon-debruijn:
### Added

- Organizer create modal
- edit flow organizer
- useAutoFocus hook

### Changed

- several other things

---

Ticket: https://jira.uitdatabank.be/browse/III-4480


simon-debruijn:
I'm going to make sure `fr.json` and `de.json` are synced with nl.json now. But the rest of the code can be reviewed 😃

src/hooks/api/cities.ts
```diff
@@ -0,0 +1,69 @@
...
+
+type City = {
+  label: string;
+  name: string;
+  zip: string;
+};
+
+type GetCitiesArguments = {
+  q: string;
+  country: 'BE' | 'NL';
```
brampauwelyn, line 15: Make a seperate type for country or is it overkill?

src/hooks/api/organizers.ts
```diff
@@ -20,18 +20,48 @@ type HeadersAndQueryData = {
...
 
-type GetOrganizersArguments = {
+type GetOrganizersArgumentsByQuery = {
   headers: Headers;
   embed: string;
   q: string;
 };
 
-const getOrganizers = async ({ headers, q, embed }: GetOrganizersArguments) => {
+const useGetOrganizersByQueryQuery = (
```
brampauwelyn, line 29: confusing with the double Query in the name 😛

src/hooks/api/cities.ts
simon-debruijn: I've added a constant and type now ✅

src/pages/OrganizerAddModal.tsx
```diff
@@ -0,0 +1,283 @@
...
+
+  useEffect(() => {
+    setValue('name', prefillName);
+  }, [prefillName, setValue]);
+
+  const validateUniquenessUrl = async () => {
+    if (isUrlUnique) return true;
+    setError('url', { type: 'not_unique' });
+    return false;
+  };
```
brampauwelyn, line 130: Since you're not using the return of this method.
You can simplify it I think? Also why is this function `async` ?
```suggestion
  const validateUniquenessUrl = (): void => {
    if (!isUrlUnique)  setError('url', { type: 'not_unique' });
  };
```

src/hooks/api/cities.ts
```diff
@@ -0,0 +1,69 @@
...
+      zip: city.zip,
+    }));
+
+const getCitiesNl = (): City[] => citiesNL.cities;
+
+const matchesQuery = (query: string) => {
+  return (city: City) => city.label.toLowerCase().includes(query);
+};
+
+const byLevenshtein = (query: string) => {
```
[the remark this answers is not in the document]
simon-debruijn: I named it like that because it reads nicely when used here:

`.sort(byLevenshtein(query))`

But I'm totally fine with sortByLevenshtein too 🙂

src/pages/OrganizerAddModal.tsx
```diff
@@ -0,0 +1,283 @@
...
+
+  const handleConfirm = async () => {
+    const isUniqueUrl = await validateUniquenessUrl();
+    if (!isUniqueUrl) return;
```
brampauwelyn, line 134: I'm a bit lost with the `isUniqueUrl`.
But can probably a bit simpler? 
In this case the above suggestion can also be ignored.

```suggestion
    if (!isUrlUnique) {
    setError('url', { type: 'not_unique' });
    return
    };
```

src/pages/steps/OrganizerPicker.tsx
```diff
@@ -1,75 +1,124 @@
...
+              onClick={() => setAddButtonHasBeenPressed(true)}
+            >
+              {t('create.additionalInformation.organizer.add_new_button')}
+            </Button>
+          ) : (
+            <Typeahead<Organizer>
+              options={organizers}
+              labelKey={getLabelKey(
+                i18n.language as Values<typeof SupportedLanguages>,
+              )}
```
brampauwelyn, line 100: You could use an arrow function here I guess
```suggestion
              labelKey={(org: Organizer) => getOrganizerName(org, language)}
```

brampauwelyn (approved):

simon-debruijn:
I've done the syncing of translations in a different PR:

https://github.com/cultuurnet/udb3-frontend/pull/400

src/pages/steps/AdditionalInformationStep.tsx
```diff
@@ -151,14 +164,31 @@ const AdditionalInformationStep = ({
     },
   });
 
+  const createOrganizerMutation = useCreateOrganizerMutation();
+
+  const addOrganizerToEventMutation = useAddOrganizerToEventMutation({
+    onSuccess: async () => {
+      await invalidateEventQuery('organizer');
```
geoffreydhuyvetters, line 171: Do we need async / await here?

geoffreydhuyvetters (approved):

src/pages/steps/AdditionalInformationStep.tsx
simon-debruijn: I could write it as:

`onSuccess: () => invalidateEventQuery('organizer')`

I read that it is better to await promises explicitly. Later on the stacktrace of an error will be clearer, if it fails in that async function.

https://github.com/goldbergyoni/nodebestpractices#-212-always-await-promises-before-returning-to-avoid-a-partial-stacktrace

geoffreydhuyvetters: I see. Interesting.

PR #405, opened 2022-07-01
III-4475 -  Tabs follow up, extract query logic and mutations from AdditionnalStepInformation component

brampauwelyn:
### Changed
- Tabs follow up, extract step components from AdditionnalStepInformation to seperate  components

---
Ticket: https://jira.uitdatabank.be/browse/III-4475


src/pages/steps/OrganizerStep.tsx
```diff
@@ -0,0 +1,112 @@
...
+    };
+    const { organizerId } = await createOrganizerMutation.mutateAsync(payload);
+
+    await addOrganizerToEventMutation.mutateAsync({ eventId, organizerId });
+
+    setIsOrganizerAddModalVisible(false);
+  };
+
+  return (
+    <>
```
simon-debruijn, line 87: Best use a <Stack> here instead of a fragment. If we decide to add spacing to the parent component later this won't be able to assign margin to the comonent.

Also pass the props to the Stack component

`<Stack {...getStackProps(props)}>`

brampauwelyn: Done ✅

simon-debruijn (approved):

PR #407, opened 2022-07-25
III-4488 - Theme and Type step

brampauwelyn:
### Added
- Theme and Type step

---
Ticket: https://jira.uitdatabank.be/browse/III-4488


src/pages/create/EventForm.tsx
```diff
@@ -27,13 +37,14 @@ const EventForm = () => {
       }}
       configuration={[
         // @ts-expect-error
```
brampauwelyn, line 39: Can we get rid of the `ts-expect-error` ?

src/pages/steps/ScopeStep.tsx
```diff
@@ -10,7 +10,7 @@ import { ToggleBox } from '@/ui/ToggleBox';
 
 import { FormDataIntersection, StepProps } from './Steps';
 
-const TypeStep = <TFormData extends FormDataIntersection>({
+const ScopeStep = <TFormData extends FormDataIntersection>({
```
brampauwelyn, line 13: I'm still struggling with the name `scope` 😅  but other than that it looks good 👌

src/hooks/api/types.ts
```diff
@@ -0,0 +1,30 @@
...
+    enabled: !!scope,
+    ...configuration,
+  });
+
+  const data = useMemo(() => {
+    if (!getTermsQuery.data?.terms) return;
+    const terms = getTermsQuery.data.terms.filter(
+      (term) => term.scope.includes(scope) && term.domain === 'eventtype',
+    );
+    return terms;
```
brampauwelyn, line 24: We can probably just return the filter here?
What do  you think?

```suggestion
    return getTermsQuery.data.terms.filter(
      (term) => term.scope.includes(scope) && term.domain === 'eventtype',
    );
```

src/pages/AvailabilityPageSingle.js
```diff
@@ -136,7 +136,7 @@ const AvailabilityPageSingle = ({ offer, error, useChangeStatusMutation }) => {
             </Title>,
             <StatusForm
               key="reason-and-type"
-              offerType={offerType}
+              offerType={`${offerType}s`}
```
brampauwelyn, line 139: This `s` can probably be removed?
```suggestion
              offerType={`${offerType}`}
```

simon-debruijn: No this was needed, because in this component, the offerType that is picked from the api response is singular 😵‍💫

src/hooks/api/types.ts
simon-debruijn: https://github.com/cultuurnet/udb3-frontend/pull/407/commits/fa2f27657b6b57bfba007e795d2cfe3cf7dd101c

simon-debruijn (approved):

PR #409, opened 2022-08-02
DDT-39 - Add less strict checks @context

simon-debruijn:
### Changed

-  less strict checks @context

---

Ticket: https://jira.uitdatabank.be/browse/DDT-39

### DIFF
--- a/src/types/Event.ts
+++ b/src/types/Event.ts
@@ -16,7 +16,8 @@ type Event = Offer & {
 };
 
 const isEvent = (value: unknown): value is Event => {
-  return value['@context'] === '/contexts/event';
+  if (typeof value?.['@context'] !== 'string') return false;
+  return value['@context'].endsWith('/contexts/event');
 };
 
 const areEvents = (value: unknown[]): value is Event[] => {
--- a/src/types/Event.ts
+++ b/src/types/Event.ts
@@ -17,7 +17,7 @@ type Event = Offer & {
 
 const isEvent = (value: unknown): value is Event => {
   if (typeof value?.['@context'] !== 'string') return false;
-  return value['@context'].endsWith('/contexts/event');
+  return value['@context'].endsWith('/event');
 };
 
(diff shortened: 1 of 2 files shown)


brampauwelyn (approved):

src/types/Event.ts
```diff
@@ -16,7 +16,8 @@ type Event = Offer & {
 };
 
 const isEvent = (value: unknown): value is Event => {
-  return value['@context'] === '/contexts/event';
+  if (typeof value?.['@context'] !== 'string') return false;
+  return value['@context'].endsWith('/contexts/event');
```
brampauwelyn, line 20: You could even make it less specific by just checking the end
```suggestion
  return value['@context'].endsWith('event');
```

Not sure what @bertramakers  thinks

simon-debruijn: Yes sure, I agree :)

bertramakers: Less reliance on the `@context` URL structure is always better I think. And I don't imagine we'd ever have another context that would end with "event", that would just get confusing. So yeah `.endsWith('event')` is slightly better.

Thinking one step further for a moment: Another way to differentiate events and places is based on the `location` and `address` properties. Events always have a `location` (even online events), and places always have an `address` property. That's maybe even safer in general, because we will definitely never change that (in UDB v3) for backward compatibility reasons.

The value of `@context` is just supposed to be a URI that can be resolved, so in theory it could also be `https://io.uitdatabank.be/whwieghwuehgw`. And the context could in theory even be the same for both events and places, as a JSON-LD context is not necessarily specific to one resource type. So to differentiate on that is possible right now (and understandable without another clear "type" property), but differentiating on something that is inherent to events or places like the `location` vs `address` properties is even better IMHO.

That being said, we'll probably take care to not change the `@context` value too much aside from it becoming an absolute URL because it's been like this for so long and I can imagine other integrators also checking it.

Just wanted to share it as a heads up if you ever integrate with other JSON-LD APIs.

PR #418, opened 2022-09-14
III-4909 - Use new Event and Location flaticons in ScopeStep

brampauwelyn:
### Changed
- icon prop in `ToggleBox` to JSX.element for more flexibility
- use new Event and Location flaticons in ScopeStep
- check icon on active `ToggleBox` state

Screenshot 2022-09-14 at 11-39-31 UiTdatabank

---
Ticket: https://jira.uitdatabank.be/browse/III-4909


src/ui/ToggleBox.tsx
```diff
@@ -27,39 +25,33 @@ const ToggleBox = ({
     <Stack
       forwardedAs="button"
       onClick={onClick}
-      padding={5}
+      padding={5.5}
```
simon-debruijn, line 28: I would keep using non-decimal numbers. Since we have only used normal ints as padding, margin and spacing. But if its really an obvious improvement like this, keep it 😉

simon-debruijn (approved):

PR #421, opened 2022-09-14
III-4924 - Only show errors when fields are touched

brampauwelyn:
### Fixed
- PriceInformation step: only show errors when fields are touched

---
Ticket: https://jira.uitdatabank.be/browse/III-4924


src/pages/steps/PriceInformation.tsx
```diff
@@ -154,7 +154,7 @@ const PriceInformation = ({
     watch,
     setValue,
     trigger,
-    formState: { errors },
+    formState: { errors, touchedFields },
```
simon-debruijn, line 157: I'm still not sure about this. But does it work with `dirtyFields` since we haven't had to use touchedFields yet

brampauwelyn: Jup it also works with `dirtyFIelds` 👍

simon-debruijn (approved):

PR #433, opened 2022-10-07
III-5004 - New BookingInfoStep

brampauwelyn:
### Added
-new  BookingInfoStep

Screenshot 2022-10-10 at 13 22 03

---
Ticket: https://jira.uitdatabank.be/browse/III-5004

### DIFF
--- a/src/pages/steps/BookingInfoStep.tsx
+++ b/src/pages/steps/BookingInfoStep.tsx
@@ -1,5 +1,5 @@
 import { yupResolver } from '@hookform/resolvers/yup';
-import { ChangeEvent, useEffect, useState } from 'react';
+import { ChangeEvent, useEffect, useRef, useState } from 'react';
 import { useForm } from 'react-hook-form';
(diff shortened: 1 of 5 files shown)


src/pages/steps/BookingInfoStep.tsx
```diff
@@ -0,0 +1,494 @@
...
+
+type BookingInfo = {
+  email?: string;
+  phone?: string;
+  url?: string;
+  urlLabel?: {
+    de: string;
+    en: string;
+    fr: string;
+    nl: string;
```
simon-debruijn, line 81:
```suggestion
    de?: string;
    en?: string;
    fr?: string;
    nl?: string;
```

Are these always there, so empty strings if not set? Or can they be undefined

brampauwelyn: They should always be there with a "predefined" value.
https://documentatie.uitdatabank.be/content/json-ld-crud-api/latest/events/event-bookingInfo.html

PR #444, opened 2022-10-21
III-4923 - Don't make alerts full-width

brampauwelyn:
### Changed
- Don't make alerts full-width

**Screenshots**
Screenshot 2022-10-26 at 10 24 30

Screenshot 2022-10-26 at 10 08 22

---
Ticket: https://jira.uitdatabank.be/browse/III-4923

### DIFF
--- a/src/pages/steps/PriceInformation.tsx
+++ b/src/pages/steps/PriceInformation.tsx
@@ -391,7 +391,7 @@ const PriceInformation = ({
         </Inline>
       ))}
       {hasGlobalError && (
-        <Alert marginTop={3} variant={AlertVariants.INFO}>
+        <Alert display="flex" marginTop={3} variant={AlertVariants.INFO}>
           <Box
             forwardedAs="div"
             dangerouslySetInnerHTML={{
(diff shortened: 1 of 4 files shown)


src/pages/steps/PriceInformation.tsx
```diff
@@ -391,7 +391,7 @@ const PriceInformation = ({
         </Inline>
       ))}
       {hasGlobalError && (
-        <Alert marginTop={3} variant={AlertVariants.INFO}>
+        <Alert display="flex" marginTop={3} variant={AlertVariants.INFO}>
```
simon-debruijn, line 394:
```suggestion
        <Alert marginTop={3} variant={AlertVariants.INFO}>
```

It could be an idea to just add `display="flex"` in the Alert component itself on the Box component

simon-debruijn: Or use the `Inline` components instead of Box, which is 

```
display: flex;
flex-direction: row;
```

simon-debruijn (approved):

src/pages/steps/PriceInformation.tsx
brampauwelyn: Used an inline component ✅ 
Updated screenshots ☝️

PR #448, opened 2022-10-21
III-5022 - Success icon improvements

brampauwelyn:
### Changed
- hex value of color green3
- use global value for successIcon instead of too specific theming
- increase default size of icons `15px` -> `18px`

## Screenshots
Screenshot 2022-10-21 at 11 50 50

Screenshot 2022-10-21 at 11 50 45
8727-a874-4541-82a0-703a197e599f.png">

---
Ticket: https://jira.uitdatabank.be/browse/III-5022

### DIFF
--- a/src/ui/Icon.tsx
+++ b/src/ui/Icon.tsx
@@ -132,8 +132,8 @@ const Icon = ({ name, width, height, className, ...props }: Props) => {
 };
 
 Icon.defaultProps = {
-  width: 15,
-  height: 15,
+  width: 18,
+  height: 18,
 };
 
 export { Icon, Icons };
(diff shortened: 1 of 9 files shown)


src/ui/Icon.tsx
```diff
@@ -132,8 +132,8 @@ const Icon = ({ name, width, height, className, ...props }: Props) => {
 };
 
 Icon.defaultProps = {
-  width: 15,
-  height: 15,
+  width: 18,
+  height: 18,
 };
```
simon-debruijn, line 137: This does affect all icons used, have you checked other pages?

brampauwelyn: Yes it does affect all the other icons, since it's only 3px difference, it shouldn't be an issue.
But I can do an extra check

brampauwelyn: @simon-debruijn  I did an extra check

Icons in navigation are bigger 
Icons in buttons are bigger as well... 

Screenshot 2022-10-26 at 10 37 21

Screenshot 2022-10-26 at 10 37 47

simon-debruijn (approved):

PR #457, opened 2022-11-03
III-4472 - Set up CalendarStep + add basic hourpicker

brampauwelyn:
### Added
- CalendarStep boilerplate
- basic hour picker for calendar step

---

Ticket: https://jira.uitdatabank.be/browse/III-4472

### DIFF
--- a/src/pages/steps/CalendarStep.tsx
+++ b/src/pages/steps/CalendarStep.tsx
@@ -41,7 +41,7 @@ const CalendarStep = ({ ...props }: CalendarStepProps) => {
             onBlur={() => {
               console.log('on blur');
             }}
-            selected=""
+            selected={['']}
             minLength={0}
             inputType="time"
             customFilter={() => true}
--- a/src/pages/steps/CalendarStep.tsx
+++ b/src/pages/steps/CalendarStep.tsx
@@ -9,22 +9,22 @@ import { FormDataUnion, StepsConfiguration } from './Steps';
 
 type CalendarStepProps = StackProps;
 
+const getHourOptions = () => {
+  const hours = Array(24).fill(0);
+  const quarterHours = ['00', '15', '30', '45'];
+  const times = [];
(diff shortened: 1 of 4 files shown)


src/pages/steps/CalendarStep.tsx
```diff
@@ -0,0 +1,63 @@
...
+    const hours = Array(24).fill(0);
+    const quarterHours = ['00', '15', '30', '45'];
+    const times = [];
+    hours.forEach((_hour, i) => {
+      quarterHours.forEach((quarterHour) =>
+        times.push(`${i > 9 ? i : `0${i}`}:${quarterHour}`),
+      );
+    });
+    return times;
+  };
```
simon-debruijn, line 25: This function can move outside of the component

src/pages/steps/CalendarStep.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+  const hourOptions = getHourOptions();
```
simon-debruijn, line 27:
```suggestion
  const hourOptions = useMemo(getHourOptions, []);
  ```

File: src/ui/Typeahead.tsx [hunk 3]
```diff
@@ -128,7 +133,9 @@ const Typeahead: TypeaheadFunc = forwardRef(
         selected={selected}
         inputProps={{
           id,
+          type: inputType ?? 'text',
```
simon-debruijn, line 136:
```suggestion
          type: inputType,
```

This can go since you use defaultProps

src/pages/steps/CalendarStep.tsx
brampauwelyn: Why would you use "useMemo"? I think it's not necessary here? 
I can just move this const outside the component function?

simon-debruijn: Yes that's true 😄

simon-debruijn (approved):

PR #471, opened 2022-11-30
Improvements to fix unwanted logout on failing card systems get

simon-debruijn:
### Changed

- [Move logic 'Failed to fetch' only to user calls](https://github.com/cultuurnet/udb3-frontend/commit/02c945cc52bd3ab575476ded3425d4e6630aef28)
- [Set max of 2 retries for uitpas cardsystem calls](https://github.com/cultuurnet/udb3-frontend/commit/66263130b30f24352499d76fb54ca09a32dd1550)

### Fixed

- Fix unwanted logout on failing card systems get

### DIFF
--- a/src/hooks/api/user.ts
+++ b/src/hooks/api/user.ts
@@ -1,41 +1,23 @@
-import {
-  ErrorObject,
-  FetchError,
-  fetchFromApi,
-  isErrorObject,
-} from '@/utils/fetchFromApi';
+import { User } from '@/types/User';
+import { fetchFromApi, isErrorObject } from '@/utils/fetchFromApi';
 
 import {
   ServerSideQueryOptions,
   useAuthenticatedQuery,
 } from './authenticated-query';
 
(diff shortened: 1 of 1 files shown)


src/hooks/api/uitpas.ts
```diff
@@ -42,6 +38,7 @@ const useGetCardSystemForEventQuery = (
     queryFn: getCardSystemForEvent,
     queryArguments: { eventId },
     enabled: !!eventId,
+    retry: 2,
```
simon-debruijn, line 41: This could even be changed to only 1 retry

simon-debruijn:
After a conversation in `mwl-publiq-backend` we came to the conclusion that the frontend doesn't need custom logic to handle these `Failed to fetch` errors. The backend will be changed to respond the error status back to the frontend without missing CORS headers.

https://cultuurnet.slack.com/archives/CM1814JMU/p1669822347047879

brampauwelyn (approved):

PR #486, opened 2022-12-07
Add ci script

simon-debruijn:
### Added

- Add ci script


simon-debruijn:
The & operator is to run npm scripts concurrently. Instead of && that runs them sequentially

brampauwelyn (approved):

PR #536, opened 2023-01-19
III-5238 - Show broken vimeo video thumbnails

brampauwelyn:
### Changed
- Use https to get vimeo thumbnails 
- Show successful video thumbnails instead of 0 if an api calls fails

### Fixed
- `Blocked loading mixed active content "http://vimeo.com/api/v2/video/117686240.json"`

---
Ticket: https://jira.uitdatabank.be/browse/III-5238

### DIFF
--- a/src/pages/steps/AdditionalInformationStep/MediaStep.tsx
+++ b/src/pages/steps/AdditionalInformationStep/MediaStep.tsx
@@ -18,6 +18,7 @@ import { Inline } from '@/ui/Inline';
 import { getStackProps, Stack } from '@/ui/Stack';
 import { Breakpoints } from '@/ui/theme';
 import { parseOfferId } from '@/utils/parseOfferId';
+import { isFulfilledResult } from '@/utils/promise';
 
 import type { ImageType } from '../../PictureUploadBox';
 import { PictureUploadBox } from '../../PictureUploadBox';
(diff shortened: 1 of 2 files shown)


src/pages/steps/AdditionalInformationStep/MediaStep.tsx
```diff
@@ -193,9 +193,13 @@ const MediaStep = ({
       return enrichedVideo;
     });
 
-    const data = await Promise.all(convertAllVideoUrlsPromises);
+    const data = await Promise.allSettled(convertAllVideoUrlsPromises);
 
-    setVideos(data);
+    const successVideos = data
+      .filter((res) => res.status === 'fulfilled')
+      .map((res: PromiseFulfilledResult<VideoEnriched>) => res.value);
```
simon-debruijn, line 200:
```suggestion
  const isFulfilledResult = <T>(res: PromiseSettledResult<T>): res is PromiseFulfilledResult<T> => {
    return res.status === 'fulfilled'
  }

  const successVideos = data.filter(isFulfilledResult).map(res => res.value);
```

You can use a typescript narrowing function like here so you don't need the type annotation. I would put the `isFulfilledResult` in a different file so we can reuse it for other cases.

I also implemented this on UIV https://github.com/cultuurnet/uitbe-frontend/blob/main/packages/app/utilities/Promise.ts 😄

simon-debruijn (approved):

PR #543, opened 2023-01-20
III-5259 - Show maxFileSize in error message

brampauwelyn:
### Fixed
- Show maxFileSize of 20MB instead of 5MB in error message

Screenshot 2023-01-20 at 09 33 03

### Added
- Spinner when uploading image
Screenshot 2023-01-20 at 14 26 01

---
Ticket: https://jira.uitdatabank.be/browse/III-5259


simon-debruijn (approved):

src/pages/steps/modals/PictureUploadModal.tsx
```diff
@@ -243,6 +243,15 @@ const PictureUploadModal = ({
...
   const registerDescriptionProps = register('description');
 
+  const fileError =
+    errors.file?.type === 'size'
+      ? t(`pictures.upload_modal.validation_messages.file.size`, {
+          maxFileSize: MAX_FILE_SIZE / 1_000_000,
+        })
+      : errors.file?.type
+      ? t(`pictures.upload_modal.validation_messages.file.${errors.file?.type}`)
+      : undefined;
```
Anahkiasen, line 253: Not sure the nested ternary is necessary here since you could pass the placeholders to the translation function and it'd ignore it for the other usecases no? Then you can always use the template string version

brampauwelyn: Yes good suggestion, never thought of it this way 👍

simon-debruijn (approved):

PR #546, opened 2023-01-25
III-5274 - Fix step 3 and 4 disappearing

simon-debruijn:
### Changed

- [Improve typing StepsConfiguration](https://github.com/cultuurnet/udb3-frontend/pull/546/commits/24cd8a7ac8ac29c3ba6780e03bb3be68a9b8bd13)
- [Also update form value calendar on initial load](https://github.com/cultuurnet/udb3-frontend/pull/546/commits/951295cbaaa0f9e9926b83489d3b76d23eb60f97)

### Fixed

- Fix step 3 and 4 disappearing
- [Fix calendar hiding by changing the useFormState so it becomes touched](https://github.com/cultuurnet/udb3-frontend/pull/546/commits/4c7ee778b891ed4b59b6458ebf2dca1aef150cf6)

---

Ticket: https://jira.uitdatabank.be/browse/III-5274

### DIFF
--- a/src/pages/steps/LocationStep.tsx
+++ b/src/pages/steps/LocationStep.tsx
@@ -146,6 +146,7 @@ const LocationStep = ({
                         isOnline: e.target.checked,
                       };
                       field.onChange(updatedValue);
+                      field.onBlur();
                       onChange(updatedValue);
                     }}
                     css={`
--- a/src/pages/steps/LocationStep.tsx
+++ b/src/pages/steps/LocationStep.tsx
@@ -242,6 +242,7 @@ const LocationStep = ({
                       };
                       field.onChange(updatedValue);
                       onChange(updatedValue);
+                      field.onBlur();
                     }}
                   >
                     {t('create.location.country.change_location')}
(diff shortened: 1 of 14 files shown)


src/pages/steps/LocationStep.tsx
```diff
@@ -284,6 +286,7 @@ const LocationStep = ({
                       };
                       field.onChange(updatedValue);
                       onChange(updatedValue);
+                      field.onBlur();
```
brampauwelyn, line 289: why are these field.onBlurs necessary?

brampauwelyn (approved):

src/pages/steps/LocationStep.tsx
simon-debruijn: In `Steps` we check if the field has a value true on is `isTouched`.

Description from docs:

**isTouched**: field has received a focus and blur event.

https://react-hook-form.com/api/useform/getfieldstate#main

It can probably also be done with dirty fields and checking `isDirty`. But since it was working this way already I kept that logic.

brampauwelyn: 👍

PR #551, opened 2023-01-25
III-5249 - Fix online participation url is not added to an event

brampauwelyn:
### Changed
- [Make util to prefix Url with http://](https://github.com/cultuurnet/udb3-frontend/pull/551/commits/7913dd35e9ff16900d20b73046da6f1908c08cb9)

### Fixed
- [Submit onlineUrl](https://github.com/cultuurnet/udb3-frontend/pull/551/commits/9b4bc5aa50c5e3dc18b038645fbcd18f21f48255)
- [Show onlineUrl in location step](https://github.com/cultuurnet/udb3-frontend/pull/551/commits/3adbb74f6b15ead173a9cb78c2cc4f2fa8791cd6)
- [Delete onlineUrl when switching to OFFLINE event](https://github.com/cultuurnet/udb3-frontend/pull/551/commits/0bcab0e5fa7ed84d885f8510470847031abcef11)
- laggy onlineUrl field, every change in the onlineUrl input field would trigger an api request...  Now only submit on blur

---
Ticket: https://jira.uitdatabank.be/browse/III-5249


src/utils/url.ts
```diff
@@ -0,0 +1,7 @@
+const prefixUrlWithHttp = (url: string) => {
+  if (!url.startsWith('http')) return url;
```
simon-debruijn, line 2:
```suggestion
  if (url.startsWith('http')) return url;
```

Shouldn't it be 
if url.startsWith('http') -> return url;
else -> return `http://${url}`;

brampauwelyn: Ooh yes 🤦

simon-debruijn (approved):

PR #554, opened 2023-01-25
Add warning icon on Price tab when empty with UiTPas organizer

Anahkiasen:
### Added

- Add warning icon on Price tab when empty with UiTPas organizer

### Changed

- Moved Organiser tab next to Price tab

---

Ticket: https://jira.uitdatabank.be/browse/III-5276

### DIFF
--- a/src/pages/steps/AdditionalInformationStep/AdditionalInformationStep.tsx
+++ b/src/pages/steps/AdditionalInformationStep/AdditionalInformationStep.tsx
@@ -50,7 +50,7 @@ type TabContentProps = {
   onSuccessfulChange: (
     data?: any,
   ) => typeof data extends any ? void : Promise<void>;
-  onValidationChange?: (status: boolean | ValidationStatus) => void;
+  onValidationChange?: (status: ValidationStatus) => void;
 };
 
 type TabConfig = {
--- a/src/pages/steps/AdditionalInformationStep/AdditionalInformationStep.tsx
+++ b/src/pages/steps/AdditionalInformationStep/AdditionalInformationStep.tsx
@@ -22,6 +22,7 @@ import { MediaStep } from './MediaStep';
 import { OfferScore } from './OfferScore';
 import { OrganizerStep } from './OrganizerStep';
 import { PriceInformation } from './PriceInformation';
+import { map, mapValues } from 'lodash';
 
 const getGlobalValue = getValueFromTheme('global');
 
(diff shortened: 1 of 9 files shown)


brampauwelyn (approved):
Looks good to me 👍

src/pages/steps/AdditionalInformationStep/AdditionalInformationStep.tsx
```diff
@@ -136,6 +139,12 @@ type Props = StackProps & {
   variant?: Values<typeof AdditionalInformationStepVariant>;
 };
 
+export enum ValidationStatus {
+  NONE = 'none',
+  WARNING = 'warning',
+  SUCCESS = 'success',
+}
```
simon-debruijn, line 146:
```suggestion
import { Values } from '@/types/Values';

export const ValidationStatus = {
  NONE: 'none',
  WARNING: 'warning',
  SUCCESS: 'success',
} as const;

export type ValidationStatus = Values<typeof ValidationStatus>;
```
We prefer using javascript objects and infering the type from it like this. See reasons why enums are not the best construct in typescript: https://www.youtube.com/watch?v=jjMbPt_H3RQ

And most of the time we prefer exporting the variables at the end of the file, so we can get a quick overview of what's exported.

src/pages/steps/AdditionalInformationStep/AdditionalInformationStep.tsx
```diff
@@ -1,3 +1,4 @@
+import { map, mapValues } from 'lodash';
```
simon-debruijn, line 1:
```suggestion
import { mapValues } from 'lodash';
```

I don't think this import is used

src/pages/steps/AdditionalInformationStep/AdditionalInformationStep.tsx
```diff
@@ -216,20 +225,26 @@ const AdditionalInformationStep = ({
...
                   offerId={offerId}
                   scope={scope}
-                  onChangeCompleted={(isCompleted) => {
-                    if (completedFields[field] === isCompleted) return;
+                  onValidationChange={(status) => {
+                    if (typeof status === 'boolean') {
+                      status = status
+                        ? ValidationStatus.SUCCESS
+                        : ValidationStatus.NONE;
+                    }
```
simon-debruijn, line 241: I would only accept the `ValidationStatus` for `onValidationChange` then you can pass `ValidationStatus.SUCCESS` instead of `true`and `ValidationStatus.NONE` instead of `false`

Anahkiasen: That's what I started to do but it was a heavier change so since this is my first semi-big PR I was wary of changing everything :D

simon-debruijn (approved):

PR #558, opened 2023-01-27
Fix setting price to Free not updating in the API

Anahkiasen:
### Fixed

- Fix setting price to Free not updating in the API

---

Ticket: https://jira.uitdatabank.be/browse/III-5283

The button would use the current rates from the `useWatch` which has a [delayed update cycle](https://react-hook-form.com/api/usewatch/#rules) to ensure best performance. This meant the value sent to the API would be the one before the update to free done by the `setValue`. 
This changes it to use the current value from `getValue` to ensure it reflects the current state.

Also removed a `console.log` that was likely left over


brampauwelyn (approved):
Works like a charm 👍

simon-debruijn (approved):

PR #566, opened 2023-02-02
Refactor PriceInformation form to optimistic updates

Anahkiasen:
### Changed

- Refactor PriceInformation form to optimistic updates

---

Ticket: https://jira.uitdatabank.be/browse/III-5256

It wasn't originally my intention to refactor the form entirely, but due to the multiple tickets revolving around issues with out of sync states and such, and the suggestion to move to optimistic updates in the ticket, I felt it was perhaps the way to go so I went all in.

The diff is a bit scary but the crux of the changes are the following:
1. I moved a bunch of things up and down to group related things more closely together. 
2. I moved the "unique name" validation to Yup to be able to use the same logic for all errors and validation.
3. I removed most local states in order to use the form's data as state directly (which was the crux of most out of sync issues)
4. I made the whole form submit on blur (with a throttle to avoid conflicting requests) which removed the need to manually watch for changes and submit. I still manually submit when removing/appending rows since for some reason that doesn't trigger `react-hook-form`'s `watch` but that could be a mistake on my part somewhere
5. I didn't put back the automatic price formatting as I wasn't sure this was intentional before? and it felt a bit weird UX wise to have the fields change under your hands but I can likely add it back if need be.

Here is a small video of the form acting (in theory) correctly:

https://user-images.githubusercontent.com/1321596/216367348-328488a0-78f8-484b-b8a4-32c694324fd6.mp4

Considering this is a refactor I'd be happy to have some more hands testing this to be sure there aren't edge cases not working

1ea2bdef-35ee-4700-8454-dc3b1ed439b3_text

### DIFF
--- a/src/pages/steps/AdditionalInformationStep/PriceInformation.tsx
+++ b/src/pages/steps/AdditionalInformationStep/PriceInformation.tsx
@@ -1,8 +1,9 @@
 import { yupResolver } from '@hookform/resolvers/yup';
-import { useEffect, useMemo, useRef, useState } from 'react';
-import { useForm, useWatch } from 'react-hook-form';
+import { useCallback, useEffect, useMemo, useRef } from 'react';
+import { useFieldArray, useForm } from 'react-hook-form';
 import { useTranslation } from 'react-i18next';
 import * as yup from 'yup';
+import { ValidationError } from 'yup';
 
 import { OfferTypes } from '@/constants/OfferType';
 import { useGetEventByIdQuery } from '@/hooks/api/events';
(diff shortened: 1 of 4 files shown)


src/pages/steps/AdditionalInformationStep/PriceInformation.tsx
```diff
@@ -81,26 +83,49 @@ const shouldHaveAName = (value: any): boolean => {
...
+            })
+            .test(`name-is-required`, 'name is required', shouldHaveAName)
+            .test(`name-is-not-uitpas`, 'should not be uitpas', isNotUitpas)
+            .required(),
+          category: yup.string(),
+          price: yup.string().matches(PRICE_REGEX).required(),
+          priceCurrency: yup.string(),
+        }),
+      )
+      .uniqueName(),
```
simon-debruijn, line 128:
```suggestion
      .test("uniqueName", yourValidationFunction);
```

I prefer just chaining a test function to it instead of adding a method to `yup.array`

Anahkiasen: Ah nice that's better indeed, plus better addition/deletion ratio :p

simon-debruijn (approved):

PR #583, opened 2023-02-08
Simplify time filtering logic

Anahkiasen:
### Changed

- Simplify time filtering logic

---

Ticket: https://jira.uitdatabank.be/browse/III-5247

The times are now always filtered based on the starting time since the ending time will always be dependant on it. This prevents the issue where dropdowns gradually had less and less items and the issue where the default ending time of 23:59 would mean nothing in the dropdown.

### DIFF
--- a/src/ui/TimeSpanPicker.tsx
+++ b/src/ui/TimeSpanPicker.tsx
@@ -100,7 +100,7 @@ const TimeSpanPicker = ({
           inputRequired={true}
           name="startTime"
           id={`${idPrefix}-start`}
-          customFilter={filterTimes}
+          customFilter={isQuarterHour}
           defaultInputValue={startTime}
           options={hourOptions}
           labelKey={(option) => option}
--- a/src/ui/TimeSpanPicker.tsx
+++ b/src/ui/TimeSpanPicker.tsx
@@ -84,7 +84,8 @@ const TimeSpanPicker = ({
 
   const filterTimes = useCallback(
     (time: string) =>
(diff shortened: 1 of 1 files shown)


src/ui/TimeSpanPicker.tsx
```diff
@@ -95,6 +82,12 @@ const TimeSpanPicker = ({
   const { t } = useTranslation();
   const idPrefix = `${id}-time-span-picker`;
 
+  const filterTimes = useCallback(
+    (time: string) =>
+      isQuarterHour(time) && timeToNumeric(time) >= timeToNumeric(startTime),
+    [startTime],
+  );
+
```
simon-debruijn, line 90: I think the value '23:59' now won't be included in the endTime picker

Anahkiasen: I wasn't sure if that was still needed, what was the original use case for it? Since the user can still type 23:59 if need be

simon-debruijn: That is the initial value of the endTime when creating an event. It indicates till the end of the day, so it would be weird if you can't select it via the picker imo. It makes it harder to enter full day time values.

Anahkiasen: Ah yes ok it's for full day events makes sense, will add it back

brampauwelyn (changes requested):
@Anahkiasen after discussion with Corneel we decided to **always** show the full list in the startTimes.  The endTimes are good as you implemented 👍 . 
Screenshot 2023-02-08 at 14 29 43

brampauwelyn (approved):

PR #586, opened 2023-02-08
Add back missing error filtering throwing empty errors and silently failing validation

Anahkiasen:
### Fixed

- Add back missing error filtering throwing empty errors and silently failing validation

---

Ticket: https://jira.uitdatabank.be/browse/III-5317

Seems one of the merges conflict resolution resulted in this incorrect code without the `errors` filtering, resulting in validation still failing silently


brampauwelyn (approved):
Code itself looks good.
But the prices are not being sent to the API.
Same issue we had on Tuesday.

simon-debruijn (approved):

PR #590, opened 2023-02-09
Iii 5297 make clear price and name are required

simon-debruijn:
### Added

- [Added ...]

### Changed

- [Changed ...]

### Removed

- [Removed ...]

### Fixed

- [Fixed ...]

---

### DIFF
--- a/src/pages/steps/AdditionalInformationStep/PriceInformation.tsx
+++ b/src/pages/steps/AdditionalInformationStep/PriceInformation.tsx
@@ -10,7 +10,7 @@ import { OfferTypes } from '@/constants/OfferType';
 import { useGetEventByIdQuery } from '@/hooks/api/events';
 import { useAddOfferPriceInfoMutation } from '@/hooks/api/offers';
 import { useGetPlaceByIdQuery } from '@/hooks/api/places';
-import i18n from '@/i18n/index';
+import i18n, { SupportedLanguage } from '@/i18n/index';
 import {
   TabContentProps,
   ValidationStatus,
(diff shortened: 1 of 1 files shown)


brampauwelyn (approved):

src/pages/steps/AdditionalInformationStep/PriceInformation.tsx
```diff
@@ -153,95 +142,75 @@ const PriceInformation = ({
...
+    control,
+  });
 
-  const ratesField = useFieldArray({ name: 'rates', control });
-  const rates = watch('rates');
-  const controlledRates = ratesField.fields.map((field, index) => ({
-    ...field,
-    ...rates[index],
-  }));
+  const rates = fields;
```
Anahkiasen, line 165: I think `fields` is only the default value not the current value so I don't think you can infer `hasUitpasPrices` from it no?

Screenshot 2023-02-10 at 12 02 24

src/pages/steps/AdditionalInformationStep/PriceInformation.tsx
```diff
@@ -247,7 +247,7 @@ const PriceInformation = ({
             errors.rates?.at(index)?.type;
 
           return (
-            <Stack key={`rate_${rate.id}`}>
+            <Stack key={`rate_${rate.name}_${rate.price}_${rate.category}`}>
```
Anahkiasen, line 250: The `id` comes from the `fields` object originally that's why I had to build the `controlledFields` like in their example to get the benefit from it https://react-hook-form.com/api/usefieldarray/#shouldUnregister-example

simon-debruijn: Ok, good to know 👍🏻

brampauwelyn (commented):
Code looks good but the green check mark doesn't appear when you add rates for the first time. When you refresh the page it works.

brampauwelyn (approved):

PR #591, opened 2023-02-09
Fixed some issues with client and API rates reconciliation

Anahkiasen:
### Fixed

- Fixed some issues with client and API rates reconciliation

---

Ticket: https://jira.uitdatabank.be/browse/III-5323


src/utils/reconcileRates.test.ts
```diff
@@ -0,0 +1,259 @@
...
+      },
+      {
+        name: { nl: 'bar' },
+        category: PriceCategories.BASE,
+        price: 2500000,
+        priceCurrency: 'EUR',
+      },
+    ];
+
+    // @ts-ignore
```
simon-debruijn, line 58: If you must ignore a ts issue. use `@ts-expect-error` instead. typescript will show error if the `@ts-expect-error` isn't needed anymore.

simon-debruijn:
@Anahkiasen Could you give some extra info on what `reconcileRates` does exactly? From the naming of the function I don't 100% get what is should do. Is it to guarantee a specific order?

src/pages/steps/AdditionalInformationStep/PriceInformation.tsx
```diff
@@ -441,5 +417,5 @@ const PriceInformation = ({
   );
 };
 
-export type { PriceCategory };
-export { PriceInformation };
+export type { FormData, PriceCategory };
+export { PriceCategory as PriceCategories, PriceInformation };
```
simon-debruijn, line 421: You don't have to rename the constant per se. Having a contant and type with the same name actually has a nice DX to it. Since typescript let's you use the type as type and constant as constant, it feels like enums without the drawbacks 😉

Anahkiasen:
@simon-debruijn Since we basically have two sources of truth for prices – the current local data which is the truth for the normal prices, and the API which is the source of truth for UITPAS prices – this takes both and mixes them so that both are preserved while integrating changes. 

Before I did the Price refactor we used to trust the API always, but it'd cause issues when the local copy was right, like when you'd delete a price right after updating another and the deleted price would come back (since the API didn't receive the deletion yet).
When I changed it to full optimistic, it moved the balance the other way around but that made all UiTPAS related stuff not work since those come purely from the API.

I added a function so that I could add some unit tests, and also add back the feature I had removed in the first refactor (the INTL price formatting). So it's a bit "more" than was needed but it was more of a way to add some tests around the logic without going full E2E for now

src/pages/steps/AdditionalInformationStep/PriceInformation.tsx
Anahkiasen: Does seem to throw a TS error though, do I need to use a specific syntax?

Screenshot 2023-02-14 at 13 52 13

brampauwelyn:
@Anahkiasen I did some quick tests locally and everything seems to work fine 👍

brampauwelyn (approved):

simon-debruijn (approved):

PR #597, opened 2023-02-10
III-5301 - Add scroll from step 2 to step 3

brampauwelyn:
### Added
- Add scroll from step 2 (typeAndTheme) to step 3 (calendar).

---
Ticket: https://jira.uitdatabank.be/browse/III-5301

### DIFF
--- a/src/pages/steps/CalendarStep/CalendarStep.tsx
+++ b/src/pages/steps/CalendarStep/CalendarStep.tsx
@@ -126,7 +126,10 @@ const CalendarStep = ({
 }: CalendarStepProps) => {
   const { t } = useTranslation();
 
-  const scope = useWatch({ control, name: 'scope' });
+  const [scope, type] = useWatch({
+    control,
+    name: ['scope', 'typeAndTheme.type'],
+  });
 
   const calendarService = useCalendarContext();
 
@@ -253,8 +256,27 @@ const CalendarStep = ({
     handleChooseFixedDays();
   }, [scope, isIdle, handleChooseFixedDays]);
 
+  // scroll to calendar step after theme has been selected
+  useEffect(() => {
+    if (!scope || !type.id) return;
+    if (offerId) return;
+
+    const yOffset = -60;
+    const main = document.getElementsByTagName('main')[0];
+    const calendarContainer = document.getElementById(
+      'calendar-step-container',
+    );
+    const scrollToPosition =
+      calendarContainer.getBoundingClientRect().top +
+      window.pageYOffset +
+      yOffset;
+
+    main.scrollTo({ top: scrollToPosition, behavior: 'smooth' });
+  }, [scope, offerId, type]);
+
   return (
     <Stack
+      id="calendar-step-container"
       spacing={4}
       minWidth={{ l: 'auto', default: '60rem' }}
       width={{ l: '100%', default: 'min-content' }}


src/pages/steps/CalendarStep/CalendarStep.tsx
```diff
@@ -253,8 +256,27 @@ const CalendarStep = ({
...
+    const main = document.getElementsByTagName('main')[0];
+    const calendarContainer = document.getElementById(
+      'calendar-step-container',
+    );
+    const scrollToPosition =
+      calendarContainer.getBoundingClientRect().top +
+      window.pageYOffset +
+      yOffset;
+
+    main.scrollTo({ top: scrollToPosition, behavior: 'smooth' });
```
Anahkiasen, line 274: I think we can do `calendarContainer.scrollIntoView({behavior: 'smooth'})` to avoid the manual calculations

brampauwelyn: Yes I tried that, the problem is that the title is not visible in that case since the title is not inside the calendar container...

src/pages/steps/CalendarStep/CalendarStep.tsx
```diff
@@ -253,8 +256,27 @@ const CalendarStep = ({
...
+  // scroll to calendar step after theme has been selected
+  useEffect(() => {
+    if (!scope || !type.id) return;
+    if (offerId) return;
+
+    const yOffset = -60;
+    const main = document.getElementsByTagName('main')[0];
+    const calendarContainer = document.getElementById(
+      'calendar-step-container',
+    );
```
simon-debruijn, line 268: You could also use a ref to select the container instead of using an id.

https://beta.reactjs.org/learn/manipulating-the-dom-with-refs#example-scrolling-to-an-element

src/pages/steps/CalendarStep/CalendarStep.tsx
Anahkiasen: That makes sense but in that case could we scroll to the title? I'm worried about the offset calculations being brittle since I've had bad experiences with those in the past depending on browser and platform although it's likely better now I suppose

brampauwelyn: Finally I tried again with the scrollIntoView and seems to work best with the `block` param  set to `center`.
https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView

src/pages/steps/CalendarStep/CalendarStep.tsx
brampauwelyn: Yes good idea

src/pages/steps/CalendarStep/CalendarStep.tsx
```diff
@@ -245,15 +250,32 @@ const CalendarStep = ({
     messages: { calendar: t('create.toast.success.calendar') },
   });
 
+  const scrollToCalendarContainer = () => {
+    calendarStepContainer.current.scrollIntoView({
+      behavior: 'smooth',
+      block: 'center',
+    });
+  };
```
simon-debruijn, line 258:
```suggestion
  const scrollToCalendarContainer = () => {
    if (!calendarStepContainer.current) {
      return;
    }
    calendarStepContainer.current.scrollIntoView({
      behavior: 'smooth',
      block: 'center',
    });
  };
```

Since the ref can be null in the beginning

src/pages/steps/CalendarStep/CalendarStep.tsx
```diff
@@ -245,15 +250,32 @@ const CalendarStep = ({
...
+    if (!scope || !type.id) return;
+    if (offerId) return;
+
+    scrollToCalendarContainer();
+  }, [scope, offerId, type]);
+
   return (
     <Stack
+      ref={calendarStepContainer}
+      id="calendar-step-container"
```
simon-debruijn, line 278:
```suggestion
```

This one can be removed I think, since we're using the ref now

PR #599, opened 2023-02-10
III-5269 - Add minHeight on Organizer paragraph title so texts are al…

brampauwelyn:
### Changed
- Add minHeight on Organizer paragraph title so texts are aligned with each other
Screenshot 2023-02-10 at 16 01 36

---
Ticket: https://jira.uitdatabank.be/browse/III-5269


Anahkiasen:
Would it be possible to leverage Flexbox's stretch and space-between? I fear the more we had all these hardcoded heights and widths the less robust and responsive it makes the design :/

simon-debruijn:
> Would it be possible to leverage Flexbox's stretch and/or space-between here perhaps? I fear the more we add all these hardcoded heights and widths the less robust and responsive it makes the design

I would merge it like this. 
You could make a follow up PR to rewrite this to leverage Flexbox's stretch and/or space-between if it isn't to time consuming. Otherwise it's best to focus on releasing the form asap I think.

PR #601, opened 2023-02-14
Fixup Storybook setup

Anahkiasen:
### Fixed

- Fixup Storybook setup

---

Wasn't able to run Storybook locally, I've tried a few different node versions and such but no luck. Tweaking the dependencies seems to have done the trick coupled with not running the Sentry plugin. Not sure if I was doing something wrong to begin with but I'll see what the CI says


simon-debruijn (approved):
The changes themselves look good. The 2 failing ci checks don't seem related, but that means PR's have been merged without succeeding ci checks. So we'll have to fix those separately.

Anahkiasen:
@simon-debruijn I think the types errors come from me having to run a yarn update which may have updated some types. I've tried to fix them by using the native types from the packages instead of redeclaring which should be more future proof in theory but not sure

simon-debruijn:
Ok for me 👍🏻

PR #604, opened 2023-02-14
Fix inability to create organizer from Typeahead

Anahkiasen:
### Fixed

- Fix inability to create organizer from Typeahead

---

Also fixed the two warnings about the missing `id` prop and the `labelKey` which must be a string when using `allowNew` (didn't see this cause any issue when picking existing organizations).

Ticket: https://jira.uitdatabank.be/browse/III-5332

### DIFF
--- a/src/pages/OrganizerAddModal.tsx
+++ b/src/pages/OrganizerAddModal.tsx
@@ -325,6 +325,7 @@ const OrganizerAddModal = ({
           <Title size={2}>Contact</Title>
           <ContactInfoStep
             isOrganizer={true}
+            onValidationChange={() => ({})}
             onSuccessfulChange={handleSetContactInfo}
             organizerContactInfo={contactInfo}
           />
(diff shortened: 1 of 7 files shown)


src/pages/OrganizerAddModal.tsx
```diff
@@ -325,6 +325,7 @@ const OrganizerAddModal = ({
           <Title size={2}>Contact</Title>
           <ContactInfoStep
             isOrganizer={true}
+            onValidationChange={() => ({})}
```
brampauwelyn, line 328: Why is this necessary?
If it's optional it should be ok right?

simon-debruijn: Also `() => ({})` doesn't match the type specified `(status: ValidationStatus) => void;`

SInce you're returning an object instead of void

Anahkiasen: I worked backward, since it was crashing because the prop was not provided I made it not optional, then provided it in a way that wouldn't crash, but it's probably better to amend the code to make it support the prop not being passed

simon-debruijn (approved):

PR #608, opened 2023-02-17
Only submit steps if they have changed

Anahkiasen:
### Changed

- Only trigger Steps onChange if they are dirty
- Make `Name` field always use the offer's main language instead of NL or the UI's

---

The reason for the Name change is because, since my offers have `fr` as main language, the code would always backfill `nl`, causing the field to be dirty when it wasn't. I also figured this was the correct way the code should behave but since the localization of the app has some edge cases I'm not sure this is correct or not.
I was also unsure of the best way to retrieve the offer's main language all the way down to the field since we only keep the ID for most of the components chain, so if there's a terser way that doesn't require passing so many props I'm up for it. I used the query cache in the mutation part but it's not always the safest option I think so not sure if we could/should use it in the Name field component as well

Ticket: https://jira.uitdatabank.be/browse/III-5251

### DIFF
--- a/src/pages/steps/NameAndAgeRangeStep.tsx
+++ b/src/pages/steps/NameAndAgeRangeStep.tsx
@@ -22,6 +22,7 @@ const useEditNameAndAgeRange = ({
   scope,
   onSuccess,
   offerId,
+  mainLanguage,
 }: UseEditArguments) => {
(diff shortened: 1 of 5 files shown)


brampauwelyn (approved):

simon-debruijn (approved):
Looks good 👍🏻

PR #644, opened 2023-03-01
Port migrate?location=true screen to React app

Anahkiasen:
### Added

- Port `migrate?location=true` screen to React app

---

Ticket: https://jira.uitdatabank.be/browse/III-5082

### DIFF
--- a/src/pages/steps/StepsForm.tsx
+++ b/src/pages/steps/StepsForm.tsx
@@ -23,7 +23,12 @@ import { usePublishOffer } from './hooks/usePublishOffer';
 import { PublishLaterModal } from './modals/PublishLaterModal';
 import { Steps, StepsConfiguration } from './Steps';
 import { Alert, AlertVariants } from '@/ui/Alert';
-import { locationStepConfiguration } from '@/pages/steps/LocationStep';
+import {
+  locationStepConfiguration,
+  useEditLocation,
+} from '@/pages/steps/LocationStep';
+import { useChangeLocationMutation } from '@/hooks/api/events';
+import { parseOfferId } from '@/utils/parseOfferId';
 
 const getValue = getValueFromTheme('createPage');
 
@@ -100,6 +105,18 @@ const StepsForm = ({
 
   const toast = useToast(toastConfiguration);
 
+  const useGetOffer = scope === OfferTypes.EVENTS ? useGetEvent : useGetPlace;
+
+  const offer = useGetOffer({
+    id: offerId,
+    onSuccess: (offer: Offer) => {
+      reset(convertOfferToFormData(offer), {
+        keepDirty: true,
+      });
+    },
+    enabled: !!scope,
+  });
+
   const publishOffer = usePublishOffer({
     scope,
     id: offerId,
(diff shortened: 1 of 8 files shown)


brampauwelyn (changes requested):
The location is already prefilled but the payload which is submitted to the api doesn't contain a `place['id']` which results in a `400`. 
This should be empty by default so the user is forced to select one...

Screenshot 2023-03-09 at 09 56 34

brampauwelyn:
But for the rest it looks good 👍

src/pages/steps/StepsForm.tsx
```diff
@@ -178,43 +199,70 @@ const StepsForm = ({
...
-          onChange={handleChange}
-          fieldLoading={fieldLoading}
-          onChangeSuccess={handleChangeSuccess}
-          offerId={offerId}
-          mainLanguage={offer?.mainLanguage}
-          scope={scope}
-          form={form}
-        />
+        {needsLocationMigration ? (
+          <>
```
simon-debruijn, line 203: In general you shouldn't use fragments in our codebase. Because our ui lib relies on passing margin props to its children. When you use a fragment instead of `Stack` the spacing prop of the parent will not work.

src/pages/steps/StepsForm.tsx
```diff
@@ -178,43 +199,70 @@ const StepsForm = ({
...
+            <Alert variant={AlertVariants.DANGER} marginY={5}>
+              <Trans
+                i18nKey={'create.migration.alert'}
```
simon-debruijn, line 206:
```suggestion
                i18nKey="create.migration.alert"
```

src/types/Offer.ts
```diff
@@ -138,6 +138,9 @@ type BaseOffer = {
 
 type Offer = Place | Event;
 
+export const usesLegacyLocation = (offer) =>
```
simon-debruijn, line 141:
```suggestion
export const hasLegacyLocation = (offer) =>
```

The uses is a bit confusing, because it looks like a hook

simon-debruijn (changes requested):

src/pages/steps/StepsForm.tsx
Anahkiasen: Oooh good to know 👍

src/types/Offer.ts
Anahkiasen: Yeah very true

src/pages/steps/StepsForm.tsx
```diff
@@ -178,43 +200,70 @@ const StepsForm = ({
...
+              offerId={offerId}
+              onChange={handleChange}
+              mainLanguage={offer?.mainLanguage}
+              scope={scope}
+              form={form}
+            />
+          </Stack>
+        ) : (
+          <Steps
+            configurations={configurations}
```
brampauwelyn, line 225: I wonder if you need to add a new `Steps` component inside the `needsLocationMigration` condtional.
Maybe you could write something like this? 

```suggestion
            configurations={needsLocationMigration ? [locationStepConfiguration] : configurations}
```

brampauwelyn (approved):

src/pages/steps/StepsForm.tsx
Anahkiasen: This was originally to not have the onChange/onHandleSuccess but ended up needing them anyway, will simplify the code

simon-debruijn (approved):

simon-debruijn (approved):

PR #650, opened 2023-03-06
Remove first outro

corneelwille:
### Changed
* Removed outro 1
* Made the "old" outro 2 outro 1 now

---

Ticket: https://jira.uitdatabank.be/browse/III-5413

### DIFF
--- a/src/i18n/de.json
+++ b/src/i18n/de.json
@@ -26,8 +26,7 @@
         "text": "Über die Vollständigkeitsbewertung erhalten Sie sofortiges Feedback und Tipps, wie Sie Ihre Veranstaltung oder Ihren Veranstaltungsort so vollständig wie möglich eingeben können."
       }
     ],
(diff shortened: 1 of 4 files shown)


corneelwille:
There is now no outro 2 anymore, I assume this won't break any code?

simon-debruijn:
> There is now no outro 2 anymore, I assume this won't break any code?

This will leave an empty p tag and too much spacing now. We also have to remove the `Paragraph` component that showed outro 2.

simon-debruijn (changes requested):
This will leave an empty p tag and too much spacing now. We also have to remove the Paragraph component that showed outro 2.

corneelwille:
Alternative solution that doesn't require any copy changes: set the max-width on `.hDJskr` to 90% (or so) again. Sorry about that!

simon-debruijn:
Screenshot 2023-03-07 at 09 22 45

simon-debruijn (approved):

PR #651, opened 2023-03-07
Add e2e testing setup

simon-debruijn:
### Added

- Setup [Playwright](https://playwright.dev/)
- Auth flow test/setup

### Changed

- Other ci jobs to use more recent versions of checkout


Anahkiasen (approved):
Niiice can't wait to play with it 👍

brampauwelyn (commented):
Noice 🧪

bertramakers (approved):
Very cool! 👏

.github/workflows/chromatic.yml
```diff
@@ -1,14 +1,17 @@
 name: 'Chromatic'
-on: push
+on:
+  pull_request:
+    types:
+      - closed
```
bertramakers, line 5: Does this only run the tests on closed PRs?

bertramakers: Ah no my bad this is the Chromatic job. Out of curiosity, why did that have to be changed?

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,40 @@
+name: E2E
+
+on:
+  pull_request:
+    types: [ready_for_review]
```
bertramakers, line 5: Will this also re-run the tests when you push new changes to a PR that is already in ready for review?

.github/workflows/chromatic.yml
simon-debruijn: I changed it so that it only published when a Pull Request has been merged, see the additional if check.
This is a optimisation separate from the e2e testing. We never really use the storybook generated per push, it seems like overkill.

.github/workflows/e2e.yml
simon-debruijn: I'm afraid not at the moment. I could probably add on `push` as type and than a conditional that checks if the PR was already marked as `ready_for_review`.

PR #658, opened 2023-03-10
Make Description field into a rich text editor

Anahkiasen:
### Changed

- Make Description field into a rich text editor

### Fixed

- Fixed description being filled with other language's when clearing the field

---

Two small notes:
1. There is currently a bug with Draft-related libraries because they switched to Rollup for bundling and didn't configure it properly for server-side rendering, breaking them in Next. This is why the imports are dynamic, I put a link to the relevant issues in the specific commit's description
2. The Description field in the preview currently uses `white-space: pre-line` which adds lots of whitespace, but this isn't an issue with the HTML generated, it's purely visual

https://user-images.githubusercontent.com/1321596/224306125-44b27837-01b3-4916-b382-40eac871d791.mp4

Ticket: https://jira.uitdatabank.be/browse/III-5115


src/pages/RichTextEditor.tsx
```diff
@@ -0,0 +1,35 @@
...
+    >
+      <Editor
+        {...props}
+        toolbar={{
+          options: ['inline', 'list', 'history'],
+          inline: {
+            inDropdown: false,
+            options: ['bold', 'italic', 'underline'],
+          },
+        }}
```
simon-debruijn, line 29:
```suggestion
        toolbar={{
          options: ['inline', 'list', 'history'],
          inline: {
            inDropdown: false,
            options: ['bold', 'italic', 'underline'],
          },
        }}
        {...props}
```

So props from higher up can override the `toolbar` prop, unless you want to prevent this.

src/pages/steps/AdditionalInformationStep/DescriptionStep.tsx
```diff
@@ -85,7 +92,11 @@ const DescriptionStep = ({
   // TODO: refactor
   const eventId = offerId;
 
-  const [description, setDescription] = useState('');
+  const [editorState, setEditorState] = useState(EditorState.createEmpty());
+  const plainTextDescription = useMemo(
+    () => editorState.getCurrentContent().getPlainText(),
+    [editorState],
```
simon-debruijn, line 98: is editorState a stable reference? because if it is a regular un-reffed object ti will cause the `useMemo` to never memoize the value.

package.json
```diff
@@ -34,16 +34,21 @@
...
     "crypto-browserify": "^3.12.0",
     "date-fns": "^2.16.1",
     "deep-object-diff": "^1.1.0",
+    "draft-js": "^0.11.7",
+    "draftjs-to-html": "^0.9.1",
     "fastest-levenshtein": "^1.0.12",
+    "html-to-draftjs": "^1.5.0",
     "i18next": "^19.8.3",
     "i18next-browser-languagedetector": "^6.0.1",
+    "immutable": "^4.2.4",
```
simon-debruijn, line 51: Didn't it work without specifying this specific version? 
Since it is included when defining the `draft-js` dependency

src/pages/steps/AdditionalInformationStep/DescriptionStep.tsx
Anahkiasen: Yeah true need to check that 🤔

simon-debruijn: Actually I think it's fine like this. The fact that it is initialized with `useState` and not a `const` with an object value.

Never mind my comment 😉

package.json
simon-debruijn: 👍🏻

simon-debruijn (approved):

src/pages/steps/AdditionalInformationStep/DescriptionStep.tsx
Anahkiasen: Ok I just checked and because it's an immutable structure it does seem to be able to memoize it properly even when the rest of the component changes 👍

simon-debruijn: I tested it quickly in [codesandbox](https://codesandbox.io/s/nifty-mountain-qq1r3g?file=/src/App.js).

Since the object is a value returned by `useState` it won't retrigger the `useMemo` when other state is changed.
When the object is declared as a plain `const` it will. So we're fine here I think 👍🏻

PR #661, opened 2023-03-15
Add Recent Locations to Location step

Anahkiasen:
### Added

- Add Recent Locations to Location step

---

The changes in `LocationStep` are not the easiest to follow along but it's because the way the step is built up is rather complex with lots of different branches so since the layout changed a lot this required a bunch of changes.

https://user-images.githubusercontent.com/1321596/225295765-24b38837-18e6-49cf-94b6-06b4244d8dd5.mp4

Ticket: https://jira.uitdatabank.be/browse/III-5268

### DIFF
--- a/src/pages/steps/LocationStep.tsx
+++ b/src/pages/steps/LocationStep.tsx
@@ -83,7 +83,7 @@ const RecentLocations = () => {
   const locations = uniqBy(
     offers?.map((offer) => offer.location),
     '@id',
-  );
+  ).filter((location) => location.name.nl !== 'Online');
   console.log(locations);
 
   return (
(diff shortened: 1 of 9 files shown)


src/pages/steps/LocationStep.tsx
```diff
@@ -58,6 +68,92 @@ const API_URL = publicRuntimeConfig.apiUrl;
...
+      },
+      paginationOptions: { start: 0, limit: 20 },
+    },
+    {
+      queryArguments: {
+        workflowStatus: 'DRAFT,READY_FOR_VALIDATION,APPROVED',
+        addressCountry: '*',
+      },
+    },
+  ) as { data: { member: (Offer & { location: any })[] } };
```
Anahkiasen, line 90: I know this is a bit hacky but it seems `UseAuthenticatedQueryResult` is currently not quite working since all the other places that use these queries have to resort to `ts-expect-error` every time :/

simon-debruijn: Yes you're correct, this is something we can improve on when we have time.

There is already a PR in draft (has been for a while)

To improve the typing of our react-query wrappers:

https://github.com/cultuurnet/udb3-frontend/pull/393

simon-debruijn: So for now casting or `ts-expect-error` is fine 👍🏻

src/ui/ButtonCard.tsx
```diff
@@ -0,0 +1,76 @@
...
+      variant={ButtonVariants.UNSTYLED}
+      customChildren
+      marginBottom={4}
+      title={title}
+      width="20rem"
+      {...props}
+      css={`
+        flex-direction: column;
+        align-items: flex-start;
+        background-color: rgba(255, 255, 255, 1);
```
brampauwelyn, line 31: Maybe use the colors from themes.ts? 
If it's not there yet, you can add it.

brampauwelyn (approved):

PR #664, opened 2023-03-17
III-5426 - Fix users getting stuck on step 4

simon-debruijn:
### Added

- `useCallback` in `useCalendarHandlers`

### Changed

- rename `event` to `offer`

### Removed

- Unnecessary extra logic to map xstate context to formData
Because of this code that was added later the formState is up to date.
Before this code was added the xstate context had to be mapped to formData before submitting POST request

### Fixed

- Fix scenarios that result in an invalid payload when creating an event/place
- `startDate` and `endDate` logic 

---

Ticket: https://jira.uitdatabank.be/browse/III-5426
Specific ticket: https://jira.uitdatabank.be/browse/III-5439


src/pages/steps/CalendarStep/FixedDays.tsx
```diff
@@ -106,8 +106,8 @@ export const FixedDays = ({
           spacing={3}
           id={`calendar-step-fixed`}
           minDate={new Date()}
-          dateStart={startDate ? new Date(startDate) : undefined}
-          dateEnd={endDate ? new Date(endDate) : undefined}
+          dateStart={new Date(startDate)}
+          dateEnd={new Date(endDate)}
```
brampauwelyn, line 110: What if startDate is undefined the Date constructor will throw an error?
https://github.com/cultuurnet/udb3-frontend/commit/a02c1a5a1881bf7a1a1ea0160fc9478287ded244

brampauwelyn (approved):
Did some quick testing and everything is working as supposed.

src/pages/steps/CalendarStep/FixedDays.tsx
simon-debruijn: The `startDate` and `endDate` will always be a valid date because they get a default value of today by `createInitialContext`.
The changes on this line make sure the startDate and endDate from the api can only be undefined or a Date:
https://github.com/cultuurnet/udb3-frontend/pull/664/files#diff-5984f8de41ce4e6471470d14c8344a51be07cf0a4c718eba8ae9d021a3f2f293R237

PR #690, opened 2023-04-03
III-5285 - Implement auth0 login

brampauwelyn:
### Changed
- Use UiTID domain in authentication flow with Auth0 instead of Auth0 domain

---

Ticket: https://jira.uitdatabank.be/browse/III-5285


src/pages/places/[placeId]/availability.test.js
```diff
@@ -90,8 +90,8 @@ test('I can save a status with a reason', async () => {
 
   await waitForFetch(`/places/${page.router.query.placeId}/status`);
 
-  // 3rd API call, [url, payload] tuple
-  expect(fetch.mock.calls[3][1].body).toEqual(
+  // 4th API call, [url, payload] tuple
```
Comment (Anahkiasen) [inline src/pages/places/[placeId]/availability.test.js:93]: Not sure I understand all those comments, I don't feel they provide much information that the code itself doesn't and they force you to have to keep both in sync

src/pages/events/[eventId]/availability.multiple.test.js
```diff
@@ -73,7 +73,7 @@ test('I can save a status', async () => {
   await waitForFetch(`/events/${page.router.query.eventId}/subEvents`);
 
   // 3rd API call, [url, payload] tuple
```
Comment (Anahkiasen) [inline src/pages/events/[eventId]/availability.multiple.test.js:75]: This should be 4th then perhaps

Anahkiasen (approved):
Would maybe remove the fetch comments but all good to me otherwise

src/pages/places/[placeId]/availability.test.js
Comment (brampauwelyn) [reply src/pages/places/[placeId]/availability.test.js]: @Anahkiasen yes you're right, I removed them 👍

src/context/UserContext.tsx
```diff
@@ -0,0 +1,7 @@
+import { UserProvider } from '@auth0/nextjs-auth0/client';
+
+const UserContext = ({ children }) => {
+  return <UserProvider>{children}</UserProvider>;
+};
```
simon-debruijn, line 5: Is this component still relevant? Since `UserProvider` looks like it can be used by itself.

simon-debruijn (approved):

PR #691, opened 2023-04-04
Do not add zipcode to /places query

Anahkiasen:
### Fixed

- Do not add zipcode to /places query

---

Ticket: https://jira.uitdatabank.be/browse/III-5504

### DIFF
--- a/src/hooks/api/places.ts
+++ b/src/hooks/api/places.ts
@@ -125,7 +125,6 @@ const getPlacesByQuery = async ({
   headers,
   name,
   terms,
-  zip,
   addressLocality,
   addressCountry,
 }: Headers & GetPlacesByQueryArguments) => {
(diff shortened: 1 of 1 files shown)


simon-debruijn (approved):

src/hooks/api/places.ts
```diff
@@ -135,7 +134,6 @@ const getPlacesByQuery = async ({
   );
   const queryArguments = [
     termsString,
-    zip ? `address.\\*.postalCode:"${zip}"` : '',
```
brampauwelyn, line 138: I think you should only omit the postalCode if the country is not belgium.

> However, for the Netherlands and Germany (not Belgium!) the postal code should be omitted in the dropdown.

brampauwelyn (changes requested):

src/hooks/api/places.ts
Anahkiasen: That was for the dropdown labels but I was wondering indeed it this wasn't just bringing us back to square one 🤔 Will add a condition around it

brampauwelyn (approved):

PR #693, opened 2023-04-04
Allow . as decimal separators in prices

Anahkiasen:
### Changed

- Allow . as decimal separators in prices

---

Since we use `Intl.NumberFormat` both `.` and `,` get automatically converted to `,` correctly already

https://user-images.githubusercontent.com/1321596/229780898-56224942-0ddc-474a-a606-29e5e0e068bf.mp4

Ticket: https://jira.uitdatabank.be/browse/III-5519


Anahkiasen:
It's converted to a float when sent in the backend so this is purely as far as the field formatting goes

PR #697, opened 2023-04-12
Ensure checking if a given form has a location doesn't crash on Movie…

Anahkiasen:
### Fixed

- Ensure checking if a given form has a location doesn't crash on Movie…


Anahkiasen:
@simon-debruijn That worked 👍  updated the fix

brampauwelyn (approved):

simon-debruijn (approved):

PR #711, opened 2023-04-27
Add validation to Labels management

Anahkiasen:
### Added

- Add validation to Labels management

---

Ticket: https://jira.uitdatabank.be/browse/III-5267

### DIFF
--- a/src/pages/steps/AdditionalInformationStep/LabelsStep.tsx
+++ b/src/pages/steps/AdditionalInformationStep/LabelsStep.tsx
@@ -98,6 +98,7 @@ function LabelsStep({
                   return setIsInvalid(true);
                 }
 
+                setIsInvalid(false);
                 await addLabelMutation.mutateAsync({
                   id: offerId,
                   scope,
--- a/src/pages/steps/AdditionalInformationStep/LabelsStep.tsx
+++ b/src/pages/steps/AdditionalInformationStep/LabelsStep.tsx
@@ -24,6 +24,7 @@ import { getGlobalBorderRadius } from '@/ui/theme';
 import { Typeahead } from '@/ui/Typeahead';
 
 type LabelsStepProps = StackProps & TabContentProps;
+const LABEL_PATTERN = /^[0-9a-zA-Z][0-9a-zA-Z-_]{0,48}[0-9a-zA-Z]$/;
 
 function LabelsStep({
   offerId,
@@ -57,7 +58,7 @@ function LabelsStep({
   }, [labels, onValidationChange]);
 
   const isWriting = addLabelMutation.isLoading || removeLabelMutation.isLoading;
-
+  const [isInvalid, setIsInvalid] = useState(false);
   return (
     <Stack {...getStackProps(props)} opacity={isWriting ? 0.5 : 1} spacing={2}>
       <FormElement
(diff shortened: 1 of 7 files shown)


src/pages/steps/RichAlert.tsx
```diff
@@ -0,0 +1,29 @@
+import { Alert } from '@/ui/Alert';
+import { Box, parseSpacing } from '@/ui/Box';
+
+function RichAlert({ children, ...props }) {
```
brampauwelyn, line 4: Is it necessary to create this new component? 
Is it not possible to use the regular `Alert` component?

brampauwelyn (changes requested):

src/pages/steps/RichAlert.tsx
```diff
@@ -0,0 +1,29 @@
...
+  return (
+    <Alert flex="1 0 auto" {...props}>
```
simon-debruijn, line 6:
```suggestion
    <Alert flex="1 0 auto" {...getInlineProps(props)}>
```

src/pages/steps/RichAlert.tsx
simon-debruijn, line 4: Type the props with the `Props` type from `Alert` the alert component

We also tend to stick to arrow functions everywhere in this project
```suggestion
const RichAlert = ({ children, ...props }: AlertProps) => {
```

simon-debruijn (changes requested):

src/pages/steps/RichAlert.tsx
Anahkiasen: It's possible since it's what it does underneath but I noticed quite a few places had these HTML alerts and repeated the same manual CSS to style them since by default ul/li/etc don't appear styled in alerts so it looked like an ok mid-level component to add to me. I can perhaps rephrase it as some kind of variant? Or should I perhaps add the custom `css` to all `Alert`s?

src/pages/steps/AdditionalInformationStep/LabelsStep.tsx
```diff
@@ -57,79 +59,98 @@ function LabelsStep({
...
+              allowNew
+              newSelectionPrefix={t(
+                'create.additionalInformation.labels.add_new_label',
+              )}
+              onSearch={setQuery}
+              onChange={async (newLabels: Label[]) => {
+                const label = newLabels[0]?.name;
+                if (!label || !label.match(LABEL_PATTERN)) {
+                  return setIsInvalid(true);
+                }
```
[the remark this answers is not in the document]
Anahkiasen: I can but I'd still need the early return which wouldn't save that much no?

```
const label = newLabels[0]?.name;
const isInvalid = !label || !label.match(LABEL_PATTERN);

setIsInvalid(isInvalid);
if (isInvalid) {
  return;
}

await addLabelMutation.mutateAsync({
  id: offerId,
  scope,
  label,
});
```

src/pages/steps/RichAlert.tsx
brampauwelyn: Yes I would prefer this approach.
What do you think @simon-debruijn

simon-debruijn: I Agree embedding it in the `Alert` component would be the preferred approach. Maybe by adding using a `richText` prop

simon-debruijn (changes requested):

brampauwelyn (approved):

simon-debruijn (approved):
I dit not check this time 😉

PR #714, opened 2023-05-15
III-5591 - Only do mutation when payload is different

brampauwelyn:
### Added
- `mutationKeys` to all mutations
- test for mutation cache mechanism

### Changed
- only execute mutation when the payload is different to the previous payload of the mutation with the same cacheKey

---
Ticket: https://jira.uitdatabank.be/browse/III-5591

### DIFF
--- a/src/hooks/api/authenticated-query.ts
+++ b/src/hooks/api/authenticated-query.ts
@@ -157,21 +157,29 @@ const useAuthenticatedMutation = ({ mutationFn, ...configuration }) => {
   const { removeAuthenticationCookies } = useCookiesWithOptions();
 
   const innerMutationFn = useCallback(async (variables) => {
-    const mutationCache = queryClient.getMutationCache().getAll();
+    try {
+      const mutationCache = queryClient.getMutationCache().getAll();
 
-    const cacheForMutationKey = mutationCache.filter((mutation) => {
-      return mutation.options.mutationKey === configuration.mutationKey;
-    });
+      const cacheForMutationKey = mutationCache.filter((mutation) => {
+        return (
+          typeof configuration.mutationKey !== 'undefined' &&
+          mutation.options.mutationKey === configuration.mutationKey
+        );
+      });
 
-    // get previous item from cache?
-    const latestMutation = cacheForMutationKey.at(-2);
+      // get previous item from cache?
+      const latestMutation = cacheForMutationKey.slice(-2)[0];
 
-    if (
-      // @ts-expect-error
-      latestMutation?.options?.variables &&
-      isEqual(latestMutation.options.variables, variables)
-    ) {
-      return false;
+      if (
+        // @ts-expect-error
+        cacheForMutationKey.length > 1 &&
+        latestMutation?.options?.variables &&
+        isEqual(latestMutation.options.variables, variables)
+      ) {
+        return false;
+      }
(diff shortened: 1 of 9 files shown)


src/hooks/api/authenticated-query.ts
```diff
@@ -151,10 +152,36 @@ const prefetchAuthenticatedQuery = async <TData>({
...
+      if (
+        // @ts-expect-error
+        cacheForMutationKey.length > 1 &&
+        latestMutation?.options?.variables &&
+        isEqual(latestMutation.options.variables, variables)
+      ) {
+        return false;
+      }
+    } catch (err) {
+      console.log({ err });
```
simon-debruijn, line 182:
```suggestion
      console.log(err);
```

simon-debruijn (approved):

src/hooks/api/authenticated-query.ts
```diff
@@ -151,10 +152,36 @@ const prefetchAuthenticatedQuery = async <TData>({
...
   const innerMutationFn = useCallback(async (variables) => {
+    try {
+      const mutationCache = queryClient.getMutationCache().getAll();
+
+      const cacheForMutationKey = mutationCache.filter((mutation) => {
+        return (
+          typeof configuration.mutationKey !== 'undefined' &&
+          mutation.options.mutationKey === configuration.mutationKey
+        );
+      });
```
Anahkiasen, line 168: I believe we can do `getMutationCache().find({ mutationKey: configuration.mutationKey })` here to retrieve a potential mutation from cache, judging from the TS types. This would return a single mutation object which would also prevent having to do the array stuff

Anahkiasen (changes requested):
Really cool solution, didn't think it was actually possible but I'm glad it is :D

src/hooks/api/authenticated-query.ts
simon-debruijn: Yes we also saw the function, but it isn't documented though. We suspect it maybe is a internal function.

https://tanstack.com/query/v3/docs/react/reference/MutationCache

Anahkiasen (approved):

src/hooks/api/authenticated-query.ts
simon-debruijn: Yeah we can use the `find` function, since it will possibly be a bit more performant

simon-debruijn: @Anahkiasen Could you take another look after my latest changes?

simon-debruijn: We still have to use the `findAll` instead of `find` because we always want the second last mutation. Because the mutation that gets executed gets added to the cache first before this check happens. Thus comparing against `mutations[0]` is comparing against itself.

src/hooks/api/authenticated-query.ts
```diff
@@ -149,6 +154,30 @@ const prefetchAuthenticatedQuery = async <TData>({
 
 /// /////////////////////////////////////////////////////////////////////////////////////////////
 
+const isDuplicateMutation = (
```
brampauwelyn, line 157: what a beautiful name 🦚

simon-debruijn: chatgipity is our friend

PR #722, opened 2023-05-17
III-5582 -  Make urls in entry form https

brampauwelyn:
### Changed
- Make urls https in:

  -     step 4 - input field for the online url
  -     step 6 contact tab - website url
  -     step 6 reservation tab - website url

---
Ticket: https://jira.uitdatabank.be/browse/III-5582

### DIFF
--- a/src/utils/url.ts
+++ b/src/utils/url.ts
@@ -1,7 +1,12 @@
 const prefixUrlWithHttp = (url: string) => {
-  if (url.startsWith('http')) return url;
+  if (url.startsWith('https://')) return url;
 
-  return `http://${url}`;
+  if (url.startsWith('http://')) {
+    const httpsUrl = url.replace('http://', 'https://');
+    return httpsUrl;
+  }
+
+  return `https://${url}`;
 };
 
 export { prefixUrlWithHttp };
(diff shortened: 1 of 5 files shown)


src/utils/url.ts
```diff
@@ -1,7 +1,12 @@
...
+  if (url.startsWith('https://')) return url;
 
-  return `http://${url}`;
+  if (url.startsWith('http://')) {
+    const httpsUrl = url.replace('http://', 'https://');
+    return httpsUrl;
+  }
+
+  return `https://${url}`;
 };
```
Anahkiasen, line 10: Could also do with a regex like

```suggestion
const prefixUrlWithHttps = (url: string) => 
    url.replace(/(https?:\/\/)?(.+)/, 'https://$2');
```

But depends how regex-averse yall feel 😄

simon-debruijn: 🤢

simon-debruijn: Just kidding, I'm fine with either approach

brampauwelyn: Yes, I try to avoid regexes as much as possible 😅 .
But I went for the regex since the name of the function should be descriptive enough.

PR #760, opened 2023-06-20
III-5688 - Add create place test

brampauwelyn:
### Added
- create place test

---
Ticket: https://jira.uitdatabank.be/browse/III-5688


upload/e2e-image.jpg
Anahkiasen, line 1: Would maybe put this in a Playwright folder instead like `src/test/e2e`

src/test/e2e/create-place.spec.ts
```diff
@@ -0,0 +1,148 @@
+import { expect, test } from '@playwright/test';
+
+const randomUniqueUrl = `https://e2e-organizer-${Math.random()
+  .toString(36)
+  .substring(2, 7)}.com`;
```
Anahkiasen, line 5: Would maybe add [faker](https://fakerjs.dev/) to the project since we'll likely need it in multiple E2E tests. We can also generate some `dummyPlace` attributes with it maybe

Anahkiasen (approved):
Two small suggestions but otherwise all good to me!

src/test/e2e/create-place.spec.ts
brampauwelyn: Yes good suggestion 👍

upload/e2e-image.jpg
brampauwelyn: I had some issues making it work when I placed the upload folder inside the playwright folder...

PR #761, opened 2023-06-23
III-5687 - Add E2E test for creating a movie

Anahkiasen:
### Added

- Add end-to-end test for creating a movie

---

Ticket: https://jira.uitdatabank.be/browse/III-5687


Anahkiasen:
Marking as ready for review purely to run E2E CI

package.json
```diff
@@ -80,7 +80,7 @@
   },
   "devDependencies": {
     "@faker-js/faker": "^8.0.2",
-    "@playwright/test": "^1.31.2",
+    "@playwright/test": "1.31.2",
```
brampauwelyn, line 83: Is it still necessary to lock the version to `1.31.2`?

src/test/e2e/setup/auth.setup.ts
```diff
@@ -23,7 +23,7 @@ setup('authenticate', async ({ baseURL, page }) => {
   // Wait for network to be idle, if we save storage too early, needed storage values might not yet be available
   await page.waitForLoadState('networkidle');
 
-  await page.getByText('Welkom, end to end').waitFor();
+  await page.getByText('Welkom').waitFor();
```
brampauwelyn, line 26: Was it necessary to remove the name?
```suggestion
  await page.getByText('Welkom, end to end').waitFor();
```

src/ui/TimeTable.tsx
```diff
@@ -440,6 +440,7 @@ const TimeTable = ({ id, className, onChange, value, ...props }: Props) => {
         onDateEndChange={handleDateEndChange}
       />
       <Stack
+        id={'timetable'}
```
brampauwelyn, line 443:
```suggestion
        id="timetable"
```

brampauwelyn (approved):

brampauwelyn:
Cool 🎬

src/test/e2e/setup/auth.setup.ts
Anahkiasen: This was to make the tests more flexible to using different `E2E_TEST_` variables since currently you can configure them but the tests expect this very specific user

package.json
Anahkiasen: No it was cause my local Node is too recent so it installs the wrong one, will revert the locking

PR #787, opened 2023-07-24
Add images step to organizer form

Anahkiasen:
### Added

- Add images step to organizer form

---

Ticket: https://jira.uitdatabank.be/browse/III-5676

### DIFF
--- a/src/pages/steps/AdditionalInformationStep/FormScore.tsx
+++ b/src/pages/steps/AdditionalInformationStep/FormScore.tsx
@@ -185,7 +185,7 @@ const getMinimumScore = (weights: Weights): number => {
   return minimumScore;
 };
 
-const FormScore = ({ completedFields, offerId, scope, ...props }: Props) => {
+const FormScore = ({ completedFields, offerId, scope }: Props) => {
   const { t } = useTranslation();
 
   const router = useRouter();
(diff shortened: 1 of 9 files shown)


src/pages/steps/AdditionalInformationStep/FormScore.tsx
```diff
@@ -128,21 +129,51 @@ const scoreWeightMapping: Weights = {
...
-    weights.location.weight = 10;
-    weights.description.weight = 15;
-  }
+type EntityWithMedia =
+  | (Offer & { images: undefined })
+  | (Organizer & {
+      terms: undefined;
+      mediaObject: undefined;
+      videos: undefined;
+    });
```
simon-debruijn, line 171: Is your intention to show offer can have images or no images, etc?

Because you can make the properties optional on the `Offer` and `Organizer` if they're not correct.

simon-debruijn (approved):

Anahkiasen:
Oops forgot the translations once again 😅 Lemme just add that

src/pages/steps/AdditionalInformationStep/FormScore.tsx
Anahkiasen: No all the properties are correct, but their presence/name/required is not the same for offers as for organizers. Offers have `mediaObject` while Organizer uses `images`, in the same way organizers don't have `terms`. This type is just a "trick" to make not have to `expect-error` 20 lines in the component cause we do a lot of checks to properties that only exist in one or the other :/

Anahkiasen: I think the correct handling of this could perhaps be having a `isOrganizer` [type predicate function](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates) which would be able to make Typescript differentiate which object we're dealing with, as currently even with coalesce operators TS complains since it's more about the existence of the property on the objects than them being defined

PR #796, opened 2023-08-10
Allow common accented letters in labels

Anahkiasen:
### Changed

- Allow common accented letters in labels

---

Ticket: https://jira.uitdatabank.be/browse/III-5768

### DIFF
--- a/src/pages/steps/AdditionalInformationStep/LabelsStep.tsx
+++ b/src/pages/steps/AdditionalInformationStep/LabelsStep.tsx
@@ -26,7 +26,7 @@ import { Typeahead } from '@/ui/Typeahead';
 
 type LabelsStepProps = StackProps & TabContentProps;
 
-const LABEL_PATTERN = /^[0-9a-zA-Z][0-9a-zA-Z-_\s]{1,49}$/;
+const LABEL_PATTERN = /^[0-9a-zA-ZÀ-ÿ][0-9a-zA-ZÀ-ÿ\-_\s]{1,49}$/;
 
 function LabelsStep({
   offerId,


brampauwelyn (approved):

src/pages/steps/AdditionalInformationStep/LabelsStep.tsx
```diff
@@ -26,7 +26,7 @@ import { Typeahead } from '@/ui/Typeahead';
 
 type LabelsStepProps = StackProps & TabContentProps;
 
-const LABEL_PATTERN = /^[0-9a-zA-Z][0-9a-zA-Z-_\s]{1,49}$/;
+const LABEL_PATTERN = /^[0-9a-zA-ZÀ-ÿ][0-9a-zA-ZÀ-ÿ\-_\s]{1,49}$/;
```
simon-debruijn, line 29: According to @LucWollants this REGEX in the backend may be used by the frontend to validate labels.
https://github.com/cultuurnet/udb3-backend/blob/master/src/Model/ValueObject/Taxonomy/Label/LabelName.php#L19

Maybe we should align the regexes. Maybe there are important differences @corneelwille knows about

simon-debruijn (changes requested):

src/pages/steps/AdditionalInformationStep/LabelsStep.tsx
LucWollants: Also important to note is that inside the admin interface where labels are managed, this rule should not be applied. Not sure if any code is shared.

Anahkiasen: I think they're in essence the same but back-end uses a negative lookahead while we repeat the capturing group twice, so the backend's is terser albeit less readable. But from tests they behave the same, although would prefer confirmation to be sure in any case

simon-debruijn (approved):

PR #797, opened 2023-08-11
Update user queries to new query function

Anahkiasen:
### Changed

- Update user queries to new query function

---

Ticket: https://jira.uitdatabank.be/browse/III-5778

### DIFF
--- a/src/pages/dashboard/index.page.tsx
+++ b/src/pages/dashboard/index.page.tsx
@@ -725,7 +725,7 @@ const Dashboard = (): any => {
 
 const getServerSideProps = getApplicationServerSideProps(
   async ({ req, query, cookies: rawCookies, queryClient }) => {
-    const user = await useGetUserQueryServerSide({
+    const { data: user } = await useGetUserQueryServerSide({
       req,
       queryClient,
     });
--- a/src/pages/dashboard/index.page.tsx
+++ b/src/pages/dashboard/index.page.tsx
@@ -733,11 +733,10 @@ const getServerSideProps = getApplicationServerSideProps(
     await Promise.all(
       Object.entries(UseGetItemsByCreatorMap).map(([key, hook]) => {
         const page =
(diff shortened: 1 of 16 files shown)


src/pages/dashboard/index.page.tsx
```diff
@@ -730,18 +725,19 @@ const Dashboard = (): any => {
...
       req,
       queryClient,
-    })) as User;
+    });
 
     await Promise.all(
       Object.entries(UseGetItemsByCreatorMap).map(([key, hook]) => {
         const page =
-          query.tab === key ? (query.page ? parseInt(query.page) : 1) : 1;
+          query.tab === key ? parseInt((query.page as string) ?? '1') : 1;
```
simon-debruijn, line 736: This logic doesn't handle `query.page` with value empty string well

src/utils/getApplicationServerSideProps.ts
```diff
@@ -76,22 +81,31 @@ const redirectToLogin = (cookies, req, resolvedUrl) => {
...
     const { publicRuntimeConfig } = getConfig();
     if (publicRuntimeConfig.environment === 'development') {
-      process.env.NODE_TLS_REJECT_UNAUTHORIZED = 0;
+      process.env.NODE_TLS_REJECT_UNAUTHORIZED = '0';
     }
 
-    const rawCookies = req?.headers?.cookie ?? '';
-
-    const cookies = new Cookies(rawCookies, defaultCookieOptions);
+    const cookies = new Cookies(req.cookies);
```
simon-debruijn, line 102: Aren't the `defaultCookieOptions` important?

simon-debruijn (approved):
Looks good overal, nice work ⭐

src/utils/getApplicationServerSideProps.ts
Anahkiasen: Typescript caught that this is actually not the place to pass them, this only accepts a custom cookie decoder but the options were ignored otherwise

PR #808, opened 2023-10-11
III-5746 - Also use hiddenLabels for showing cultuurkuur tab on organiser

brampauwelyn:
### Changed
- Also use hiddenLabels for showing cultuurkuur tab on organiser

---
Ticket: https://jira.uitdatabank.be/browse/III-5746

### DIFF
--- a/src/pages/organizers/create/OrganizerForm.tsx
+++ b/src/pages/organizers/create/OrganizerForm.tsx
@@ -111,15 +111,10 @@ const OrganizerForm = () => {
   // @ts-expect-error
   const organizer = getOrganizerByIdQuery?.data;
 
-  const organizerLabelsSet = new Set<string>();
-
-  (organizer?.labels ?? []).map((label: string) =>
-    organizerLabelsSet.add(label),
-  );
-
-  (organizer?.hiddenLabels ?? []).map((label: string) =>
-    organizerLabelsSet.add(label),
-  );
+  const organizerLabelsSet = new Set([
+    ...(organizer?.labels ?? []),
+    ...(organizer?.hiddenLabels ?? []),
+  ]);
 
   const organizerLabels = [...organizerLabelsSet];
 
--- a/src/pages/organizers/create/OrganizerForm.tsx
+++ b/src/pages/organizers/create/OrganizerForm.tsx
@@ -111,7 +111,17 @@ const OrganizerForm = () => {
   // @ts-expect-error
   const organizer = getOrganizerByIdQuery?.data;
 
-  const organizerLabels = organizer?.labels;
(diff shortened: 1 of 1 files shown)


src/pages/organizers/create/OrganizerForm.tsx
```diff
@@ -111,7 +111,17 @@ const OrganizerForm = () => {
   // @ts-expect-error
   const organizer = getOrganizerByIdQuery?.data;
 
-  const organizerLabels = organizer?.labels;
+  const organizerLabelsSet = new Set();
```
simon-debruijn, line 114:
```suggestion
  const organizerLabelsSet = new Set<string>();
```

simon-debruijn: This will fix your ts issue

simon-debruijn (changes requested):

src/pages/organizers/create/OrganizerForm.tsx
brampauwelyn: Nice thx

src/pages/organizers/create/OrganizerForm.tsx
```diff
@@ -111,7 +111,17 @@ const OrganizerForm = () => {
...
-  const organizerLabels = organizer?.labels;
+  const organizerLabelsSet = new Set<string>();
+
+  (organizer?.labels ?? []).map((label: string) =>
+    organizerLabelsSet.add(label),
+  );
+
+  (organizer?.hiddenLabels ?? []).map((label: string) =>
+    organizerLabelsSet.add(label),
+  );
```
Anahkiasen, line 122: I would maybe build the Set by providing the arrays directly to avoid having to loop and add each entry separately:

```ts
const organizerLabelsSet = new Set(
    organizer?.labels ?? [], 
    organizer?.hiddenLabels ?? []
);
```

Should work equally in theory:

```js
new Set(['foo','bar'], ['foo'], ['bar'])
> Set [ "foo", "bar" ]
```

brampauwelyn: I doesn't work since new Set(), only allows 1 argument. The 2nd argument will be ignored...
So e.g : 

```suggestion
new Set(['foo','bar'], ['foo'], ['hello'])
```

Will return
```suggestion
['foo','bar']
```
The "hello" is ignored...

Anahkiasen: Yes sorry I should have tried with better dummy string 😅 but I would still do like `new Set([...a, ...b])` perhaps at least to avoid remaping the entries by one

simon-debruijn (approved):

PR #815, opened 2023-10-18
III-5851 - Implementation new design

brampauwelyn:
### Changed
- Implementation new design

Screenshots:
Screenshot 2023-11-07 at 14 51 08

Screenshot 2023-11-06 at 12 54 09

Screenshot 2023-11-07 at 13 56 12

Angular part: https://github.com/cultuurnet/udb3-angular-app/pull/205

---
Ticket: https://jira.uitdatabank.be/browse/III-5851


src/layouts/Sidebar.tsx
```diff
@@ -58,6 +59,8 @@ const getValueForMenuItem = getValueFromTheme('menuItem');
 const getValueForSidebar = getValueFromTheme('sidebar');
 const getValueForMenu = getValueFromTheme('menu');
 
+const getValue = getValueFromTheme('sidebar');
```
Anahkiasen, line 62: Can likely reuse `getValueForSidebar` from above

src/layouts/Sidebar.tsx
```diff
@@ -264,6 +297,16 @@ const NotificationMenu = memo(
...
         onClick: onClickJobLoggerButton,
       },
+      {
+        iconName: Icons.SIGN_OUT,
+        children: t('menu.logout'),
+        onClick: async () => {
+          removeAuthenticationCookies();
+          await queryClient.invalidateQueries('user');
+
+          window.location.assign('/api/auth/logout');
```
Anahkiasen, line 307: Would maybe use `router.push` here to be sure it doesn't do anything special underneath in addition to switching location

src/pages/steps/Steps.tsx
```diff
@@ -94,18 +94,20 @@ const StepWrapper = ({
...
-          border-bottom: 1px solid ${getValue('title.borderColor')};
-        `}
-      >
+    <Stack
+      paddingX={5}
+      paddingY={4.5}
+      borderRadius={getGlobalBorderRadius}
+      backgroundColor="white"
+      css={`
+        box-shadow: 0px 4px 16px 0px #00000012;
```
Anahkiasen, line 103: Is this hex correct?

src/layouts/Sidebar.tsx
brampauwelyn: True, I think it was like this for a reason but I don't remember and it seems to work with `router.push`

src/pages/steps/Steps.tsx
brampauwelyn: Yes, but I used a theme value instead of a hardcoded value.

Anahkiasen (approved):

src/layouts/Sidebar.tsx
```diff
@@ -92,30 +93,71 @@ const MenuItem = memo(
     onClick,
     visible = true,
   }: MenuItemProps) => {
+    const { asPath } = useRouter();
+
+    const baseUrl = publicRuntimeConfig.baseUrl;
+    const currentPath = new URL(asPath, baseUrl).pathname;
```
simon-debruijn, line 99: Doesn't `useRouter` also expose a normal path?

simon-debruijn (approved):

src/layouts/Sidebar.tsx
brampauwelyn: I don't remember exactly why but I wasn't able to do what I wanted with the variables exposed by useRouter.

PR #816, opened 2023-10-18
Fixed Typahead filtering results a second time

Anahkiasen:
### Fixed

- Fixed Typahead filtering results a second time

---

There are more files changed here than the issue warranted but here's what happened:

1. The crux of the issue was that there is a default `filterBy` prop that automatically filters the results based on the input query. But because our API already does the filtering this was removing already valid results because `Typeahead`'s logic is less lenient than ours (hence the double space issue). So when not provided we now pass a filter of `() => true` which is the documented way to disable that extra filtering
2. Fixing this I noticed that it didn't recognize `filterBy` properly, and that this was due to us rewriting the props to Typeahead so I made the code use the types that come with the package instead.
3. Because Typeahead is just a `Box` + forward, updating the Box types uncovered some issues with other descendants of it, including `Input` which was also overriding its native props, so I also switched that one to the built-in props 
4. Finally because I moved a bunch of stuff around I just ran a quick format which optimized a few imports

All that to say, while this changes more than needed, it also means less work to maintain the Input/Typahead props and ensures we stay close to the actual API in case we update any of those packages and their props change. 

I still think it's not ideal for `Box` to have the combined types of all its possible descendants, as a lot of props have the same name and thus the signature get merged, creating weird invalid types. But that's a much bigger change so I let that as is.

Ticket: https://jira.publiq.be/browse/III-5802


src/pages/steps/AdditionalInformationStep/BookingInfoStep.tsx
```diff
@@ -485,9 +485,7 @@ const BookingInfoStep = ({
           as="form"
           width="45%"
           spacing={4}
-          onBlur={handleSubmit((data) => {
-            handleAddBookingInfoMutation(data);
-          })}
+          onBlur={() => handleAddBookingInfoMutation(getValues())}
```
brampauwelyn, line 488: Is the `handleSubmit` not necessary anymore?

brampauwelyn (approved):

src/pages/steps/AdditionalInformationStep/BookingInfoStep.tsx
Anahkiasen: The issue was that the signatures were incompatible since you can't usually use it on an ù onBlur`, it would still technically work because we don't rely on the event itself but this was bending `handleSubmit`'s intended purpose a bit so I went for a pure `getValues` to please TS + me :D

PR #819, opened 2023-10-20
Only show URL requirements when URL is unique but incorrect

Anahkiasen:
### Changed

- Only show URL requirements when URL is unique but incorrect

---

Ticket: https://jira.publiq.be/browse/III-5858


Anahkiasen:
Missing the success alert, will add it

brampauwelyn (approved):

PR #850, opened 2023-11-22
Changed editor Link popup to a custom one

Anahkiasen:
### Changed

- Changed editor Link popup to a custom one

---

Unfortunately `react-draft-wysiwyg` does not provide any way to override simply the popup, or render things into it. It also doesn't expose any of the internal components for extension. So I had to reimplement the Link component from its source, removing the things we don't use to avoid too much dead code in it. 

On the upside since this then allows us to use our own component, the popup is now much closer to the UI and in a more future-proof way than through CSS overrides, which is a big plus so to me it was still worth it. 

Screenshot 2023-11-22 at 16 32 57

Ticket: https://jira.publiq.be/browse/III-5893


src/pages/CustomRichTextEditorLink.tsx
```diff
@@ -0,0 +1,189 @@
...
+import { Button, ButtonVariants } from '@/ui/Button';
+import { FormElement } from '@/ui/FormElement';
+import { Inline } from '@/ui/Inline';
+import { Input } from '@/ui/Input';
+import { Stack } from '@/ui/Stack';
+
+/**
+ * Source: https://github.com/jpuri/react-draft-wysiwyg/blob/master/src/components/Option/index.js#L8
+ */
+function Option({
```
simon-debruijn, line 14: Could you use arrow functions for codebase consistency? Unless y'all have changed the convention since I last worked on it 😄

src/pages/CustomRichTextEditorLink.tsx
```diff
@@ -0,0 +1,189 @@
...
+  onClick,
+  children,
+  value,
+  className,
+  activeClassName = '',
+  active = false,
+  disabled = false,
+  title,
+}) {
```
simon-debruijn, line 23: I would still add prop types here

src/pages/CustomRichTextEditorLink.tsx
```diff
@@ -0,0 +1,189 @@
...
+      {children}
+    </div>
+  );
+}
+
+/**
+ * Source: https://github.com/jpuri/react-draft-wysiwyg/blob/master/src/controls/Link/Component/index.js#L203
+ */
+
+function CustomRichTextEditorLink({
```
simon-debruijn, line 47: Same remark about arrow functions

src/pages/CustomRichTextEditorLink.tsx
```diff
@@ -0,0 +1,189 @@
...
+  const { t } = useTranslation();
+  const [showModal, setShowModal] = useState(false);
+  const [linkTarget, setLinkTarget] = useState('');
+  const [linkTitle, setLinkTitle] = useState('');
+  const [linkTargetOption, setLinkTargetOption] = useState(
+    config.defaultTargetOption,
+  );
+
+  useEffect(() => {
+    if (expanded && !expanded) {
```
simon-debruijn, line 65:
```suggestion
    if (expanded === false) {
```
Is this the expected behaviour? The current check looks strange

src/pages/CustomRichTextEditorLink.tsx
```diff
@@ -0,0 +1,189 @@
...
+      setShowModal(false);
+      setLinkTarget('');
+      setLinkTitle('');
+      setLinkTargetOption(config.defaultTargetOption);
+    }
+  }, [expanded]);
+
+  const updateValue = (event) => {
+    const { name, value } = event.target;
```
simon-debruijn, line 74: The event type can be added

src/pages/CustomRichTextEditorLink.tsx
```diff
@@ -0,0 +1,189 @@
...
+  config,
+  currentState,
+  doCollapse,
+  expanded,
+  onChange,
+  onExpandEvent,
+  translations,
+}: CustomRichTextEditorLinkProps) {
+  const { t } = useTranslation();
+  const [showModal, setShowModal] = useState(false);
```
simon-debruijn, line 57:
```suggestion
  const [isModalVisible, setIsModalVisible] = useState(false);
```
This naming makes it more clear that it is a boolean and a some function

src/pages/CustomRichTextEditorLink.tsx
```diff
@@ -0,0 +1,189 @@
...
+  const updateValue = (event) => {
+    const { name, value } = event.target;
+    if (name === 'linkTarget') {
+      setLinkTarget(value);
+    } else if (name === 'linkTitle') {
+      setLinkTitle(value);
+    }
+  };
+
+  const signalExpandShowModal = () => {
```
simon-debruijn, line 82: Can you find a clearer name for this function?

Usually we start functions that handle some event with `handle...`

e.g. `handleExpandModal` or something

simon-debruijn (changes requested):

Anahkiasen:
@simon-debruijn Sorry for the messy code, this is the code from the package I just got carried away converting it hastily out of class components but the result is a mix of their conventions and ours, will make it a bit nicer

src/pages/CustomRichTextEditorLink.tsx
```diff
@@ -143,7 +151,7 @@ function CustomRichTextEditorLink({
             Component={
               <Input
                 onChange={updateValue}
-                onBlur={updateValue}
+                onBlur={(event) => {}}
```
simon-debruijn, line 154: I see you have a no-op here now, is this intentional?

simon-debruijn (changes requested):
Thanks for the changes 👍🏻 just 1 remark I left now

src/pages/CustomRichTextEditorLink.tsx
Anahkiasen: Oops not just leftover tinkering to check the proper types

src/pages/login/illustrations/WomanIllustration.tsx
```diff
@@ -157,173 +157,173 @@ const WomanIllustrationSvg = ({ width, ...props }: Props) => {
...
           d="M414.303 193.433v.456c0 .183-.092.457-.183.73l-1.642 6.114c-1.004 3.742-4.289 6.297-8.122 6.479l-8.943.365c-4.38.183-8.212-2.92-9.034-7.209l-.912-4.927-.639-3.559a9.144 9.144 0 0 1 8.213-10.95h.73c4.38 0 8.213 3.193 8.943 7.482l1.095 6.023.182.821.639-.821 2.737-3.468c.73-.912 1.917-1.46 3.103-1.46 2.099 0 3.833 1.825 3.833 3.924Z"
           stroke="#009FDF"
-          stroke-width="1.825"
-          stroke-miterlimit="10"
-          stroke-linecap="round"
-          stroke-linejoin="round"
+          strokeWidth="1.825"
+          strokeMiterlimit="10"
+          strokeLinecap="round"
+          strokeLinejoin="round"
```
brampauwelyn, line 222: Ooh sorry for those

brampauwelyn (approved):

PR #876, opened 2024-01-09
Allow picking dates in the past in Datepicker

Anahkiasen:
### Changed

- Allow picking dates in the past in Datepicker

---

Ticket: https://jira.publiq.be/browse/III-5967

### DIFF
--- a/src/pages/steps/CalendarStep/Days.tsx
+++ b/src/pages/steps/CalendarStep/Days.tsx
@@ -102,7 +102,6 @@ export const Days = ({
               <DatePeriodPicker
                 spacing={3}
                 id={`calendar-step-day-${day.id}`}
-                minDate={new Date()}
                 dateStart={new Date(day.startDate)}
                 dateEnd={new Date(day.endDate)}
                 onDateStartChange={(newDate) =>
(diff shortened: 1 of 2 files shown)


src/pages/steps/CalendarStep/Days.tsx
```diff
@@ -102,7 +102,6 @@ export const Days = ({
               <DatePeriodPicker
                 spacing={3}
                 id={`calendar-step-day-${day.id}`}
-                minDate={new Date()}
```
brampauwelyn, line 105: What is the default date when you enter a new event?

brampauwelyn (approved):

src/ui/DatePeriodPicker.tsx
```diff
@@ -67,7 +67,7 @@ const DatePeriodPicker = ({
           minDate={
             dateStart && dateStart.getTime() > new Date().getTime()
               ? dateStart
-              : new Date()
+              : minDate
```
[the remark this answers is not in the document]
Anahkiasen: It's the prop received by the component, it's to avoid that prop being overriden in case that ternary condition fails, otherwise the prop is not respected.

src/pages/steps/CalendarStep/Days.tsx
Anahkiasen: Still the current date so should be similar in behavior

PR #890, opened 2024-04-08
III-5789 - Improve hasLegacyLocation check

brampauwelyn:
### Fixed
- Logic in hasLegacyLocation check (should only return true on Events with a missing location id, so it shouldn't show the alert on places).

Before:
Screenshot 2024-04-08 at 16 06 44

After:
Screenshot 2024-04-08 at 15 56 01

---
Ticket: https://jira.uitdatabank.be/browse/III-5789

### DIFF
--- a/src/types/Offer.ts
+++ b/src/types/Offer.ts
@@ -2,6 +2,7 @@ import type { BookingAvailabilityType } from '@/constants/BookingAvailabilityTyp
 import type { CalendarType } from '@/constants/CalendarType';
 import type { OfferStatus } from '@/constants/OfferStatus';
 import { PriceCategory } from '@/pages/steps/AdditionalInformationStep/PriceInformation';
+import { parseOfferType } from '@/utils/parseOfferType';
 
 import type { SupportedLanguages } from '../i18n';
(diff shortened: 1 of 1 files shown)


src/types/Offer.ts
```diff
@@ -146,8 +147,15 @@ type Label = {
   excluded: boolean;
 };
 
-export const hasLegacyLocation = (offer) =>
-  offer?.location && !offer?.location?.['@id'];
+export const hasLegacyLocation = (offer: Offer) => {
+  if (!offer) return false;
+
+  if (parseOfferType(offer['@context']) === 'place') {
```
simon-debruijn, line 153: I think there is also a isPlace utility function?

simon-debruijn: If a isEvent function exists, you can check if (!isEvent(offer)) return false.

Then (offer as Event won't be needed)

simon-debruijn (changes requested):

src/types/Offer.ts
brampauwelyn: 👍

simon-debruijn (approved):

PR #896, opened 2024-04-17
Changed appearance of character counter

Anahkiasen:
### Changed

- Changed appearance of character counter

---

Ticket: https://jira.publiq.be/browse/III-6114

Screenshot 2024-04-17 at 12 02 25
Screenshot 2024-04-17 at 12 02 28

### DIFF
--- a/src/ui/FormElement.tsx
+++ b/src/ui/FormElement.tsx
@@ -36,6 +36,7 @@ const MaxLengthCounter = ({
       fontSize="0.9rem"
       className="text-right"
       maxWidth="43rem"
+      color={currentLength >= maxLength ? 'red' : 'inherit'}
     >
       {currentLength} / {maxLength}
     </Text>
--- a/src/ui/FormElement.tsx
+++ b/src/ui/FormElement.tsx
@@ -120,7 +120,7 @@ const FormElement = ({
         width={labelPosition === LabelPositions.RIGHT ? 'auto' : '100%'}
         minWidth={50}
       >
-        {maxLength && (
+        {typeof maxLength !== 'undefined' && (
           <Inline justifyContent="space-between" maxWidth="43rem">
             <span>{info && infoElement}</span>
             <MaxLengthCounter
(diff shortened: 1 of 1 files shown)


src/ui/FormElement.tsx
```diff
@@ -36,6 +36,7 @@ const MaxLengthCounter = ({
       fontSize="0.9rem"
       className="text-right"
       maxWidth="43rem"
+      color={currentLength >= maxLength ? 'red' : 'inherit'}
```
simon-debruijn, line 39: What in the case of maxLength undefined?

src/ui/FormElement.tsx
```diff
@@ -112,6 +120,15 @@ const FormElement = ({
         width={labelPosition === LabelPositions.RIGHT ? 'auto' : '100%'}
         minWidth={50}
       >
+        {maxLength && (
```
simon-debruijn, line 123:
```suggestion
        {typeof maxLength !== 'undefined' && (
```

maxLength 0 is a bit stupid, but better to be explicit to avoid bugs where 0 evaluates to false

src/ui/FormElement.tsx
```diff
@@ -121,32 +138,7 @@ const FormElement = ({
...
-          />
-        )}
-        {!maxLength &&
-          info &&
-          (typeof info === 'string' ? (
-            <Text variant={TextVariants.MUTED}>{info}</Text>
-          ) : (
-            info
-          ))}
+        {!maxLength && info && infoElement}
```
simon-debruijn, line 141:
```suggestion
        {typeof maxLength === 'undefined' && info && infoElement}
```

simon-debruijn (approved):

src/ui/FormElement.tsx
Anahkiasen: This should be good now since this component never renders if we have no maxLength but I can build the security in the component itself worst case

PR #897, opened 2024-04-18
Fixed setting price to free not triggering validation refresh

Anahkiasen:
### Fixed

- Fixed setting price to free not triggering validation refresh

---

Overriding the price value does not seem to trigger the `useEffect` that handles validation, neither did trying to refetch or invalidate or blur the fields or something. Since this form has been a lot of back and forth to make it both reactive and optimistic I just reset the offer data to trigger the effect, although I'm still confused as to why that's the only solution that seems to work but since we handle the price mutation in a special way I'm guessing it's related? 
I'm unsure why `refetch` doesn't work but `remove` does, I thought both would yield the same result in the end but could be mistaken since the data is automatically refetched upon removal but I don't know 🤔 

Ticket: https://jira.publiq.be/browse/III-6008

### DIFF
--- a/src/pages/steps/AdditionalInformationStep/PriceInformation.tsx
+++ b/src/pages/steps/AdditionalInformationStep/PriceInformation.tsx
@@ -357,6 +357,7 @@ const PriceInformation = ({
                                 shouldValidate: false,
                               });
                               onSubmit();
+                              // @ts-expect-error
                               getOfferByIdQuery.remove();
                             }}
                           >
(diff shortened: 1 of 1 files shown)


brampauwelyn (approved):
The change itself looks good.
Did you test if the edit functionality of the price still works as expected?

Anahkiasen:
> Did you test if the edit functionality of the price still works as expected?

Just did a bunch of back and forth and everything seemed persisted properly, same for the check it appeared at the right time when not immediately setting to Free

PR #902, opened 2024-04-29
"Fixed" calendar not listening for changes after initial save

Anahkiasen:
### Fixed

- "Fixed" calendar not listening for changes after initial save

---

I couldn't find what the bug was, every time I tracked a location which, when modified, made it work, it would break down as I start to fix it or un-isolate parts. I'm not saying this is a hard bug, I'm saying I'm missing it, perhaps it's tunnel vision. 😅 

ANYWAYS I did know that it was due to the page transition to the `/edit` page messing with the state machine, so of course any good old refresh fixed it. So uh yeah now it refreshes to avoid the bug. It still keeps scroll position and all so it's just an extra flash of white but I'm open to leads for a more future-proof fix :/ 

Ticket: https://jira.publiq.be/browse/III-6110

### DIFF
--- a/playwright.config.ts
+++ b/playwright.config.ts
@@ -62,5 +62,8 @@ export default defineConfig({
   webServer: {
     command: 'yarn dev',
     port: 3000,
+    env: {
+      NEXT_PUBLIC_ENVIRONMENT: 'e2e',
+    },
   },
 });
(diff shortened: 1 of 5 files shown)


simon-debruijn:
I would fic the react query issue like this:

```tsx
{publicRuntimeConfig.environment === 'development' && (
          <ReactQueryDevtools initialIsOpen={false} />
 )}
```

And in the playwright config

```ts
webServer: {
    env: {
      NEXT_PUBLIC_ENVIRONMENT: 'e2e',
    },
    command: 'yarn dev',
    port: 3000,
  },
  ```

Anahkiasen:
@simon-debruijn That worked, thanks!

simon-debruijn (approved):
The fix itself doesn't look that great, but maybe you can make a jira ticket (and a todo comment) to later fix this bug more cleanly.

PR #907, opened 2024-06-05
Add timezone package to date-fns

Anahkiasen:
### Added

- [Added ...]

### Changed

- [Changed ...]

### Removed

- [Removed ...]

### Fixed

- [Fixed ...]

---

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,6 +43,7 @@
     "clipboard-copy": "^4.0.1",
     "crypto-browserify": "^3.12.0",
     "date-fns": "^2.16.1",
+    "date-fns-tz": "^2.0.1",
     "deep-object-diff": "^1.1.0",
     "draft-js": "^0.11.7",
     "draftjs-to-html": "^0.9.1",
--- a/src/hooks/api/events.ts
+++ b/src/hooks/api/events.ts
@@ -378,50 +378,6 @@ const useChangeNameMutation = (configuration = {}) =>
     ...configuration,
   });
 
-const changeCalendar = async ({
-  headers,
-  id,
-  calendarType,
-  timeSpans,
-  subEvent,
-  start,
-  end,
-  startDate,
-  endDate,
-  openingHours,
-  dayOfWeek,
-  opens,
-  closes,
-}) => {
-  return fetchFromApi({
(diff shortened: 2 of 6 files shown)


Anahkiasen:
This one was sometimes a headache to debug and fix, and I'm sure it's just as confusing to review, so to highlight the issue I've included two videos. One for the behavior from without the fix, one for the behavior with it.

The crux of the issue is that all Date instances in Javascript use the local timezone, and that to handle opening hours more easily we cast the plain time strings to Date instances, which then inherit that timezone. So when you set a time of 1AM in Tokyo, it actually sends 4PM on the day before to the API (since Tokyo is GMT+9 and we send UTC to the API, so 1AM - 9h). This means for the user filling the form the times are all correct but they aren't in the API, and as such aren't in the actual published event schedules.

Here in the "Before" video you see me set a time from Tokyo, then the wrong time being sent to the API. Then I refresh the page in Tokyo, the time is still shown correct to the user, but then viewing it from the standpoint of someone from Belgium, the time is suddenly incorrect.

https://github.com/user-attachments/assets/5415bd8f-c98a-4da2-9e25-007993be94d7

So to fix it we needed to ignore the user's local timezone. But Date instances don't actually have one, they only store the time internally and pick up on the timezone from the user. So you can't "switch timezone" besides by adding/removing the relevant amount of minutes. So that's what the fix does, before sending a Date to the API, we strip the offset of the user's local timezone and add back Belgium's, to "pretend" that the user is in Belgium. Of course since we also make Date instances off the datetimes received by the API, this has to be done the other way around on reception, otherwise the contributor will see the "corrected" time.

In the After video we see the Tokyo user filling the same times but now the API times correctly match Belgium. And now both the Tokyo and Belgium users see the same times when editing the offer, no matter what their local timezone is.

https://github.com/user-attachments/assets/3bf0f526-787e-43e0-8633-5a6702bb6c6c

simon-debruijn (changes requested):
Looks good, nice find 👍🏻 

Could you add a e2e test emulating the flow from the video?
https://playwright.dev/docs/emulation#locale--timezone

simon-debruijn (approved):

PR #915, opened 2024-09-04
Ensure the timezone offset of the date being checked is used instead of local

Anahkiasen:
### Fixed

- Ensure the timezone offset of the date being checked is used instead of local

---

Ticket: https://jira.publiq.be/browse/III-6277

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,7 +10,7 @@ jobs:
       - uses: actions/checkout@v3
       - uses: actions/setup-node@v3
         with:
-          node-version: 18
+          node-version: 16
 
       - name: Install dependencies
         run: yarn install --frozen-lockfile
@@ -24,7 +24,7 @@ jobs:
       - uses: actions/checkout@v3
       - uses: actions/setup-node@v3
         with:
-          node-version: 18
+          node-version: 16
 
       - name: Install dependencies
         run: yarn install --frozen-lockfile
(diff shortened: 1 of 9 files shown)


Anahkiasen:
Just marking as ready for review for E2E tests

Anahkiasen:
proof of It Works On My Machine™ 

Screenshot 2024-09-04 at 18 44 33

.github/workflows/ci.yml
```diff
@@ -10,7 +10,7 @@ jobs:
       - uses: actions/checkout@v3
       - uses: actions/setup-node@v3
         with:
-          node-version: 14.19
```
Anahkiasen, line 13: I've had to use a later version of Node to run the version of Playwright that has the Clock API, I understand we most likely don't want to diverge too much, but this felt like an okay compromise. In particular since later versions of Playwright also come with UI mode which is great for TDD, so that's the version I've had personally installed for a while, but tests wouldn't pass on it before, now they do, so could be a good occasion to upgrade

simon-debruijn: 16 is atm the version on jenkins for build, and also the version on the server for nodejs runtime

Anahkiasen: @simon-debruijn Sorry just to double check, are you saying I should use 16 here or that we should push for 18 on the actual app?

simon-debruijn (approved):
I'll approve already, hope your current approach works out for ci

PR #918, opened 2024-09-30
Sync ci node version with runtime

simon-debruijn:
### Changed

- Sync ci node version with runtime


simon-debruijn:
Ok the version used on the server is not compatible with the ci it seems.

Best look into updating to nodejs 18/20 soon

PR #921, opened 2024-10-02
III-6039-New dashboard design

vhande:
### Added
- xs calendar
- uit logo svg
- DashboardRow

### Changed
- Dashboard layout
- DynamicBarometerIcon
- Pagination

Screenshot 2024-10-09 at 15 22 53
Screenshot 2024-10-09 at 15 12 18

---

Ticket: https://jira.uitdatabank.be/browse/III-6039


public/assets/no-image-available.svg
brampauwelyn, line 1: Maybe you could just use the `UiT` logo with svg and use css to make the background grey?

src/i18n/de.json
```diff
@@ -893,8 +893,8 @@
     }
   },
   "pagination": {
-    "next": "Weiter",
-    "previous": "Zurück"
+    "next": "\u003E",
+    "previous": "\u003C"
```
brampauwelyn, line 897: Is there a reason why we use entity characters instead of an Icon? 
I think it would be more esthetic if we use an icon...

src/pages/dashboard/index.page.tsx
```diff
@@ -38,8 +42,11 @@ import type { Place } from '@/types/Place';
 import { Values } from '@/types/Values';
 import { WorkflowStatus } from '@/types/WorkflowStatus';
 import { Alert, AlertVariants } from '@/ui/Alert';
-import { Box, parseSpacing } from '@/ui/Box';
+import { Box } from '@/ui/Box';
 import { Dropdown, DropDownVariants } from '@/ui/Dropdown';
+import { Icon } from '@/ui/Icon';
+import { Icons } from '@/ui/Icon';
```
brampauwelyn, line 48: You can probably combine the imports?
```suggestion
import { Icon, Icons } from '@/ui/Icon';
```

src/pages/dashboard/index.page.tsx
```diff
@@ -745,12 +1001,12 @@ const getServerSideProps = getApplicationServerSideProps(
           req,
           queryClient,
           creator: user,
-          ...(key === 'events' && {
+          ...{
```
brampauwelyn, line 1004: This spread seems weird.
I think you can just remove it if it's not necessary to check the key...
```suggestion
```

src/pages/dashboard/index.page.tsx
```diff
@@ -745,12 +1001,12 @@ const getServerSideProps = getApplicationServerSideProps(
...
             sortOptions: {
               field: sortingField,
               order: sortingOrder,
             },
-          }),
+          },
```
brampauwelyn, line 1009:
```suggestion
```

src/pages/dashboard/index.page.tsx
```diff
@@ -166,54 +184,264 @@ const StatusIndicator = ({
...
-            {finishedAt}
-          </Text>
-        ) : (
-          actions.length > 0 && (
-            <Dropdown variant={DropDownVariants.SECONDARY} isSplit>
-              {actions}
-            </Dropdown>
-          )
+    <Inline spacing={5} {...getInlineProps(props)} flex={1}>
+      <Inline width="100" alignItems="center">
```
brampauwelyn, line 297: There's a lot of code on this page.
Maybe we could abstract the the offer row of the dashboard into it's own component?

src/i18n/de.json
simon-debruijn: Same remark. It also doesn't really make sense to have these in the translations if it is just `<` and `>`

src/ui/Pagination.tsx
```diff
@@ -98,12 +102,22 @@ const Pagination = ({
           z-index: ${getValue('zIndex')};
         }
 
-        .prev-btn {
-          margin-right: 0.8rem;
+        .prev-btn.disabled > span {
+        background-color: ${udbMainLightGrey};
```
simon-debruijn, line 106: Normally you shouldn't import colors directly.

If I'm not mistaken you can get to the color by doing

`background-color: ${({ theme }) => theme.colors.udbMainLightGrey};`

simon-debruijn (changes requested):

src/ui/Pagination.tsx
simon-debruijn:
```ts
const getColor = getValueFromTheme('colors');
`background-color: ${getColor('udbMainLightGrey')}`
```

src/pages/dashboard/DashboardRow.tsx
```diff
@@ -0,0 +1,302 @@
...
+            <Image
+              src={`/assets/uit-logo.svg`}
+              alt="No image available"
+              width={45}
+              height={45}
+            />
+          </Box>
+        )}
+      </Inline>
+      <PictureUploadModal
```
brampauwelyn, line 215: I think the picture upload should probably be on the index page.
Since now there's a modal for every row, which might be a bit overkill.

The modal should be on the index page and you trigger it from the child `DashboardRow`

brampauwelyn (changes requested):

simon-debruijn (approved):

brampauwelyn (approved):

PR #922, opened 2024-10-07
Added ability to leave street blank for organisers

Anahkiasen:
### Added

- Added ability to leave street blank for organisers

---

Ticket: https://jira.publiq.be/browse/III-6231


simon-debruijn (changes requested):
The e2e tests give a valid error

 ```
Error: locator.fill: Error: strict mode violation: getByLabel('Straat en nummer') resolved to 2 elements:
        1) <input type="text" name="address.streetAndNumber" id="organizer-address-streetAndNumber" class="Box__StyledBox-sc-ogeep9-0 dTTsTB form-control"/> aka getByLabel('Straat en nummer', { exact: *** })
        2) <input value="" type="checkbox" id="blank_address" name="blank_address" class="Box__StyledBox-sc-ogeep9-0 hiedRv custom-control-input"/> aka getByLabel('Ik wil mijn straat en nummer')
```

File: src/test/e2e/create-place.spec.ts [hunk 1]
```diff
@@ -51,9 +51,10 @@ test('create a place', async ({ baseURL, page }) => {
   await page
     .getByRole('option', { name: dummyPlace.address.municipality })
     .click();
-  await page.getByLabel('Straat en nummer').click();
+  await page.getByLabel('Straat en nummer').nth(0).click();
```
simon-debruijn, line 54: I slightly prefer the first method

simon-debruijn (approved):

src/test/e2e/create-place.spec.ts
Anahkiasen: What do you mean? 🤔

Anahkiasen: I think the issue is maybe placing an actual field inside a field's info slot, which means the label applies to both. perhaps, I can find a way to work around it so that it's better from a semantics/a11y standpoint

PR #924, opened 2024-10-11
III-6257-Sort the days of the opening hours

vhande:
### Added
- Sort the days of the opening hours

---

Ticket: https://jira.uitdatabank.be/browse/III-6257

### DIFF
--- a/src/pages/steps/CalendarStep/CalendarOpeninghoursModal.tsx
+++ b/src/pages/steps/CalendarStep/CalendarOpeninghoursModal.tsx
@@ -145,11 +145,15 @@ const CalendarOpeninghoursModal = ({
     idToChange: string,
   ) => {
     const checked = event.target.checked;
+    const daysOrder = Object.values(DaysOfWeek);
 
     openingHoursField.replace(
       openingHours.map((openingHour) => {
         if (openingHour.id === idToChange && checked) {
           openingHour.dayOfWeek.push(dayOfWeek);
+          openingHour.dayOfWeek.sort((a, b) => {
+            return daysOrder.indexOf(a) - daysOrder.indexOf(b);
+          });
         }
         if (
           openingHour.id === idToChange &&


src/pages/steps/CalendarStep/CalendarOpeninghoursModal.tsx
```diff
@@ -145,11 +145,15 @@ const CalendarOpeninghoursModal = ({
     idToChange: string,
   ) => {
     const checked = event.target.checked;
+    const daysOrder = Object.values(DaysOfWeek);
```
simon-debruijn, line 148: We cannot rely on the order of the values from an object.
It may work now, but it may not work at some point according to the js spec.

keys and values in an object don't maintain a certain order, because they use a hashmap datastructure behind the scenes.
Since the codebase only uses the values anyways, never specific days we can refactor the DaysOfWeek to this:

```ts
const DaysOfWeek = [
  'monday',
  'tuesday',
  'wednesday',
  'thursday',
  'friday',
  'saturday',
  'sunday',
] as const;

type DayOfWeek = (typeof DaysOfWeek)[number];

```

Arrays do maintain their order of elements

src/pages/steps/CalendarStep/CalendarOpeninghoursModal.tsx
```diff
@@ -145,11 +145,15 @@ const CalendarOpeninghoursModal = ({
...
 
     openingHoursField.replace(
       openingHours.map((openingHour) => {
         if (openingHour.id === idToChange && checked) {
           openingHour.dayOfWeek.push(dayOfWeek);
+          openingHour.dayOfWeek.sort((a, b) => {
+            return daysOrder.indexOf(a) - daysOrder.indexOf(b);
+          });
```
simon-debruijn, line 156: With the change above this becomes

```suggestion
          openingHour.dayOfWeek.sort((a, b) => {
            return DaysOfWeek.indexOf(a) - DaysOfWeek.indexOf(b);
          });
```

simon-debruijn (changes requested):

simon-debruijn (approved):

PR #938, opened 2024-10-21
Added ability to invite user as owner

Anahkiasen:
### Added

- Added ability to invite user as owner

---

Waiting for https://github.com/cultuurnet/udb3-backend/pull/1847 to wrap up

Ticket: https://jira.publiq.be/browse/III-6358


src/pages/organizers/[organizerId]/edit/index.page.tsx
```diff
@@ -18,6 +18,7 @@ export const getServerSideProps = getApplicationServerSideProps(
 
     return {
       props: {
+        organizer,
```
Comment (simon-debruijn) [inline src/pages/organizers/[organizerId]/edit/index.page.tsx:21]: Normally the line below should be enough, and then you use the same query hook to fetch the (cached already server fetched) organizer

simon-debruijn (dismissed):

src/pages/organizers/[organizerId]/edit/index.page.tsx
Comment (simon-debruijn) [reply src/pages/organizers/[organizerId]/edit/index.page.tsx:21]: Sorry I see you're still in Draft. but normally you shouldn't have to pass the organizer prop separately 😉

Comment (Anahkiasen) [reply src/pages/organizers/[organizerId]/edit/index.page.tsx]: Sorry yes was leftover from local tests but the query is enough indeed

src/pages/organizers/[organizerId]/ownerships/index.page.tsx
```diff
@@ -90,6 +97,25 @@ const Ownership = () => {
...
+    return approveOwnershipRequestMutation.mutateAsync({ ownershipId });
+  };
+
+  const requestOwnership = useRequestOwnershipMutation({
+    onError: async (error: FetchError<ErrorBody>) => {
+      if (error.body.status === 409) {
+        const ownershipId = error.body.detail.split('with id ')[1];
+        await approveRequestedOwnership(ownershipId);
+      } else {
+        setError('email', { message: error.body.detail || error.title });
```
Comment (Anahkiasen) [inline src/pages/organizers/[organizerId]/ownerships/index.page.tsx:113]: @simon-debruijn I did this originally to cascade the email errors back to the user UI, but ended up reverting it to a generic message but it allowed users to probe for existing emails in the system, but now it's sometimes cryptic why an email is rejected. Do you think we should show the actual errors?

simon-debruijn (approved):

simon-debruijn (approved):

PR #943, opened 2024-10-25
III-6359 - Delete ownership request

vhande:
### Added
- useDeleteOwnershipRequestMutation

Screenshot 2024-10-25 at 15 49 02

---

Ticket: https://jira.uitdatabank.be/browse/III-6359

### DIFF
--- a/src/ui/Button.tsx
+++ b/src/ui/Button.tsx
@@ -224,7 +224,7 @@ const customCSS = css`
       background-color: ${getValue('icon.hoverBackgroundColor')};
     }
 
-    &:focus {
+    &:active {
       background-color: ${getValue('icon.focusBackgroundColor')};
       animation: pulse 0.3s ease-out;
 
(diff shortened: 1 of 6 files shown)


src/ui/Button.tsx
```diff
@@ -224,7 +224,7 @@ const customCSS = css`
       background-color: ${getValue('icon.hoverBackgroundColor')};
     }
 
-    &:focus {
+    &:active {
```
brampauwelyn, line 227: Maybe you can add the active state instead of replacing the :focus state?
Unless the focus state is not necessary?

```suggestion
    &:active,
    &:focus {
```

brampauwelyn (approved):

src/ui/Button.tsx
vhande: I added the focus state by mistake; what I actually meant to add was the active state.

simon-debruijn (approved):

PR #944, opened 2024-10-29
Fix looping iframe url change

simon-debruijn:
### Fixed

- Fix looping bugs on iframe url change

---

Ticket: https://jira.publiq.be/browse/III-6325

### DIFF
--- a/src/middleware.api.ts
+++ b/src/middleware.api.ts
@@ -54,7 +54,9 @@ export const middleware = async (request: NextRequest) => {
     const isOwnershipEnabled =
       process.env.NEXT_PUBLIC_OWNERSHIP_ENABLED === 'true';
 
-    const redirectUrl = new URL(request.nextUrl);
+    const redirectUrl = new URL(request.url);
+    // remove the path variables from nextjs routing
+    redirectUrl.searchParams.delete('params');
 
     if (
       isOwnershipEnabled &&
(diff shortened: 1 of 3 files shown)


src/middleware.api.ts
```diff
@@ -45,14 +45,51 @@ export const middleware = async (request: NextRequest) => {
...
+    (request.nextUrl.pathname.startsWith('/organizer') &&
+      !request.nextUrl.pathname.endsWith('/ownerships')) ||
+    request.nextUrl.pathname.startsWith('/search');
 
-  if (isOwnershipPage && !process.env.NEXT_PUBLIC_OWNERSHIP_ENABLED) {
-    const url = new URL('/404', request.url);
-    return NextResponse.redirect(url);
+  if (isOwnershipPage) {
+    const isOwnershipEnabled =
+      process.env.NEXT_PUBLIC_OWNERSHIP_ENABLED === 'true';
```
Anahkiasen, line 55: I would maybe bypass this entire condition if this is not the case, I'd even argue leaving the extra search param in is harmless. Would allow you to really simplify the logic and merge conditions I think

Anahkiasen (approved):

src/middleware.api.ts
simon-debruijn: I'm going to merge it for now, so other PRs can use this logic

PR #945, opened 2024-10-29
III-6325 - Add tabs to search page

vhande:
### Added
- Search page
- useLegacyPath

Screenshot  (1)

---

Ticket: https://jira.uitdatabank.be/browse/III-6325

### DIFF
--- a/src/hooks/useLegacyPath.tsx
+++ b/src/hooks/useLegacyPath.tsx
@@ -1,0 +1,44 @@
+import getConfig from 'next/config';
+import { useRouter } from 'next/router';
+import { useMemo } from 'react';
+
+import { useCookiesWithOptions } from './useCookiesWithOptions';
+
+const useLegacyPath = () => {
+  const { cookies } = useCookiesWithOptions(['token', 'udb-language']);
+  const router = useRouter();
+  const { publicRuntimeConfig } = getConfig();
+  const prefixWhenNotEmpty = (value, prefix) =>
+    value ? `${prefix}${value}` : value;
+  const {
+    // eslint-disable-next-line no-unused-vars
+    query: { params = [], ...queryWithoutParams },
+    asPath,
+  } = router;
+
+  const useLegacyPath = useMemo(() => {
+    const path = new URL(`http://localhost${asPath}`).pathname;
+    const ownershipPaths =
+      router.asPath.startsWith('/organizer') &&
+      !router.asPath.endsWith('/ownerships');
+    const queryString = prefixWhenNotEmpty(
+      new URLSearchParams({
+        ...queryWithoutParams,
+        jwt: cookies.token,
+        lang: cookies['udb-language'],
+        ...(ownershipPaths &&
+          publicRuntimeConfig.ownershipEnabled === 'true' && {
+            ownership: 'true',
+          }),
+      }),
+      '?',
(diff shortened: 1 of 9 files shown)


src/pages/search/index.page.tsx
```diff
@@ -0,0 +1,131 @@
...
+                }
+
+                &.active {
+                  color: white;
+                }
+
+                &.active:hover {
+                  background-color: ${udbMainDarkBlue};
+                }
+              }
```
Anahkiasen, line 98: I'm surprised we need this much custom styling here, since we already have tabs on the Dashboard. Could we maybe it an actual variant to reuse in both places for consistency and brievty

src/pages/search/index.page.tsx
```diff
@@ -0,0 +1,131 @@
...
+              {tab === 'event-places' && (
+                <TabContent>
+                  <Stack flex={1}>
+                    {isClientSide && (
+                      <iframe
+                        height={
+                          iframeHeight
+                            ? `${iframeHeight}px`
+                            : `${window.innerHeight}`
+                        }
```
Anahkiasen, line 111: I'm still a little bit on the fence on this solution, I understand the need for it since contrary to other embedded pages this one is embedded within another. But could we perhaps use Grid here or a calc to avoid having to do these recalculations back and forth with the iframe?

Anahkiasen (changes requested):

src/pages/search/index.page.tsx
vhande: In other pages, we scroll within the iframe. However, here we want to scroll the entire page so that the title and tabs do not remain sticky. I haven't found a solution to enable scrolling of the whole page without knowing the exact height of the child component. Otherwise, the scrolling is limited to within the iframe.

src/hooks/useLegacyPath.tsx
```diff
@@ -0,0 +1,44 @@
...
+        ...(ownershipPaths &&
+          publicRuntimeConfig.ownershipEnabled === 'true' && {
+            ownership: 'true',
+          }),
+      }),
+      '?',
+    );
+
+    return `${publicRuntimeConfig.legacyAppUrl}${path}${queryString}`;
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
simon-debruijn, line 38: You normally now can merge in the latest changes from main. Passing `ownership` shouldn't be necessary anymore.
Also I would fix this dependency array like I did, since it is an easy fix to make eslint happy

src/ui/Tabs.tsx
```diff
@@ -6,19 +6,30 @@ import {
...
-type Props<T> = BoxProps & TabsProps & { activeBackgroundColor?: string };
+export const TabsCustomVariants = {
+  DEFAULT: 'default',
+  OUTLINED: 'outlined',
+} as const;
+
+type Props<T> = BoxProps &
+  TabsProps & {
+    activeBackgroundColor?: string;
+    customVariant?: Values<typeof TabsCustomVariants>;
```
simon-debruijn, line 25: Why not just a `variant` prop? And `TabVariant`

simon-debruijn (changes requested):

simon-debruijn (approved):

PR #946, opened 2024-10-30
III-6151 fix timezone issues latest

simon-debruijn:
### Added

- Add test for timezone

### Changed

- Install more recent date-fns and tz
- Convert timezone in changeOfferCalendar
- Convert timezone in addEvent
- Convert timezone in getEventById
- Convert start and endDate in addEvent

---

Ticket: https://jira.uitdatabank.be/browse/III-6151


src/test/e2e/events/create-calendar-single.spec.ts
```diff
@@ -44,3 +46,60 @@ test('create an event with calendarType single', async ({ baseURL, page }) => {
...
+
+    url = page.url();
+  });
+
+  await withTimezone(
+    { browser, timezoneId: 'Europe/Brussels' },
+    async (page) => {
+      await page.goto(url);
+      const startHour = page.getByLabel('Beginuur');
+      expect(await startHour.inputValue({ timeout: 10_000 })).toEqual('01:00');
```
brampauwelyn, line 102: It will always be 01:00 I guess?
Even if would change the timezone here to Tokyo? 

Or will it display the time with the Tokyo timezine?

brampauwelyn (approved):

simon-debruijn:
I'm going to close this PR. The feature seems to work, but I think the risk of introducing new bugs is larger than the demand.

PR #953, opened 2024-11-13
III-6393 - Display creator email

vhande:
### Added
- useGetOwnershipCreatorQuery

---

Ticket: https://jira.uitdatabank.be/browse/III-6393

### DIFF
--- a/src/pages/organizers/[organizerId]/ownerships/index.page.tsx
+++ b/src/pages/organizers/[organizerId]/ownerships/index.page.tsx
@@ -63,7 +63,8 @@ const Ownership = () => {
   const organizer: Organizer = getOrganizerByIdQuery?.data;
   const organizerName =
     organizer?.name?.[i18n.language] ??
-    organizer?.name?.[organizer.mainLanguage];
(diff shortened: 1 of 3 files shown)


src/pages/organizers/[organizerId]/ownerships/index.page.tsx
```diff
@@ -76,6 +81,9 @@ const Ownership = () => {
     [getOwnershipRequestsQuery.data],
   );
 
+  // @ts-expect-error
+  const creator = getOwnershipCreatorQuery.data;
```
Comment (simon-debruijn) [inline src/pages/organizers/[organizerId]/ownerships/index.page.tsx:85]:
```suggestion
  const creator: OwnershipCreator | undefined = getOwnershipCreatorQuery.data;
```
I think you can manually add the type

simon-debruijn (approved):

src/pages/organizers/[organizerId]/ownerships/index.page.tsx
Comment (vhande) [reply src/pages/organizers/[organizerId]/ownerships/index.page.tsx:88]: The problem is that the type does not include a data property.

PR #956, opened 2024-11-20
III-6365 - Add ownerships managements page

simon-debruijn:
### Added

- management page ownerships
- useOwnershipActions hook that we could reuse on organizer detail page in later PR

### Changed

- Refactor and reuse OwnershipsTable
### Removed

---

Ticket: https://jira.uitdatabank.be/browse/III-6365


brampauwelyn (approved):

src/hooks/api/organizers.ts
```diff
@@ -48,8 +44,8 @@ const useGetOrganizersByQueryQuery = (
       start: paginationOptions.start,
       limit: paginationOptions.limit,
     },
-    enabled: !!name,
     ...configuration,
+    enabled: !!name && configuration.enabled === true,
```
Anahkiasen, line 48: Should this maybe account for when `enabled` is not passed as part of `configuration`? I'm worried it might make `enabled` always false then 🤔

src/hooks/api/ownerships.ts
```diff
@@ -50,15 +64,37 @@ const useRequestOwnershipMutation = (configuration: UseQueryOptions = {}) =>
...
+}: {
+  headers: Headers;
+  itemId?: string;
+  ownerId?: string;
+  state?: OwnershipState;
+} & PaginationOptions) => {
+  const searchParams = new URLSearchParams();
+  if (paginationOptions) {
+    searchParams.set('limit', `${paginationOptions.limit}`);
+    searchParams.set('offset', `${paginationOptions.start}`);
```
Anahkiasen, line 84: Are the template strings to cast to string or something?

Anahkiasen (approved):
Nice reabstraction of the logic and nice cleanup on the types as well 👏  I think this should be the last major chunk? so that's nice that the hooks and stuff will get a bit more stable now

src/hooks/api/organizers.ts
simon-debruijn: This is probably more correct
`!!name && configuration.enabled !== false`

I think we make the same mistake in different parts of the codebase

src/hooks/api/ownerships.ts
simon-debruijn: They are needed because searchParams is a `Record<string, string>` and paginationOptions are passed as numbers

PR #966, opened 2025-01-08
Added ability to request suggested ownership

Anahkiasen:
### Added

- Added ability to request suggested ownership

---

Ticket: https://jira.publiq.be/browse/III-6362

### DIFF
--- a/src/pages/dashboard/index.page.tsx
+++ b/src/pages/dashboard/index.page.tsx
@@ -641,7 +641,7 @@ const Dashboard = (): any => {
   const createOfferUrl = CreateMap[tab];
   const { udbMainDarkBlue, textColor } = colors;
 
-  const [isVisible, setIsVisible] = useState(false);
+  const [isRequestModalVisible, setIsRequestModalVisible] = useState(false);
   const [currentOrganizer, setCurrentOrganizer] = useState<Organizer | null>(
     null,
   );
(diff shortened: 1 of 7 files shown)


src/pages/organizers/[organizerId]/preview/RequestOwnershipModal.tsx
```diff
@@ -0,0 +1,117 @@
...
+      </Modal>
+      {isSuccessAlertVisible && (
+        <Alert
+          variant={AlertVariants.SUCCESS}
+          marginBottom={4}
+          closable
+          fullWidth
+          onClose={() => setIsSuccessAlertVisible(false)}
+        >
+          <Trans
```
Comment (brampauwelyn) [inline src/pages/organizers/[organizerId]/preview/RequestOwnershipModal.tsx:96]: Is there a reason why we use "**Trans**" here in a lot of places?
Can we not use the regular syntax?
```
{t(organizers.ownerships.request.success, {organizerName)}
```

brampauwelyn (approved):

src/pages/dashboard/index.page.tsx
```diff
@@ -643,7 +641,36 @@ const Dashboard = (): any => {
   const createOfferUrl = CreateMap[tab];
   const { udbMainDarkBlue, textColor } = colors;
 
-  return [
+  const [isVisible, setIsVisible] = useState(false);
```
brampauwelyn, line 644: Maybe you can be more specific here?
```suggestion
  const [isRequestModalVisible, setIsRequestModalVisible] = useState(false);
```

src/pages/organizers/[organizerId]/preview/RequestOwnershipModal.tsx
Comment (Anahkiasen) [reply src/pages/organizers/[organizerId]/preview/RequestOwnershipModal.tsx:96]: Some do some more advanced stuff so sometimes you need it but I think this is just a leftover from a copy from somewhere else, I prefer the string version too, will switch it

PR #967, opened 2025-01-13
Fixed unusable locations showing in Recent Locations

Anahkiasen:
### Fixed

- Fixed unusable locations showing in Recent Locations

---

Ticket: https://jira.publiq.be/browse/III-6333

### DIFF
--- a/src/pages/steps/hooks/useRecentLocations.ts
+++ b/src/pages/steps/hooks/useRecentLocations.ts
@@ -3,6 +3,7 @@ import { uniqBy } from 'lodash';
 import { useGetOffersByCreatorQuery } from '@/hooks/api/offers';
 import { useGetUserQuery } from '@/hooks/api/user';
 import { Offer } from '@/types/Offer';
+import { WorkflowStatus } from '@/types/WorkflowStatus';
 
 const useRecentLocations = () => {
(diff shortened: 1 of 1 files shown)


src/pages/steps/hooks/useRecentLocations.ts
```diff
@@ -38,6 +39,8 @@ const useRecentLocations = () => {
     .filter(
       (location) =>
         location &&
+        location?.workflowStatus !== WorkflowStatus.REJECTED &&
+        location?.workflowStatus !== WorkflowStatus.DELETED &&
         location?.name?.nl !== 'Online' &&
```
simon-debruijn, line 44: Cant we drop the ? after location?

simon-debruijn (approved):

src/pages/steps/hooks/useRecentLocations.ts
Anahkiasen: No unfortunately some places don't have the attribute so it would return a false true and exclude them :/

Anahkiasen: It was to skip a preliminary filter() loop but I can make the check a different way maybe

PR #975, opened 2025-02-13
Reduce amount of console warnings

Anahkiasen:
### Changed

- Reduced amount of console warnings

---

Ticket: https://jira.publiq.be/browse/III-6514

### DIFF
--- a/src/ui/Notification.tsx
+++ b/src/ui/Notification.tsx
@@ -50,7 +50,6 @@ const Notification = ({ icon, header, body }: Props) => {
         paddingBottom={4}
         paddingRight={5}
         paddingLeft={4}
-        as={BootstrapToast.Body}
         alignItems="center"
         color={getValue('textColor.dark')}
       >
--- a/src/layouts/Sidebar.tsx
+++ b/src/layouts/Sidebar.tsx
@@ -610,9 +610,7 @@ const Sidebar = () => {
       key="sidebar"
       forwardedAs="nav"
       height="100%"
-      css={`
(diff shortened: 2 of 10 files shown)


Anahkiasen:
I'm giving up on the last one (FetchError Unauthorized) as mentioned in the call. It's supposed to be thrown at that moment, but not supposed to show in the logs since it's caught (in `getApplicationServerSideProps`). Despite being caught it still ends in the logs, so I've tried catching it elsewhere any number of ways, with or without sentry etc but it still ends up in the logs for some reason. I've sunk too much hours on the ticket already so letting that one go, I'm sure it's fixable and that it's something dumb but I can't find the right clue to find what :s

src/ui/Notification.tsx
```diff
@@ -50,7 +50,6 @@ const Notification = ({ icon, header, body }: Props) => {
         paddingBottom={4}
         paddingRight={5}
         paddingLeft={4}
-        as={BootstrapToast.Body}
```
simon-debruijn, line 53: Does it still render as a bootstrap toast now?

simon-debruijn (approved):

src/ui/Notification.tsx
Anahkiasen: Yes the body is actually just a forwarded `div` I think it's there more for semantic purposes

Anahkiasen: Of course I say that but now the tests fail on it -.- Will see if I can use the Body without it complaining about refs

PR #982, opened 2025-02-17
III-6516 - DIFF of remaining changes

simon-debruijn:
---

Ticket: https://jira.uitdatabank.be/browse/III-6516


simon-debruijn:
Doesn't show the diff I was hoping for

PR #992, opened 2025-02-21
III-6516 - Enable type checking js

simon-debruijn:
### Added

- [Added ...]

### Changed

- [Changed ...]

### Removed

- [Removed ...]

### Fixed

- [Fixed ...]

---

Ticket: https://jira.uitdatabank.be/browse/III-6516


src/pages/manage/productions/create/index.page.js
```diff
@@ -188,31 +193,45 @@ const Create = () => {
...
                     imageUrl={event?.image}
                     description={
+                      // @ts-expect-error TODO: Fix type error
                       event?.description?.[i18n.language] ??
+                      // @ts-expect-error TODO: Fix type error
                       event?.description?.[event.mainLanguage]
                     }
+                    // @ts-expect-error TODO: Fix type error
                     productionName={event?.production?.title}
+                    // @ts-expect-error TODO: Fix type error
```
Anahkiasen, line 234: This is maybe counterproductive as we're likely not going to fix these one by one and it hinders readability in the meantime, is there an equivalent to the file-level nocheck for expect-error maybe? :s

Anahkiasen (approved):
This naturally pollutes the codebase quite a lot but I'm electing to see it as a positive :D as it should prevent us letting it fester too long with the errors silenced

PR #995, opened 2025-03-03
Fixed CSS not being marked by CSS template string

Anahkiasen:
### Fixed

- Fixed CSS not being tagged with `css` template string

---

Before
Screenshot 2025-03-03 at 12 13 23

After
Screenshot 2025-03-03 at 15 13 48


simon-debruijn (approved):
Nice find 👍🏻

PR #998, opened 2025-03-06
Fixed deletion modal not showing on dashboard

Anahkiasen:
### Fixed

- Fixed deletion modal not showing on dashboard
- Fixed offers not disappearing on dashboard after deletion
- Fixed React warning about duplicate `divider` keys in offer dropdowns

---

Because `Page` only accepts children [that are Page slots](https://github.com/cultuurnet/udb3-frontend/blob/main/src/ui/Page.tsx#L28), it would discard the Modal in the `filter`ing being done of the raw children. Putting it in one of the slots makes it render in the DOM again.

I also weirdly had to switch to manually await for the queries invalidation, otherwise the cache gets busted but the refetch doesn't trigger. Not entirely sure why as to me a Promise returned from an `onSuccess` would be resolved but maybe not.

Ticket: https://jira.publiq.be/browse/III-6547

### DIFF
--- a/src/pages/dashboard/index.page.tsx
+++ b/src/pages/dashboard/index.page.tsx
@@ -608,9 +608,9 @@ const Dashboard = (): any => {
   });
 
   const deleteItemByIdMutation = useDeleteItemById({
-    onSuccess: () => {
+    onSuccess: async () => {
       setIsModalVisible(false);
-      return queryClient.invalidateQueries(tab);
+      await queryClient.invalidateQueries(tab);
     },
   });
 
--- a/src/pages/dashboard/index.page.tsx
+++ b/src/pages/dashboard/index.page.tsx
@@ -291,13 +291,13 @@ const OfferRow = ({ item: offer, onDelete, ...props }: OfferRowProps) => {
         <Dropdown.Item href={previewUrl} key="preview">
           {t('dashboard.actions.preview')}
         </Dropdown.Item>,
-        offerType === 'event' && <Dropdown.Divider key="divider" />,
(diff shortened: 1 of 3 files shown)


Anahkiasen:
Lemme maybe just add quick test for it now that I think about it should be easy enough

steftrenson (commented):
LGTM

src/pages/dashboard/index.page.tsx
```diff
@@ -800,32 +800,32 @@ const Dashboard = (): any => {
...
         {i18n.language === 'nl' && <NewsletterSignupForm />}
         <Footer />
+        <Modal
+          variant={ModalVariants.QUESTION}
+          visible={isModalVisible}
+          onConfirm={async () => {
+            deleteItemByIdMutation.mutate({
+              id: parseOfferId(toBeDeletedItem['@id']),
+            });
+          }}
```
simon-debruijn, line 810:
```suggestion
          onConfirm={() => {
            deleteItemByIdMutation.mutate({
              id: parseOfferId(toBeDeletedItem['@id']),
            });
          }}
```

simon-debruijn (approved):

src/pages/dashboard/index.page.tsx
Anahkiasen: @simon-debruijn If I remove it I get the same issue with query invalidation as described in the ticket, it's like it awaits for the mutation but not onSuccess. I've tried a few combination with mutate/mutateAsync etc but I can't really grasp what makes it work or not or what is wrong with our setup because yes it should work as is, but could be the whole query hydration logic? Not sure :s

PR #1001, opened 2025-03-24
III-6558-Add Cultuurkuur types

vhande:
### Added
- Cultuurkuur types

types

---

Ticket: https://jira.uitdatabank.be/browse/III-6558

### DIFF
--- a/src/pages/steps/EventTypeAndThemeStep.tsx
+++ b/src/pages/steps/EventTypeAndThemeStep.tsx
@@ -26,6 +26,7 @@ import { getValueFromTheme } from '@/ui/theme';
 
 import { UseEditArguments } from './hooks/useEditField';
 import { FormDataUnion, StepProps, StepsConfiguration } from './Steps';
+import { cultuurkuurTypes } from '@/constants/EventTypes';
 
 const DANCE_THEME_IDS = [
   '1.9.1.0.0', // Ballet en klassieke dans
@@ -84,20 +85,6 @@ const VARIOUS_THEME_IDS = [
   '1.65.0.0.0', // Voeding
   '1.25.0.0.0', // Wetenschap
   '1.44.0.0.0', // Zingeving, filosofie en religie
-];
-
-const CULTUURKUUR_TYPE_IDS = [
-  '0.3.1.0.1', // Cursus met open sessies
-  '0.7.0.0.0', // Begeleide uitstap of rondleiding
-  '0.55.0.0.0', // Theatervoorstelling
-  '0.54.0.0.0', // Dansvoorstelling
-  '0.50.21.0.0', // Spel of quiz
-  '0.50.6.0.0', // Film
-  '0.3.2.0.0', // Lezing of congres
-  '0.0.0.0.0', // Tentoonstelling
-  '0.50.4.0.0', // Concert
-  '0.17.0.0.0', // Fiets- of wandelroute
-  '0.5.0.0.0', // Festival
 ];
 
 const groupNameToThemeIds = {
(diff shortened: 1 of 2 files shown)


src/pages/steps/EventTypeAndThemeStep.tsx
```diff
@@ -210,6 +226,10 @@ const EventTypeAndThemeStep = ({
   const [, hash] = asPath.split('#');
   const eventTypeAndThemeContainer = useRef(null);
 
+  const isCultuurkuurEvent =
+    scope === OfferTypes.EVENTS &&
+    watch('audience.audienceType') === AudienceType.EDUCATION;
```
Anahkiasen, line 231: In `LocationStep` we have a [different way to check](https://github.com/cultuurnet/udb3-frontend/blob/main/src/pages/steps/LocationStep.tsx#L252) so I wonder if these should be made to match now that the audience info was moved before the location

src/pages/steps/EventTypeAndThemeStep.tsx
```diff
@@ -230,6 +250,11 @@ const EventTypeAndThemeStep = ({
     [getTypesByScopeQuery.data, sortByLocalizedName],
   );
 
+  const cultuurkuurTypes = useMemo(
+    () => types.filter((type) => CULTUURKUUR_TYPE_IDS.includes(type.id)),
+    [types],
+  );
```
Anahkiasen, line 256: Since we already memoize the building of the `types` I would maybe merge the two. Basically move the ternary in the JSX to the original useMemo and have it memoize the correct types on the first go

src/pages/steps/EventTypeAndThemeStep.tsx
```diff
@@ -289,32 +314,34 @@ const EventTypeAndThemeStep = ({
...
-                          &.btn {
-                            padding: 0.3rem 0.7rem;
-                            box-shadow: ${getGlobalValue('boxShadow.heavy')};
-                          }
-                        `}
-                      >
-                        {name[i18n.language]}
-                      </Button>
-                    ))}
+                    {(isCultuurkuurEvent ? cultuurkuurTypes : types).map(
```
simon-debruijn, line 317:
```suggestion
                    {types.map(
```
What @Anahkiasen said ⏫ You can move the filtering logic into the memoized types variable.
Since the mapping logic here remains the same

src/pages/steps/EventTypeAndThemeStep.tsx
```diff
@@ -85,6 +86,20 @@ const VARIOUS_THEME_IDS = [
...
+  '0.55.0.0.0', // Theatervoorstelling
+  '0.54.0.0.0', // Dansvoorstelling
+  '0.50.21.0.0', // Spel of quiz
+  '0.50.6.0.0', // Film
+  '0.3.2.0.0', // Lezing of congres
+  '0.0.0.0.0', // Tentoonstelling
+  '0.50.4.0.0', // Concert
+  '0.17.0.0.0', // Fiets- of wandelroute
+  '0.5.0.0.0', // Festival
+];
```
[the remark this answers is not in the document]
simon-debruijn: Yes and if possible reuse the ids from the existing EventTypes

Similar to 

```ts
const eventTypesWithNoThemes: string[] = [
  EventTypes['Eet- of drankfestijn'],
  EventTypes['Festiviteit'],
  EventTypes['Fiets- of wandelroute'],
  EventTypes['Party of fuif'],
  EventTypes['Spel of quiz'],
];
```

simon-debruijn (changes requested):

src/pages/steps/EventTypeAndThemeStep.tsx
simon-debruijn: Yeah that is specific to that step and a weird flow we have there.

A.t.m. you can choose a country or no country (aka cultuurkuur) that is represented as country undefined.
This will no longer be the case

simon-debruijn (approved):

PR #1012, opened 2025-04-09
III-6565-Add cultuurkuur organizer toggle

vhande:
### Added
- Cultuurkuur organizer toggle
- CultuurkuurStep

organizer

---

Ticket: https://jira.uitdatabank.be/browse/III-6565


src/pages/organizers/create/OrganizerForm.tsx
```diff
@@ -78,6 +88,9 @@ const OrganizerForm = () => {
   );
 
   const convertOrganizerToFormData = (organizer: Organizer) => {
+    const isCultuurkuurOrganizer = organizer?.hiddenLabels?.includes(
+      CULTUURKUUR_ORGANIZER_LABEL,
+    );
```
Anahkiasen, line 93: Would maybe make this a small helper since we need the same constant and check in `AdditionalInformationStep`

src/pages/organizers/create/OrganizerForm.tsx
```diff
@@ -143,8 +181,10 @@ const OrganizerForm = () => {
 
   const onSuccess = () => {
     upsertOrganizer({
-      onSuccess: async (organizerId) =>
-        await push(`/organizers/${organizerId}/edit`),
+      onSuccess: async (organizerId: string) => {
+        handleCultuurkuurLabelMutation(organizerId);
```
Anahkiasen, line 185: Think you would need to await this too otherwise the label change won't be reflected yet when the page changes

src/pages/organizers/create/OrganizerForm.tsx
```diff
@@ -113,8 +127,32 @@ const OrganizerForm = () => {
...
+      });
+    }
+
+    if (!isCultuurkuur) {
+      await removeLabelMutation.mutateAsync({
+        id: organizerId,
+        scope,
+        label: CULTUURKUUR_ORGANIZER_LABEL,
+      });
+    }
```
Anahkiasen, line 153: Would maybe merge both statements into one:

```ts
const mutation = isCultuurkuur ? addLabelMutation : removeLabelMutation;
await mutation.mutateAsync({
    id: organizerId,
    scope,
    label: CULTUURKUUR_ORGANIZER_LABEL,
});
```

src/pages/steps/AdditionalInformationStep/LabelsStep.tsx
```diff
@@ -56,16 +57,27 @@ function LabelsStep({
...
   useEffect(() => {
     onValidationChange(
       labels.length ? ValidationStatus.SUCCESS : ValidationStatus.NONE,
       field,
     );
   }, [field, labels, onValidationChange]);
 
+  useEffect(() => {
+    setLabels(getUniqueLabels(entity));
+  }, [entity]);
```
Anahkiasen, line 71: Not sure you would/will need this bit in theory with the `await` added in the other file

Anahkiasen (changes requested):

src/pages/steps/AdditionalInformationStep/LabelsStep.tsx
```diff
@@ -56,16 +57,27 @@ function LabelsStep({
...
+
   const isWriting = addLabelMutation.isLoading || removeLabelMutation.isLoading;
   const [isInvalid, setIsInvalid] = useState(false);
 
+  const handleInvalidateOrganizerQuery = async () => {
+    if (scope !== ScopeTypes.ORGANIZERS) return;
+    await queryClient.invalidateQueries('organizers');
+  };
```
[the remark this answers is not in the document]
vhande: This function helps to toggle the state on/off when the cultuurkuur_organizer label is added or removed. The states work correctly, but the toggle doesn't know about the updated state.

src/pages/steps/AdditionalInformationStep/LabelsStep.tsx
vhande: I added await, but even after the organizers query is invalidated, the state doesn't update automatically.

Anahkiasen: I see, that's weird but I've had the case in other tabs where invalidation don't work properly either so let's do it like that for now. This whole step is a bit weird since it's completely detached from the form logic so it doesn't help

src/utils/hasCultuurkuurOrganizerLabel.ts
```diff
@@ -0,0 +1,3 @@
+export const hasCultuurkuurOrganizerLabel = (labels: string[]) => {
+  return (labels ?? []).includes('cultuurkuur_organizer');
```
Anahkiasen, line 2: I would likely still use a constant but you can export it from this file, because right now it's still in both files consuming the helper (OrganizerForm and [AdditionalInformationStep](https://github.com/cultuurnet/udb3-frontend/blob/feature/III-6565/src/pages/steps/AdditionalInformationStep/AdditionalInformationStep.tsx#L34))

Anahkiasen (approved):
One last quick cleanup but otherwise looks good to me 👍

PR #1026, opened 2025-05-16
Fix price information

vhande:
### Fixed 
- Clicking the Free button now correctly creates a base tariff for new events without price info.
- Show a success toast after updating the price information by clicking the Free button.

---


src/pages/steps/AdditionalInformationStep/PriceInformation.tsx
```diff
@@ -258,14 +254,7 @@ const PriceInformation = ({
         ? isUitpasOrganizer(offer?.organizer)
         : false;
 
-    if (!hasRates) {
-      replace(
-        priceInfo.length
-          ? (priceInfo as FormData['rates'])
-          : defaultPriceInfoValues.rates,
-      );
-      reset({}, { keepValues: true });
```
Anahkiasen, line 267: Should this not still kick in at some point to set the default prices when the form is truly empty, for example on the initial creation? 🤔

vhande: I realized the form is already initialized with default price information when it's created. So if no price is provided, the defaults are already in place.

vhande: But I'd appreciate it if you could test it in different scenarios to see if it's working smoothly.

Anahkiasen: It was an edge-case when the offer is a UitPAS offers, because for those we constantly need to wait for the prices with taxes. But I tried to mess with it, switch back to a normal org etc, and didn't see weird behaviors, so perhaps this particular hook wasn't needed anymore? 🤔 But seemed okay to me

Anahkiasen (approved):

src/pages/steps/AdditionalInformationStep/PriceInformation.tsx
vhande: From what I understand from the documentation, to get an Uitpas price, you first need to enter a base tariff. The Uitpas price is then calculated based on that. So, if there’s no price information, the Uitpas price will never be calculated. And once a price is entered, the block becomes irrelevant. 🤔

PR #1055, opened 2025-06-13
III-6676-Fix blank street toggle

vhande:
### Fixed
- Blank street toggle

---

Ticket: https://jira.uitdatabank.be/browse/III-6676


src/pages/organizers/create/OrganizerForm.tsx
```diff
@@ -203,6 +203,7 @@ const OrganizerForm = () => {
           form={form}
           labels={organizerLabels}
           onChange={() => {
+            if (!form.getFieldState('nameAndUrl').isDirty) return;
```
grubolsch, line 206: I recommend avoiding single-line if statements and always using curly braces. One-liners make it easy to overlook critical control flow like return and are prone to errors if someone later adds more logic, such as a logging statement, inside the if block.

grubolsch (approved):
Approved, but I do have a style remark.

grubolsch:
You can also close https://jira.publiq.be/browse/III-6680 with this, because I also noticed this bug yesterday.

PR #1059, opened 2025-06-17
III-6362-Fix ownership request for suggested organizers

vhande:
### Fixed
- ownership request for suggested organizers

---

Ticket: https://jira.uitdatabank.be/browse/III-6362

### DIFF
--- a/src/pages/organizers/[organizerId]/preview/RequestOwnershipModal.tsx
+++ b/src/pages/organizers/[organizerId]/preview/RequestOwnershipModal.tsx
@@ -32,9 +32,9 @@ const RequestOwnershipModal = ({
   const { t, i18n } = useTranslation();
   const queryClient = useQueryClient();
   const router = useRouter();
-  const organizerIdFromQuery = router.query.organizerId as string;
-  const organizerIdFromOffer = parseOfferId(organizer?.['@id'] ?? '');
-  const organizerId = organizerIdFromQuery ?? organizerIdFromOffer;
+  const organizerId =
+    (router.query.organizerId as string) ??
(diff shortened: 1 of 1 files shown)


src/pages/organizers/[organizerId]/preview/RequestOwnershipModal.tsx
```diff
@@ -31,7 +32,9 @@ const RequestOwnershipModal = ({
   const { t, i18n } = useTranslation();
   const queryClient = useQueryClient();
   const router = useRouter();
-  const organizerId = router.query.organizerId as string;
+  const organizerIdFromQuery = router.query.organizerId as string;
+  const organizerIdFromOffer = parseOfferId(organizer?.['@id'] ?? '');
+  const organizerId = organizerIdFromQuery ?? organizerIdFromOffer;
```
Comment (grubolsch) [inline src/pages/organizers/[organizerId]/preview/RequestOwnershipModal.tsx:37]: I would inline organizerIdFromQuery and organizerIdFromOffer directly into the organizerId assignment. Since they’re only used once and not reused later, creating separate variables just adds extra lines without improving clarity, because other developers don't know if this variable will be re-used later ot not.

grubolsch (changes requested):

src/pages/organizers/[organizerId]/preview/RequestOwnershipModal.tsx
Comment (grubolsch) [reply src/pages/organizers/[organizerId]/preview/RequestOwnershipModal.tsx:37]:
```suggestion
const organizerId = (router.query.organizerId as string) ?? parseOfferId(organizer?.['@id'] ?? '');
```

grubolsch (approved):

PR #1092, opened 2025-09-24
III-6378 - Show alert <

brampauwelyn:
### Added
- [copy for Brussels alert](https://github.com/cultuurnet/udb3-frontend/commit/9df9192e96ebe0e156cc5633ab62a3c47c543bf7)
- [Show alert when a brussels city has been selected](https://github.com/cultuurnet/udb3-frontend/commit/b4f7b198052bd65d8dab16467e7290084df7ea34)

### Removed
- [some unused imports](https://github.com/cultuurnet/udb3-frontend/commit/af4ef1901e6bbff1989b41c855644c6961b0e932)

---
Ticket: https://jira.uitdatabank.be/browse/III-6378

### DIFF
--- a/src/pages/steps/LocationStep.tsx
+++ b/src/pages/steps/LocationStep.tsx
@@ -511,6 +511,11 @@ const LocationStep = ({
             field?.value as OfferFormData['location'];
 
           const isDefaultZip = municipality?.zip === '0000';
+          const isBrusselsZip =
+            municipality?.zip &&
+            Number(municipality.zip) >= 1000 &&
+            Number(municipality.zip) <= 1210;
+
           const isPhysicalLocation = !!country && !isOnline && !isDefaultZip;
 
           const onFieldChange = (updatedValue) => {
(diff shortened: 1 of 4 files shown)


src/pages/steps/LocationStep.tsx
```diff
@@ -512,6 +511,11 @@ const LocationStep = ({
             field?.value as OfferFormData['location'];
 
           const isDefaultZip = municipality?.zip === '0000';
+          const isBrusselsZip =
+            municipality?.zip &&
+            Number(municipality.zip) >= 1000 &&
+            Number(municipality.zip) <= 1210;
```
bertramakers, line 517: I'm a bit late to review, but I wonder if we should also check the selected country here to be safe? I see that right now we also support The Netherlands and Germany, and unless I'm mistaken their zip codes should never be in this range. So technically it's not necessary, but it might be safe in case we add more countries at some point and forget about this check? Unless it's a lot of trouble.

brampauwelyn: Good point, I would be safer. 
I can make a follow-up PR

src/pages/steps/LocationStep.tsx
```diff
@@ -551,6 +555,17 @@ const LocationStep = ({
                         : t('create.location.recent_locations.pick')}
                     </Text>
                   )}
+                  {isBrusselsZip && (
+                    <Alert variant="primary">
+                      {t('create.location.is_brussels_alert.message')}{' '}
+                      <a
+                        href="https://helpdesk.publiq.be/hc/nl/articles/360008702979-Hoe-voeg-ik-Brusselse-activiteiten-in"
+                        target="_blank"
```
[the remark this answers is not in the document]
bertramakers: Fun fact @brampauwelyn @vhande , this is not an issue anymore in modern browsers: https://owasp.org/www-community/attacks/Reverse_Tabnabbing
But it also doesn't hurt to add it so no need to remove it

PR #1093, opened 2025-09-24
Fix i18n hydration mismatch

vhande:
### Fixed
- i18n hydration mismatch

---


brampauwelyn (commented):
Looks good to me but copilot has some legit remarks

PR #1097, opened 2025-10-02
III-6349-Add permission check on event edit page

vhande:
### Added
-  getEventPermissions
- const EventPermissionTypes
- unauthorized.page.tsx

unauth

---

Ticket: https://jira.publiq.be/browse/III-6349

### DIFF
--- a/src/constants/SortOptions.ts
+++ b/src/constants/SortOptions.ts
@@ -1,0 +1,17 @@
+import { Values } from '@/types/Values';
+
+const SortOrder = {
+  ASC: 'asc',
+  DESC: 'desc',
+} as const;
+
+const SortField = {
+  AVAILABLETO: 'availableTo',
+  CREATED: 'created',
+  COMPLETENESS: 'completeness',
+} as const;
+
+type SortOrderType = Values<typeof SortOrder>;
+
+export { SortField, SortOrder };
+export type { SortOrderType };
--- a/src/hooks/api/ownerships.ts
+++ b/src/hooks/api/ownerships.ts
@@ -9,6 +9,7 @@ import {
   prefetchAuthenticatedQuery,
   queryOptions,
   ServerSideQueryOptions,
+  SortOptions,
   useAuthenticatedMutation,
   useAuthenticatedQuery,
 } from './authenticated-query';
@@ -29,6 +30,7 @@ export type OwnershipRequest = {
   ownerEmail: string;
   requesterId: string;
   state: OwnershipState;
+  created: string;
 } & ModificationInfo;
 
 export type OwnershipCreator = {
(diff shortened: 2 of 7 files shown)


src/constants/PermissionTypes.ts
```diff
@@ -11,4 +11,10 @@ const PermissionTypes = {
   FILMS_AANMAKEN: 'FILMS_AANMAKEN',
 } as const;
 
-export { PermissionTypes };
+const EventPermissionTypes = {
+  AANBOD_BEWERKEN: 'Aanbod bewerken',
+  AANBOD_MODEREREN: 'Aanbod modereren',
+  AANBOD_VERWIJDEREN: 'Aanbod verwijderen',
+} as const;
```
brampauwelyn, line 18: Is there a reason why you made a new const?
Since you could probably use the `PermissionTypes` const?

src/pages/events/[eventId]/edit/index.page.tsx
```diff
@@ -13,11 +16,35 @@ export const getServerSideProps = getApplicationServerSideProps(
...
+        queryClient,
+      }),
+      prefetchGetEventPermissionsQuery({
+        req,
+        queryClient,
+        eventId: eventId,
+      }),
+    ]);
+
+    const permissions = queryClient.getQueryData([
```
Comment (brampauwelyn) [inline src/pages/events/[eventId]/edit/index.page.tsx:32]: Can you use desctructuring here?
```suggestion
    const { permissions } = queryClient.getQueryData([
```

src/pages/events/[eventId]/edit/index.page.tsx
```diff
@@ -13,11 +16,35 @@ export const getServerSideProps = getApplicationServerSideProps(
...
+      'event-permissions',
+      { eventId },
+    ]);
+
+    if (
+      permissions?.permissions?.length === 0 ||
+      !permissions.permissions.includes(EventPermissionTypes.AANBOD_BEWERKEN)
```
Comment (brampauwelyn) [inline src/pages/events/[eventId]/edit/index.page.tsx:39]: If you use desctructuring like suggested above, you can simplify this:
```suggestion
     permissions?.length === 0 ||
      !permissions.includes(EventPermissionTypes.AANBOD_BEWERKEN)
```

src/constants/PermissionTypes.ts
vhande: I’m not sure if event permissions are a separate set from user permissions. If they’re actually the same, I’d create a function to handle the formatting (since user permissions are in all caps with underscores, while event permissions use only the first letter capitalized with spaces).

src/pages/events/[eventId]/edit/index.page.tsx
```diff
@@ -13,11 +14,35 @@ export const getServerSideProps = getApplicationServerSideProps(
   async ({ req, query, queryClient, cookies }) => {
     const { eventId } = query;
 
-    await prefetchGetEventByIdQuery({
-      id: eventId,
-      req,
-      queryClient,
-    });
+    await Promise.all([
```
Comment (brampauwelyn) [inline src/pages/events/[eventId]/edit/index.page.tsx:17]: I wonder if we need an Promise.all 
We don't need to fetch the event if the user hasn't permission? 
So we can probably do the redirect earlier, it will be a bit more performant I think. 

So you only do the fetch of the event if the user has the correct permissions.

PR #1197, opened 2026-04-09
III-7118-Add permission check for offer history prefetch and query

vhande:
### Added
- Permission check for offer history prefetch and query

---

Ticket: https://jira.uitdatabank.be/browse/III-7118

### DIFF
--- a/src/pages/events/[eventId]/Preview.tsx
+++ b/src/pages/events/[eventId]/Preview.tsx
@@ -118,9 +118,17 @@ const Preview = () => {
     mainLanguage,
   );
 
+  const isGodUser = userPermissions?.includes(
+    PermissionTypes.GEBRUIKERS_BEHEREN,
+  );
+  const canSeeHistory = userPermissions?.includes(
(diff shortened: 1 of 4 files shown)


brampauwelyn (approved):

bertramakers:
@brampauwelyn @vhande it's good to check the permissions of the user pro-actively 👍 But I think there is also a larger issue that we handle 403 responses as "Unauthorized" in various places, like here: https://github.com/cultuurnet/udb3-frontend/blob/eb57bfa531723a70d5bde93fd461fd968ca132df/src/hooks/api/authenticated-query.ts#L119

403 is "Forbidden", which means the user is authenticated (logged in), but they don't have the right permission to perform a specific action (like reading the event history).

I think this mix stems from the past when the backend didn't always send the right status code, sometimes mixing 401 and 403. However we cleaned that up some years ago.

So I would suggest to not log out the user automatically on a 403 response, because a 403 actually indicates that they are in fact logged in but are just not allowed to do a specific action. If they were not logged in the backend would not be able to do the permission check and it would return a 401 Unauthorized instead.

PR #1204, opened 2026-04-29
III-7146-Add holidays variant to DatePeriodPicker

vhande:
### Added
- DatePeriodPicker holidays variant
- SwitchButton 

### Changed
- udbMainDarkBlue to improve accessibility

calendar holiday

---

Ticket: https://jira.uitdatabank.be/browse/III-7146


src/ui/SwitchButton.tsx
```diff
@@ -0,0 +1,79 @@
+import { Box } from './Box';
```
brampauwelyn: Is there a reason why a new component has been added?
The toggle already exists in our UI library.
See `RadioButton`, you can pass it a type: `RadioButtonTypes.SWITCH`.

src/pages/steps/AdditionalInformationStep/BookingInfoStep.tsx
```diff
@@ -252,6 +259,11 @@ const ReservationPeriod = ({
             </Text>
 
             <DatePeriodPicker
+              variant={
+                isBoaEnabled
+                  ? DatePeriodPickerVariants.HOLIDAYS
+                  : DatePeriodPickerVariants.DEFAULT
+              }
```
brampauwelyn: Would it be an idea, to instead of making it a variant, just pass a prop: `showHolidaysToggle` ? If `showHolidaysToggle` is activated the toggle is present? What do you think?

src/ui/SwitchButton.tsx
vhande: RadioButtonTypes.SWITCH renders a form input, but since the calendar holidays toggle isn’t used in a form context, I added a separate switch button instead.

src/pages/steps/AdditionalInformationStep/BookingInfoStep.tsx
vhande: I implemented it as a variant to keep Storybook stories clean and separated. Another variant will also be added after this one.

bertramakers: I was wondering the same thing earlier today :)

I don't think you necessarily need different variants to keep stories clean and separated. The StatusIndicator component also has nice overview of various stories without a "variant":

- https://main--5f69c0eb52e6b60022561d99.chromatic.com/?path=/story/components-statusindicator--published
- https://github.com/cultuurnet/udb3-frontend/blob/a2645e634cd8f2b631fd89568966f11188ae6833/src/ui/StatusIndicator.stories.tsx

What would be the other variant? It would be helpful to know to determine whether it makes sense to have different variants, or instead several props that are not necessarily mutually exclusive but instead allow us to compose various configurations depending on use cases.

src/ui/SwitchButton.tsx
bertramakers: I'd also reuse the existing one to avoid we end up having different visual styles throughout the app. What would be the impact of having an `input` outside of a form?

The [W3C validator](https://validator.w3.org/nu/#textarea) doesn't complain when I validate a snippet like this:
```
<!DOCTYPE html>
<html lang="en">
  <head>
    <title>Test</title>
  </head>
  <body>
    <div>
      <input type="radio">
    </div>
  </body>
</html>
```

src/ui/DatePeriodPicker.tsx
```diff
@@ -126,14 +127,19 @@ const DatePeriodPicker = ({
...
+        type={RadioButtonTypes.SWITCH}
+        label={
+          <Text fontWeight={isHighlighted ? 'bold' : 'normal'}>
+            {t('date_period_picker.show_holidays')}
+          </Text>
+        }
+        checked={isHighlighted}
+        onChange={(e) => handleToggleHolidays(e.target.checked)}
+        disabled={disabled}
+        color={colors.udbMainPositiveGreen}
```
bertramakers, line 294: Thanks for changing this to use RadioButtonWithLabel as suggested! 👍 

Just to double check, is it important to have green switch button here instead of the default blue? Do you know what the reasoning is for making it green in this specific case? Or was it maybe green in the prototypes/wireframes but just because the default color in the prototype tool is green or something like that?

Or if it's an conscious choice to make it green now, might it make more sense to change the default so it's always green throughout the app?

I ask because our UDB UI is not always very consistent in terms of feeling at the moment. And I think that having the same component in various colors without a specific reason can contribute to this feeling of inconsistency. Of course there are good reasons sometimes like with a button or info box, depending on the context. But I don't see the reason here yet for this switch to be green instead of the usual blue.

It's an open question to be clear, curious to know if you know the reason or what your thoughts are @vhande, especially now with your knowledge about UX and UI design 😉

src/ui/DatePeriodPicker.tsx
```diff
@@ -1,35 +1,147 @@
...
+      ({ startDate, endDate }) =>
+        startDate <= lastDayOfViewedMonth && endDate >= firstDayOfViewedMonth,
+    )
+    .map(({ startDate, endDate, name }) => {
+      if (isSameDay(startDate, endDate))
+        return `${format(startDate, 'd MMMM', { locale })}: ${name}`;
+      if (isSameMonth(startDate, endDate))
+        return `${startDate.getDate()}-${endDate.getDate()} ${format(startDate, 'MMMM', { locale })}: ${name}`;
+      return `${format(startDate, 'd MMMM', { locale })} - ${format(endDate, 'd MMMM', { locale })}: ${name}`;
+    });
```
brampauwelyn: It's not really clear. In git it looks like there are 2 return statements inside the same if.
Normally we use curly braces after our `if` statements.
Maybe you can also add some newlines between the if's so it's more readable.

brampauwelyn (approved):

PR #1225, opened 2026-05-22
III-7198-Add new BOA-info on preview form

vhande:
### Added
- publication date
- Accessibility preview
- OpeningHoursSummary to the preview pages

### Changed
- styling BookingInfoPreview
- styling PriceInfo

preview 1

preview 2

---

Ticket: https://jira.uitdatabank.be/browse/III-7198


vhande (commented):
@claude can you review?

src/hooks/usePublicationStatus.ts
```diff
@@ -22,7 +22,10 @@ export type OfferPublicationStatus = PublicationStatusInfo & {
   isFinished: boolean;
 };
 
-export const usePublicationStatus = (offer: Offer): OfferPublicationStatus => {
+export const usePublicationStatus = (
+  offer: Offer,
+  showDate = true,
```
brampauwelyn: using a boolean with the default `true` feels a bit like a bad practice.

I would omit the showDate.

And just export two variants at the bottom 
e.g.

```typescript
return { color, label, labelWithDate, isExternalCreator, isFinished };
```
Then the dashboard uses label, the previews use labelWithDate. No flag, no default — and each call site is self-documenting.


```suggestion
```

src/hooks/usePublicationStatus.ts
```diff
@@ -50,17 +53,27 @@ export const usePublicationStatus = (offer: Offer): OfferPublicationStatus => {
 
     const color = PublicationStatusToColor[publicationStatus];
 
+    const needsDate =
```
brampauwelyn, line 54: What is the `needsDate` variable ? I find it a bit a weird name ...

src/pages/preview/PreviewTable.tsx
```diff
@@ -0,0 +1,33 @@
+import type { ReactNode } from 'react';
+
+import { colors } from '@/ui/theme';
+
+type Props = {
+  children: ReactNode;
+};
+
+const PreviewTable = ({ children }: Props) => (
+  <table
```
brampauwelyn: Normally we use the `<Table>` component. 
Would it be possible to use that component?

brampauwelyn: Maybe a new `<PreviewTable>` component is not necessary?

src/pages/preview/AgePreview.tsx
```diff
@@ -35,7 +43,19 @@ const AgePreview = ({ typicalAgeRange }: Props) => {
     ? AgeRanges[ageRangeLabelKey].label
     : formatCustomAgeRange(typicalAgeRange);
 
-  return <Text>{t('preview.ages', { ages: ageText })}</Text>;
+  const ChildrenOnlyLabel = () => (
```
brampauwelyn: ChildrenOnlyLabel is a React component (capitalized, returns JSX) declared inside the body of AgePreview. Every time AgePreview re-renders, JavaScript creates a new function for ChildrenOnlyLabel. From React's perspective that's a completely new component type each render — same shape, different identity. The reconciler can't tell that "this <ChildrenOnlyLabel/> is the same thing as last render's <ChildrenOnlyLabel/>", so it unmounts the old subtree and remounts a new one every time the parent renders.

Consequences:

Wasteful — extra DOM work, throws away any internal state and effects on every parent re-render. (Not painful here because the body has no state, but the pattern is brittle.)

**Solution**
-->  hoist ChildrenOnlyLabel to module scope

brampauwelyn (commented):
Maybe extend the existing preview e2e tests with the new functionality ?

brampauwelyn (approved):

brampauwelyn (approved):

PR #1239, opened 2026-06-04
III-7169 migrate Badge component to shadcn

danmkt00:
### Added
- `@radix-ui/react-slot` dependency pulled in by shadcn Badge (enables the `asChild` prop pattern)
- `BadgeLegacy` component preserving the original Bootstrap implementation — switchable via `SHADCN_MIGRATION` feature flag
- Storybook toolbar toggle to switch between Legacy and shadcn Badge for visual comparison — defaults to shadcn so Lost Pixel captures the new design

### Changed
- Migrated Badge component from Bootstrap + styled-components to shadcn Badge — not pixel-perfect intentionally, standardizing toward shadcn conventions
- Danger and Info badge colors now use UDB design tokens (`udb-danger`, `udb-main-blue`) instead of Bootstrap defaults
- Badge is now `rounded-full` to match the shadcn standard — `pill` prop kept for legacy compatibility only, can be removed after migration
- `Badge` now uses `SHADCN_MIGRATION` feature flag to toggle between `BadgeLegacy` and the new shadcn version

### Removed
- Bootstrap Badge dependency, styled-components CSS-in-JS, and BoxProps from Badge component
- Dead CSS from QuestionCircleIcon — `aspect-ratio` and `background-color`
  overrides in the `css` prop were never applied in legacy since BadgeLegacy
  forwards props through `getBoxProps` which filters out non-Box props

### Note
- Storybook defaults to the shadcn version — Lost Pixel baselines capture the new design

### Usage
- `Sidebar.tsx` (L271, L491) — INFO variant, shows numeric counts (announcements, events to moderate)
- `OfferPreviewSidebar.tsx` (L111) — SECONDARY variant, shows a language code (e.g. `nl`, `fr`) — reachable via `/events/<id>/preview` or `/places/<id>/preview`
- `NewFeatureTooltip.tsx` (L22) — SECONDARY variant, pill badge used as a question mark icon trigger

---

Ticket: https://jira.uitdatabank.be/browse/III-7169


brampauwelyn:
@danmkt00 can you maybe add a before and after screenshot of the badge component?
Or regenerate the lost pixel baselines? Or are the differences not noticeable?

danmkt00:
> @danmkt00 can you maybe add a before and after screenshot of the badge component? Or regenerate the lost pixel baselines? Or are the differences not noticeable?

@brampauwelyn From what I can see, the differences are actually noticeable, especially in colors (since we moved away from Bootstrap defaults toward our own palette) and also in some sizing. I’m not entirely sure yet whether we want to strictly match sizes or lean more into shadcn-style conventions where slight layout differences are acceptable.

It might also be worth reconfiguring Lost Pixel again if possible, so it behaves more like a proper review tool for visual diffs, making screenshots and differences clearly visible in PRs, rather than effectively “accepting” everything by default, even when there are changes. I’m not entirely sure if that’s actually possible though, since in another project it was configured to fail on differences.

src/ui/Badge.tsx
```diff
@@ -25,20 +34,25 @@ const Badge = ({
...
+  const [isShadcnMigrationEnabled] = useFeatureFlag(
+    FeatureFlags.SHADCN_MIGRATION,
+  );
+
+  return isShadcnMigrationEnabled ? (
+    <ShadcnBadge
+      variant={shadcnVariant[variant]}
+      className={cn(
+        'tw:self-center',
+        pill ? 'tw:rounded-full' : undefined,
```
brampauwelyn: Since we want to stay as close as possible to shadcn.
I think we can remove the pill variant for now.
Regenerate the baselines and maybe verify the UI how it looks...

danmkt00: Should we delete pill? With it enabled it looks closer to the shadcn standard. By default shadcn Badge uses tw:rounded-md (driven by our --radius: 8px token), which keeps our design consistent with the rest of the shadcn components. If we ever need a fully rounded badge, having pill as an opt-in prop gives us that flexibility without breaking the default.

danmkt00: Going with rounded-full as the standard since shadcn docs use it by default. Kept pill prop for legacy comparison - can be removed after migration.

src/ui/Badge.tsx
```diff
@@ -1,21 +1,30 @@
...
   SECONDARY: 'secondary',
   INFO: 'info',
 } as const;
 
-type Props = BoxProps & {
+type Props = {
+  children: React.ReactNode;
   variant?: Values<typeof BadgeVariants>;
-  pill?: boolean;
+  pill?: boolean; // legacy only — shadcn Badge is rounded-full by default, remove after migration
```
brampauwelyn:
```suggestion
  pill?: boolean; // TODO legacy only — shadcn Badge is rounded-full by default, remove after migration
```

brampauwelyn:
Like discussed, maybe check why the cultuurkuur tooltip badge looks like a circle.
Normally the shadcn badge is more like a "pill" ...

Screenshot 2026-06-09 at 13 23 37

danmkt00:
> Like discussed, maybe check why the cultuurkuur tooltip badge looks like a circle. Normally the shadcn badge is more like a "pill" ...
> 

@brampauwelyn done✅

brampauwelyn (approved):

PR #1249, opened 2026-06-10
III-7227-Add dates only if they fall within the event date range

vhande:
### Changed
- Add dates only if they fall within the event date range

error

---

Ticket: https://jira.uitdatabank.be/browse/III-7227

### DIFF
--- a/src/pages/steps/CalendarStep/DeviatingPeriod.tsx
+++ b/src/pages/steps/CalendarStep/DeviatingPeriod.tsx
@@ -1,4 +1,4 @@
-import { format, isSameDay } from 'date-fns';
+import { format, isSameDay, startOfDay } from 'date-fns';
 import uniqueId from 'lodash/uniqueId';
 import { useState } from 'react';
 import { useTranslation } from 'react-i18next';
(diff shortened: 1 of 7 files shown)


src/pages/steps/CalendarStep/DeviatingPeriod.tsx
```diff
@@ -207,16 +208,32 @@ const DeviatingPeriod = ({
...
+                  (!eventStartDate ||
+                    startOfDay(p.startDate) >= startOfDay(eventStartDate)) &&
+                  (!eventEndDate ||
+                    startOfDay(p.endDate) <= startOfDay(eventEndDate)),
+              );
+              if (quickLinksWithinEventRange.length === 0) {
+                setQuickLinkRangeError(true);
+                return;
+              }
+              setQuickLinkRangeError(false);
```
brampauwelyn: The logic is the same as in the `ClosingPeriod`.
Probably be interesting to abstract it somewhere so you could use it in 2 places and avoid duplicate code.

If you need to modify the logic you only need to do it in 1 place :)

brampauwelyn (approved):

PR #1252, opened 2026-06-10
III-7173 migrate Paragraph component to Tailwind

danmkt00:
### Added
- ParagraphLegacy.tsx — original implementation

### Changed
- Paragraph.tsx — feature flag toggle between Tailwind and legacy; introduces
  self-contained ParagraphVariants with proper typing; considered adding size
  variants but decided className is cleaner for one-off size overrides
- ButtonCard.tsx, Toast.tsx — replaced semantically incorrect Paragraph usage
  with Box as both were using it as a layout container
- PlaceAddModal.tsx — moved marginBottom to Box wrapper, switched to ParagraphVariants

Ticket: https://jira.uitdatabank.be/browse/III-7173


brampauwelyn (commented):
Maybe also generate the lost pixel baselines?

danmkt00:
> Maybe also generate the lost pixel baselines?

since it passed Visual Regresion Tests, i think there are no changes in the baseline

src/pages/PlaceAddModal.tsx
```diff
@@ -233,9 +233,11 @@ const PlaceAddModal = ({
         </Inline>
         <Stack>
           <Text fontWeight="bold">{t('location.add_modal.labels.type')}</Text>
-          <Paragraph marginBottom={3} variant={TextVariants.MUTED}>
-            {t('location.add_modal.labels.typeInfo')}
-          </Paragraph>
+          <Box marginBottom={3}>
+            <Paragraph variant={ParagraphVariants.MUTED}>
```
brampauwelyn, line 237: Maybe it's also worth questioning if we still need the `Paragraph` component, since it's only used 1 place.
We might just use the Text component with an extra tailwind class for the muted variant?

Could be part of a later clean-up PR.
We could add a `// TODO UI MIGRATION`? 
We could use a specific comment so we could easily do a search for it.
Or we could already add a JIRA ticket for it in the epic.
What do you prefer?

danmkt00: Maybe we could do the change later, for now the focus is the migration. Deciding which components to keep or remove is better as a bigger separate task, not part of individual migration PRs.

brampauwelyn: @danmkt00 Ok, I made a cleanup ticket for it. 
So we can pick it up once the migration is done :)
https://jira.publiq.be/browse/III-7235

brampauwelyn (approved):

src/pages/PlaceAddModal.tsx
danmkt00: thanks!

PR #1255, opened 2026-06-10
Minimal stack inline migration

brampauwelyn:
### Added
- StackLegacy.tsx, InlineLegacy.tsx — original implementations
- tailwindGap.ts — maps Box spacing scale to pixel-equivalent Tailwind gap classes

### Changed
- Stack.tsx — feature flag toggle; shadcn branch renders a Box with tw:flex-col
  and CSS gap instead of cloneElement margin injection, legacy delegates to StackLegacy
- Inline.tsx — same pattern as Stack with tw:flex; fixed inconsistency where
  InlineShadcn was spreading all props instead of using getBoxProps like StackShadcn

### Notes
- Gap values are pixel-equivalent to the legacy Box spacing scale to avoid having
  to update all Lost Pixel baselines during migration; once legacy is fully dropped
  these will be standardized to Tailwind's default gap scale

### Fixed
- Non-integer and out-of-range spacing values (e.g. 3.5) now fall back to the legacy implementation instead of silently losing all spacing


src/ui/Inline.tsx
```diff
@@ -30,74 +16,47 @@ type InlineProps = {
...
+  ({ spacing, className, children, as = 'span', ...props }, ref) => (
+    <Box
+      as={as}
+      ref={ref}
+      className={cn(
+        'tw:flex',
+        getGapClass(spacing as number | undefined),
+        className,
+      )}
+      {...getBoxProps(props)}
```
alduya, line 29: This drops non-box props that legacy forwarded. Are you sure no callers add non-box props?

src/ui/Stack.tsx
```diff
@@ -30,56 +16,45 @@ type StackProps = {
...
-    );
-  },
+const StackShadcn = forwardRef<HTMLElement, Props>(
+  ({ spacing, className, children, as = 'section', ...props }, ref) => (
+    <Box
+      as={as}
+      ref={ref}
+      className={cn(
+        'tw:flex tw:flex-col',
+        getGapClass(spacing as number | undefined),
```
alduya, line 26: > Legacy only injected margins on children that were non-string, non-Fragment components (isBoxComponent); raw DOM elements, text nodes, and Fragment contents got no spacing. CSS gap spaces every flex item unconditionally, and a Fragment's children become separate items each receiving the gap.

src/ui/tailwindGap.ts
```diff
@@ -0,0 +1,20 @@
+// Pixel-equivalent to the legacy Box spacing scale (2^n / 15 rem) to avoid
+// visual regressions in Lost Pixel during migration. Once legacy is fully
+// dropped these should be standardized to Tailwind's default gap scale.
+// Strings must stay static literals so Tailwind can extract them.
+const GAP_CLASS_BY_SPACING: Record<number, string> = {
```
alduya, line 5: On the activity review page, the action buttons on the right change spacing a lot.
In src/pages/OfferPreviewSidebar.tsx this is used: `spacing={3.5}`.
> This only has integer keys 0–8. The miss falls through ?? ''. So any non-integer spacing (and any value above 8) silently loses all its spacing under the flag.
Maybe add a check like: `props.spacing in GAP_CLASS_BY_SPACING`?

src/ui/Inline.tsx
danmkt00: Yes — confirmed. Grep across all 220+ <Inline usages found zero callers passing rel, target, data-*, or any non-box event handler. The gap exists in principle (getBoxProps drops what getInlineProps allowed for those), but nothing in the codebase exercises it today.

src/ui/Stack.tsx
danmkt00: The two callers with DOM element children either have no spacing prop (no gap emitted) or are a two-item row where gap and marginRight produce identical output. No visible regression today.

src/ui/tailwindGap.ts
danmkt00: Fixed — canUseShadcn now checks props.spacing in GAP_CLASS_BY_SPACING (guarded by typeof === 'number' for the type narrowing). Values like 3.5 or anything above 8 fall back to InlineLegacy automatically.

brampauwelyn:
I cannot approve it since I made it.
So @alduya @vhande could you review and approve :) (or not :) )

Thx

alduya (approved):

PR #1257, opened 2026-06-11
III-7175 migrate Link component to Tailwind

danmkt00:
### Added
- New `LinkBadgeVariants` (`BADGE_DANGER`, `BADGE_SECONDARY`, `BADGE_INFO`) as link-as-badge support — falls back to default link style without the flag
- shadcn-based `Link` implementation, toggled via `SHADCN_MIGRATION` feature flag
- `LinkLegacy.tsx` with the legacy fixed implementation
-  Storybook story for badge variant (`AsBadge`)

### Changed
- Migrated `css` prop usages on `<Link>` to Tailwind `className` in 6 files — `css` is silently dropped by `getBoxProps` in the new implementation, legacy path unaffected
- Updated Lost Pixel baseline for split dropdown button (new DOM structure renders more correctly)

### Fixed
- `iconName` and `suffix` were dropped for button-variant links — content wrapper was incorrectly skipped for both button and badge variants (now only skipped for badges), so icons and suffixes now render correctly inside button-variant links
- Button-variant link children double-wrapped in `<Text>` via `customChildren` prop

---

Ticket: https://jira.uitdatabank.be/browse/III-7175


alduya (approved):

src/pages/Footer.tsx
```diff
@@ -78,15 +78,13 @@ type FooterLinkProps = {
...
-      &:hover {
-        color: #222;
-      }
-    `}
+    className="tw:underline tw:hover:text-udb-text"
+    style={{
+      color:
+        footerVariant === FooterVariants.LOGIN
+          ? getValueForPage('footer.linkColor')
+          : colors.udbMainDarkestBlue,
```
brampauwelyn: Could we work with tailwind color classes here as well?

danmkt00: done

PR #1266, opened 2026-06-15
III-7239-Add tabs to the translate form

vhande:
### Added
-  tabs to the translate form

translate

---

Ticket: https://jira.uitdatabank.be/browse/III-7239


alduya:
@claude review please

src/pages/TranslateForm.tsx
```diff
@@ -104,128 +134,188 @@ const TranslateForm = () => {
...
+  );
+  const hasDescriptionContent = translationLanguages.some(
+    (lang) => !!offer?.description?.[lang],
+  );
 
-    router.push(
-      scope === OfferTypes.EVENTS ? `/events/${id}` : `/places/${id}`,
-    );
-  };
+  const titleFields = languageOptions.map((language) => (
```
brampauwelyn: I always find it more elegant if we define jsx as a component, you will probably have to move it outside of the component and pass the props.

```suggestion
  const TitleFields = languageOptions.map((language) => (
```

So you can use it inside the template as `<TitleFields />` instead of {titleFields} 

But I guess it's a matter of taste.

Same remark for `descriptionFields`

brampauwelyn (approved):

PR #1267, opened 2026-06-15
III-7180 migrate Alert component to shadcn

danmkt00:
### Added
- Alert: added title prop to render a bold heading above the description, useful for communicating the subject of the alert at a glance — shadcn implementation only
- Alert: added action prop to render a call-to-action element (e.g. a retry button or link) below the description — shadcn implementation only
- Semantic color tokens for alert variants (--color-destructive-muted, --color-success-muted, --color-info, --color-info-muted, --color-warning-muted) added to Tailwind theme

### Changed
- Alert component migrated to shadcn under the SHADCN_MIGRATION feature flag; legacy implementation preserved in AlertLegacy
- Alert is now visually smaller and more compact in line with shadcn design conventions
- Alert variants now use a left accent border with a small variant-colored icon; info variant uses a full border with no icon
- Callers that used BoxProps spacing props (marginY, marginTop, marginBottom, width, css) updated to use Tailwind className equivalents
- Storybook stories updated with WithTitle and WithAction; action added to argTypes
- HTML string content (ul, li, strong, b, code) in Alert restored with !important overrides due to GlobalStyle.js CSS reset; TODOs added for cleanup

Ticket: https://jira.uitdatabank.be/browse/III-7180


src/ui/Alert.tsx
```diff
@@ -1,220 +1,163 @@
...
-            }
-            &:active {
-              background-color: inherit;
-              filter: brightness(95%) saturate(150%);
-            }
-          `}
+        <button
           onClick={onClose}
-        ></Button>
+          className={cn(
```
alduya, line 114: The legacy version didn't have an aria-label. It would be good to add it now:
`aria-label={t('alert.close')}`

alduya: Is the close behaviour tested in an e2e test?

src/ui/Alert.tsx
```diff
@@ -1,220 +1,163 @@
...
-          <Text>{children}</Text>
-        ) : (
-          <Text
-            dangerouslySetInnerHTML={{ __html: children as string }}
-            css={`
-              strong {
-                font-weight: bold;
-              }
-
-              ul {
```
alduya, line 178: This styling gets removed in the new Alert. When you create or edit an event for schools, there is an Alert next to the description with a list that loses it's styling.
We could fix it in the Alert, or more general, remove the styling that breaks many `<ul>/<li>` styling

alduya: http://localhost:3000/events/d6b26c1e-9a52-4280-9ea9-1e7bf9de6156/edit

src/ui/Alert.tsx
```diff
@@ -1,220 +1,163 @@
...
 
-const Alert = ({
+const AlertShadcn = ({
   variant = AlertVariants.PRIMARY,
   visible = true,
   children,
   fullWidth,
   closable,
   onClose,
-  ...props
```
alduya, line 136: > ## The mechanism
> AlertProps = InlineProps & { … } (Alert.tsx), so every Box/Inline spacing/sizing prop is type-valid on `<Alert>`. The wrapper destructures only the named props and collects the rest into ...props (Alert.tsx:116-127), then forwards ...props only to AlertLegacy (:141). The AlertShadcn call (:148-160) receives just the explicitly-listed props — variant, visible, children, fullWidth, closable, onClose, title, action, className. AlertShadcn's own signature has no ...props either. So in the shadcn path, any other InlineProps are dropped on the floor, and TypeScript stays silent because they're legal props.
> 
> It's not just theoretical — these callers were missed by the migration (none are in the PR's changed-files list, so they still pass Box props):
> 
> - src/pages/AlertDuplicatePlace.tsx:59 — <Alert variant={variant} maxWidth="50rem" {...props}> → maxWidth and whatever ...props carries are dropped; the alert loses its 50rem cap and renders full-width.
> - src/pages/steps/TimeTableStep.tsx:91 — <Alert visible variant={AlertVariants.DANGER} maxWidth="53rem"> → loses the 53rem cap.
> - src/pages/organizers/[organizerId]/preview/RequestOwnershipModal.tsx:99 and :110 — marginBottom={4} dropped → spacing collapses (these are also the closable ownership alerts).
> - src/pages/organizers/[organizerId]/preview/index.page.tsx:107 — marginBottom={4} dropped.
>
> While the flag is off these still work (legacy path gets ...props); the day SHADCN_MIGRATION is enabled in production, these four alerts lose their width caps / margins with no compile error and no test failure to warn anyone.
> 
> ## Two fixes, and they're complementary:
> 1. Make it loud instead of silent — narrow the shadcn-facing prop type so leftover InlineProps become type errors rather than silently-dropped props. That converts every missed caller (including these four) into a yarn types:check failure, which is how the migration should have surfaced them.
> 2. Migrate the four stragglers — convert their maxWidth/marginBottom to the className equivalents like the rest of the PR did.

src/ui/Alert.tsx
danmkt00: The aria-label is added in 778a81b6a, but we used common.close instead of alert.close since a generic "close" key fits better in the shared namespace rather than creating an alert-specific one.

danmkt00: > Is the close behaviour tested in an e2e test?

Not yet — there's no e2e test for the close behavior.

src/ui/Alert.tsx
danmkt00: Fixed in 8ef045929 with !important overrides on the dangerouslySetInnerHTML span for ul, li, strong, b, and code. The root cause is the list-style: none / font: inherit reset in GlobalStyle.js — TODOs added in both files for a proper cleanup later.

src/ui/Alert.tsx
danmkt00: Migrated the four straggler callers (AlertDuplicatePlace, TimeTableStep, RequestOwnershipModal, index.page.tsx) in c9c6d80ec from Box props to Tailwind className equivalents. Type narrowing will be done when the legacy path is removed and InlineProps is dropped from AlertProps.

.lostpixel/baseline/components-alert--with-action.png
brampauwelyn, line 1: The position of the "Retry" seems a bit off?
Shouldn't it be aligned like the other text?  Now it's a bit indented

.lostpixel/baseline/components-alert--primary.png
brampauwelyn, line 1: The icons are different.
Is this on purpose?

.lostpixel/baseline/components-alert--danger.png
brampauwelyn, line 1: Different icons.
Is this on purpose?

.lostpixel/baseline/components-alert--success.png
brampauwelyn, line 1: Different icons

src/ui/Alert.tsx
```diff
@@ -1,220 +1,170 @@
...
-  );
-};
 
 const AlertVariants = {
   PRIMARY: 'primary',
   INFO: 'info',
   SUCCESS: 'success',
   DANGER: 'danger',
   WARNING: 'warning',
-  DARK: 'dark',
```
brampauwelyn, line 109: The `DARK` variant has been removed.
But it is still used in `Events.js`.

I guess you can just remove the variant from there. It will fallback to PRIMARY

src/styles/tailwind.css
```diff
@@ -57,6 +58,12 @@
   --color-success-foreground: #ffffff;
   --color-destructive: #dd5242; /* udb-danger */
   --color-destructive-foreground: #ffffff;
+  --color-destructive-muted: #fae5e3;
+  --color-success: #6bcd69; /* udb-main-positive-green */
+  --color-success-foreground: #ffffff;
```
brampauwelyn: Duplicate property
This has already been defined on line `58`
```suggestion
```

src/styles/tailwind.css
```diff
@@ -57,6 +58,12 @@
   --color-success-foreground: #ffffff;
   --color-destructive: #dd5242; /* udb-danger */
   --color-destructive-foreground: #ffffff;
+  --color-destructive-muted: #fae5e3;
+  --color-success: #6bcd69; /* udb-main-positive-green */
```
brampauwelyn: Duplicate property

```suggestion
```

.lostpixel/baseline/components-alert--with-action.png
danmkt00: Fixed in 6d145e9a3 — the indentation was caused by the Button component's own padding. The story was using ButtonVariants.LINK as an example. Replaced with a plain <a> tag which aligns correctly with the rest of the alert text.

src/ui/Alert.tsx
danmkt00: Fixed in 992d80b28 — removed DARK from AlertLegacy.tsx and explicitly replaced its usage in Event.js with PRIMARY instead of relying on the default fallback.

src/styles/tailwind.css
danmkt00: Fixed in 410403ea7

src/styles/tailwind.css
danmkt00: Fixed in 410403ea7

.lostpixel/baseline/components-alert--primary.png
danmkt00: Intentional, legacy used custom inline SVGs, new implementation uses the shared Icon component

brampauwelyn:
There seems to be a small issue with the legacy alert.
E.g. in the description step the alert takes too much space.

Screenshot 2026-06-18 at 09 59 25


With the shadcn feature flag enabled it looks good.
Screenshot 2026-06-18 at 10 10 38


Can you have a look at it?

danmkt00:
> There seems to be a small issue with the legacy alert. E.g. in the description step the alert takes too much space.
> 
> Can you have a look at it?

@brampauwelyn AlertLegacy was missing className prop causing Tailwind classes to be silently dropped b9e30ea7d

brampauwelyn:
Looks good now 👍

brampauwelyn (approved):

PR #1269, opened 2026-06-16
III-7242 fix step title layout

danmkt00:
### Changed
- StepWrapper now uses an explicit Inline wrapper with tw:gap-2 and tw:mb-5 Tailwind classes instead of passing alignItems and spacing directly to Title

### Fixed
- Step number and title no longer stack vertically when the shadcn migration flag is enabled
Screenshot 2026-06-16 at 14 15 39

---

Ticket: https://jira.uitdatabank.be/browse/III-7242


src/pages/steps/Steps.tsx
```diff
@@ -109,9 +110,11 @@ const StepWrapper = ({
...
       {...getStackProps(props)}
     >
-      <Title color={getValue('title.color')} alignItems="center" spacing={3}>
-        <NumberIndicator>{stepNumber}</NumberIndicator>
-        <Text lineHeight="1rem">{title}</Text>
+      <Title color={getValue('title.color')} className="tw:mb-5">
+        <Inline alignItems="center" className="tw:gap-2">
+          <NumberIndicator>{stepNumber}</NumberIndicator>
+          <Text>{title}</Text>
+        </Inline>
```
brampauwelyn: Since we are migrating away from `Stack` and `Inline`.
We could probably better use a `div` element here with the appropriate tailwind.

Or better could we directly pass the classNames to `Title` component.

Also verify the behavior when the feature flag is off.


```suggestion
        <div className="tw:flex-row tw:gap-2">
          <NumberIndicator>{stepNumber}</NumberIndicator>
          <Text>{title}</Text>
        </div>
```

src/pages/steps/Steps.tsx
```diff
@@ -109,9 +110,11 @@ const StepWrapper = ({
       spacing={4}
       {...getStackProps(props)}
     >
-      <Title color={getValue('title.color')} alignItems="center" spacing={3}>
-        <NumberIndicator>{stepNumber}</NumberIndicator>
-        <Text lineHeight="1rem">{title}</Text>
+      <Title color={getValue('title.color')} className="tw:mb-5">
```
brampauwelyn: I don't know if we need the nested Inline element.
Can we pass the necessary tailwind classes directly to the `Title` component?


```suggestion
      <Title color={getValue('title.color')} className="tw:flex-row tw:mb-5">
```

brampauwelyn (approved):

PR #1282, opened 2026-06-22
III-7266-Add ff for subevent reservation info

vhande:
### Added
- ff for subevent reservation info

---

Ticket: https://jira.uitdatabank.be/browse/III-7266


brampauwelyn (approved):
Thx @vhande 🥇

PR #1265, opened 2026-06-15
III-7179 migrate Icon component to Tailwind

danmkt00:
### Added
- Icon Lucide React implementation behind `SHADCN_MIGRATION` feature flag ([c32f50e](https://github.com/cultuurnet/udb3-frontend/commit/c32f50e3079d1168c31d48c06753856998a2e77b))
- `IconLegacy.tsx` preserving the existing FontAwesome implementation for the fallback path ([c32f50e](https://github.com/cultuurnet/udb3-frontend/commit/c32f50e3079d1168c31d48c06753856998a2e77b))
- All icon choices reviewed and approved by Sarah — see full visual overview in [Figma](https://www.figma.com/design/XqFJViHomujOogausP8U7l/Untitled?node-id=0-1&t=c97VplYjSwIIl4hV-1)

### Changed
- Sizing logic moved from a `.svg-inline--fa` CSS selector (a FontAwesome-specific class injected on its SVGs) to `width`/`height` props passed directly on the Lucide SVG element — Lucide icons render as plain `<svg>` without that class so the selector approach no longer works ([c32f50e](https://github.com/cultuurnet/udb3-frontend/commit/c32f50e3079d1168c31d48c06753856998a2e77b))
- Icon `color` BoxProps replaced with Tailwind `className` semantic tokens (`tw:text-success`, `tw:text-destructive`, `tw:text-warning`) across all call sites — colors slightly adjusted to standardize (like in ProgressBar) ([3b35694](https://github.com/cultuurnet/udb3-frontend/commit/3b35694ce))
- Remaining Icon BoxProps (`marginLeft`, `onClick`, `opacity`, `alignItems`, spinner `css` animation) replaced with Tailwind classes and proper `Button` wrappers ([1ff21ef](https://github.com/cultuurnet/udb3-frontend/commit/1ff21ef67))
- Icon `width`/`height` props changed from accepting strings (e.g. `"0.8rem"`, `"10rem"`) to number only — all call sites updated to pixel equivalents ([e1bf6bf](https://github.com/cultuurnet/udb3-frontend/commit/e1bf6bfcb))
- Toast close icon `onClick` moved to a wrapping `Button` to satisfy the `IconShadcn` type contract ([3ab4bbc](https://github.com/cultuurnet/udb3-frontend/commit/3ab4bbc1a))
[...]

PR #1284, opened 2026-06-23
III-7184 migrate Tabs component to shadecn

danmkt00:
### Added
- [Add shadcn Tabs component with feature flag migration path](https://github.com/cultuurnet/udb3-frontend/pull/1284/changes/2abb218fd48deb4e55062b6df04bb83b30eb9b9c) — extracts Bootstrap implementation to `TabsLegacy`, adds Radix UI `shadcn/tabs` with CVA variants for default, outlined and floating
- [Add compact prop to Tabs floating variant](https://github.com/cultuurnet/udb3-frontend/pull/1284/changes/efb20ad6d) — reduces horizontal trigger padding on the floating variant; implemented via CVA `size` compound variant in shadcn and a conditional `css` override in `TabsLegacy`, so it works for both implementations

### Changed
- [Fix Tabs callers for shadcn/legacy compatibility](https://github.com/cultuurnet/udb3-frontend/pull/1284/changes/a6fc84b75b208a3ba3f304ccba342108df359479) — removes Bootstrap-specific `css` props, adds missing `key` props on mapped tabs, removes `activeBackgroundColor="white"` from dashboard, replaces `css` padding with `tw:pt-4` in AdditionalInformationStep

### Fixed
- [Update lostpixel baselines for Tabs component](https://github.com/cultuurnet/udb3-frontend/pull/1284/changes/ad9ca11c181f7226929445c7585da4e211f3ed7c) — updated default and outlined baselines, added floating

### Notes
- `TabTitle` renders an `Icon` (div) inside `ShadcnTabsTrigger` causing an invalid HTML nesting warning — deferred to [III-7179-icon-component](https://github.com/cultuurnet/udb3-frontend/pull/1265). No user impact, flag is off in production.
- `forceMount` keeps all tab content mounted at all times (matching Bootstrap behavior) so validation icons appear without clicking the tab first.

---

Ticket: https://jira.uitdatabank.be/browse/III-7184


brampauwelyn:
@danmkt00 we discussed this a bit with Sara.
See conversation in #mwl-publiq-frontend on slack.
We want to try to shadcn's default black version for the tabs and see how it feels...
Could you modify the PR a bit to keep the default shadcn styling?

We could still add the colors afterwards if it's really necessary.

danmkt00:
> we discussed this a bit with Sara. See conversation in #mwl-publiq-frontend on slack. We want to try to shadcn's default black version for the tabs and see how it feels... Could you modify the PR a bit to keep the default shadcn styling?
> 
> We could still add the colors afterwards if it's really necessary.

Done — see 0dec86a87. Also worth noting: if you ever need the blue back on a specific instance, you can pass activeBackgroundColor to Tabs which overrides the color of both the active underline and the active tab text.

src/ui/Tabs.stories.tsx
```diff
@@ -29,13 +34,40 @@ type Story = StoryObj<typeof Tabs>;
...
+          <Text>PLACES</Text>
+        </Tabs.Tab>
+        <Tabs.Tab eventKey="organizers" title="Organizers">
+          <Text>ORGANIZERS</Text>
+        </Tabs.Tab>
+      </Tabs>
+    );
+  },
+};
+
```
alduya, line 60:
```suggestion

export const FloatingCompact: Story = {
  args: {
    activeKey: 'events',
    variant: TabsVariants.FLOATING,
    compact: true,
  },
  render: function RenderComponent(args) {
    const [activeKey, setActiveKey] = useState(args.activeKey || 'events');

    return (
      <Tabs
        {...args}
        activeKey={activeKey}
        onSelect={(key) => setActiveKey(key as string)}
      >
        <Tabs.Tab eventKey="events" title="Events">
          <Text>EVENTS</Text>
        </Tabs.Tab>
        <Tabs.Tab eventKey="places" title="Places">
          <Text>PLACES</Text>
        </Tabs.Tab>
        <Tabs.Tab eventKey="organizers" title="Organizers">
          <Text>ORGANIZERS</Text>
        </Tabs.Tab>
      </Tabs>
    );
  },
};
```
Because there is no story using compact, there is also no vrt baseline image.

src/ui/Tabs.tsx
```diff
@@ -1,187 +1,112 @@
...
+  className?: string;
+};
 
-export const TabsVariants = {
-  DEFAULT: 'default',
-  OUTLINED: 'outlined',
-  FLOATING: 'floating',
-} as const;
+type Props<T extends string = string> = {
+  activeKey?: string;
```
alduya: Why not make this required?

src/ui/Tabs.tsx
```diff
@@ -1,187 +1,112 @@
...
+      className={className}
+    >
+      <ShadcnTabsList variant={shadcnVariant}>
+        {tabs.map((tab) => (
+          <ShadcnTabsTrigger
+            key={tab.eventKey}
+            value={tab.eventKey}
+            variant={shadcnVariant}
+            size={compact ? 'sm' : 'md'}
+            className={tab.className}
```
alduya: In shadcn this classname is placed on the trigger. In legacy it is placed on the tab content. Currently not a problem since it isn't used anywhere. Maybe move it to `<ShadcnTabsContent>`

src/ui/Tabs.tsx
```diff
@@ -1,187 +1,112 @@
...
-        margin: 0 !important;
-        background-color: white;
-
-        &:hover {
-          background-color: ${grey1};
-          border: 1px solid black;
-        }
+}: Props<T>) {
+  const tabs = Children.toArray(children)
+    .filter(isValidElement)
```
alduya: This might allow non `Tabs.Tab` elements like a `<div>`. This would create an empty, never-selectable phantom Tab.

```suggestion
    .filter((child) => isValidElement(child) && child.type === Tabs.Tab)
```

src/ui/Tabs.stories.tsx
danmkt00: [added a new story and updated baselines](https://github.com/cultuurnet/udb3-frontend/pull/1284/changes/5ce8536d9818f9d62df61c7504d0a8d4a47f8abc)

src/ui/Tabs.tsx
danmkt00: I agree, all the callers were indeed passing activeKey anyway, and making it required is the behavior that we expect [commit](https://github.com/cultuurnet/udb3-frontend/pull/1284/changes/8961f07eef1008ab9a43bac7cf476132dbbc41c0)

src/ui/Tabs.tsx
danmkt00: [done](https://github.com/cultuurnet/udb3-frontend/pull/1284/changes/877464fc9a94bd20905cb5c9476d249c719bdf55)

src/ui/Tabs.tsx
danmkt00: [done](https://github.com/cultuurnet/udb3-frontend/pull/1284/commits/d722dd6d370170bf0008112ced9e7861c6a277e2)

brampauwelyn (approved):

PR #1286, opened 2026-06-24
III-7268-Reservation information for multiple events

vhande:
### Added
- ReservationUrlSection

---

Ticket: https://jira.uitdatabank.be/browse/III-7268

### DIFF
--- a/src/i18n/nl.json
+++ b/src/i18n/nl.json
@@ -465,6 +465,8 @@
           "availability": "Controleer beschikbaarheid",
           "subscribe": "Schrijf je in"
         },
+        "reservation_type_general": "1 algemene link",
+        "reservation_type_specific": "Een link per moment of periode",
         "reservation_period": {
           "title": "Wanneer kunnen mensen zich inschrijven?",
           "toggle": "Reservatieperiode",
--- a/src/i18n/nl.json
+++ b/src/i18n/nl.json
@@ -465,8 +465,8 @@
           "availability": "Controleer beschikbaarheid",
           "subscribe": "Schrijf je in"
         },
-        "reservation_type_general": "1 algemene link",
-        "reservation_type_specific": "Een link per moment of periode",
+        "reservation_type_single": "Algemene link",
+        "reservation_type_multiple": "Links per moment/periode",
         "reservation_period": {
           "title": "Wanneer kunnen mensen zich inschrijven?",
           "toggle": "Reservatieperiode",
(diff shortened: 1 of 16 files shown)


src/i18n/nl.json
```diff
@@ -465,6 +465,8 @@
           "availability": "Controleer beschikbaarheid",
           "subscribe": "Schrijf je in"
         },
+        "reservation_type_general": "1 algemene link",
```
brampauwelyn:
```suggestion
        "reservation_type_single": "1 algemene link",
```

src/i18n/nl.json
```diff
@@ -465,6 +465,8 @@
...
+        "reservation_type_specific": "Een link per moment of periode",
```
brampauwelyn:
```suggestion
        "reservation_type_multiple": "Een link per moment of periode",
```

src/pages/steps/AdditionalInformationStep/BookingInfoStep.tsx
```diff
@@ -355,6 +571,10 @@ const BookingInfoStep = ({
   const bookingInfo = getOfferByIdQuery.data?.bookingInfo;
   const bookingAvailability = getOfferByIdQuery.data?.bookingAvailability;
   const subEvents = getOfferByIdQuery.data?.subEvent ?? [];
+  const calendarType = getOfferByIdQuery.data?.calendarType;
+  const canEditBookingAvailability =
+    calendarType !== CalendarType.PERIODIC &&
+    calendarType !== CalendarType.PERMANENT;
```
brampauwelyn, line 573: Maybe we can turn it around.
So it directly clear for which CalendarTypes we allow bookingAvailability edits
```suggestion
  const canEditBookingAvailability =
    calendarType === CalendarType.SINGLE ||
    calendarType === CalendarType.MULTIPLE;

```

src/pages/steps/AdditionalInformationStep/BookingInfoStep.tsx
```diff
@@ -277,6 +281,216 @@ const ReservationPeriod = ({
   );
 };
 
+type ReservationType = 'general' | 'specific';
```
brampauwelyn:
```suggestion
type ReservationType = 'single' | 'multiple';
```

brampauwelyn (commented):
You could extend the test so you test your new functionality.
Switching between general / specific (single/multiple).
Also when you switch from specific -> general, you'll probably have to clear the bookingInfo from the subEvents, since as a user you decided  to only provide general info. 

So the initial read will also work.
When you load an existing event, the correct tab should be visible.
Now it always shows the general tab by default even when there's already specific bookingInfo on the subevents

brampauwelyn (approved):

PR #1292, opened 2026-06-26
III-7280-Add BOA cookie page

vhande:
### Added
- BOA cookie enable page

enable
remove

---

Ticket: https://jira.uitdatabank.be/browse/III-7280


brampauwelyn:
Maybe also add an active case?
So they can also de-activate it? 
Would it not be better to load it with the sidebar?

brampauwelyn (approved):

PR #1298, opened 2026-06-30
III-7185 migrate Tooltip component to shadcn

danmkt00:
Screenshot 2026-06-30 at 12 49 32

### Added
- `Tooltip` renders under `SHADCN_MIGRATION` feature flag using shadcn/Radix primitives
- `TooltipLegacy` extracted from original `Tooltip` for legacy Bootstrap fallback
- `DefaultTrigger` — `?` badge rendered by default when no `children` passed, matching legacy behavior
- Storybook stories for `Tooltip` with default, button, icon button, and text triggers

### Changed
- `Tooltip` now accepts optional `children` prop as the trigger element — can be any React node (button, icon, text, etc.)
- `tailwind.config.js` — added `tailwindcss-animate` plugin to enable tooltip enter/exit animations
- Tooltip's placement prop renamed to side, typed directly from Radix's side ('top' | 'right' | 'bottom' | 'left') instead of react-bootstrap's Placement — the bootstrap type had extra values (auto, -start/-end variants) that weren't used by either consumer
- DefaultTrigger (the ? badge) is now keyboard-focusable (tabIndex={0}) and gets an explicit aria-label set to the tooltip's content, so it's reachable via Tab and announced properly by screen readers — previously it was skipped entirely by keyboard navigation

---

Ticket: https://jira.uitdatabank.be/browse/III-7185


src/ui/Tooltip.tsx
```diff
@@ -1,23 +1,53 @@
...
+  TooltipTrigger,
+} from '@/ui/shadcn/tooltip';
+import { TooltipLegacy } from '@/ui/TooltipLegacy';
 
 type Props = {
-  id: string;
+  id?: string; // TODO: remove after migration — Radix handles accessibility automatically
   content: string;
+  children?: ReactNode;
 } & Pick<OverlayTriggerProps, 'placement'>;
```
brampauwelyn: We could already use the type from shadcn since the positions overlap (bootstrap one has extra ones but we don't use them).  So we only have minimal legacy things here...


```suggestion
  children?: ReactNode;
  placement?: TooltipSide;
};
```

If there are not many usages of `Tooltips` it might also be worth renaming the `placement` prop to `side` so it's closer to shadcn's API.

src/ui/Tooltip.tsx
```diff
@@ -1,23 +1,53 @@
-import { OverlayTrigger, OverlayTriggerProps } from 'react-bootstrap';
-import { Tooltip as BootstrapTooltip } from 'react-bootstrap';
+import { ReactNode } from 'react';
+import { OverlayTriggerProps } from 'react-bootstrap'; // TODO: remove after migration — only used for placement type
```
brampauwelyn: This can be removed after you implement my comment below...

```suggestion
```

src/ui/Tooltip.tsx
```diff
@@ -1,23 +1,53 @@
...
+  if (!isShadcnMigration) {
+    return <TooltipLegacy id={id} content={content} placement={placement} />;
+  }
+
+  const side = placement?.toString().split('-')[0] as
+    | 'top'
+    | 'right'
+    | 'bottom'
+    | 'left'
+    | undefined;
```
brampauwelyn: Can be removed if you use my suggestion above...

src/ui/Tooltip.tsx
```diff
@@ -1,23 +1,53 @@
...
   content: string;
+  children?: ReactNode;
 } & Pick<OverlayTriggerProps, 'placement'>;
 
-const Tooltip = ({ id, content, placement }: Props) => {
+const TOOLTIP_DELAY_MS = 100;
+
+const DefaultTrigger = () => (
+  <Badge variant={BadgeVariants.SECONDARY} pill>
+    ?
```
brampauwelyn: Could you make the DefaultTrigger more accessible. 
Currently we cannot access the element via the tab button.

src/ui/Tooltip.tsx
danmkt00: Fixed in b23442f4d

src/ui/Tooltip.tsx
danmkt00: Fixed in https://github.com/cultuurnet/udb3-frontend/commit/b23442f4dd5c9c6ee3f804271332dd1976131288

src/ui/Tooltip.tsx
danmkt00: Fixed in https://github.com/cultuurnet/udb3-frontend/commit/b23442f4dd5c9c6ee3f804271332dd1976131288

src/ui/Tooltip.tsx
danmkt00: Fixed in 17766cb70 — "make Tooltip's default "?" trigger keyboard-focusable and announce content".

src/ui/Tooltip.tsx
danmkt00: and fb24f4c3d — "rename placement prop to side to match shadcn's API".

src/ui/Tooltip.tsx
alduya: > The fix makes the ? reachable by `Tab`, but it's only superficially resolved. The aria-label sits on a generic `<span>` where most screen readers ignore it, and Radix's `aria-describedby` and trigger props land on the non-focusable wrapper span rather than the focusable element. Bram's underlying concern — that the trigger works properly for keyboard and screen reader users — would be more robustly addressed by making the badge a real `<button type="button">` and passing it directly to `TooltipTrigger` `asChild`.
[III-7185-tooltip-trigger-a11y.patch](https://github.com/user-attachments/files/29600773/III-7185-tooltip-trigger-a11y.patch)

brampauwelyn: @danmkt00 yes maybe use a button as the default trigger ☝️

danmkt00: [made default trigger a real button with secondary badge looks](https://github.com/cultuurnet/udb3-frontend/pull/1298/changes/fb25841259809d0da374eb8a21d4fae281e6c76a)

brampauwelyn (approved):

PR #1299, opened 2026-06-30
III-7291 fix(Inline): handle forwardedAs in shadcn path

danmkt00:
### Fixed

- Fixed Inline not respecting forwardedAs in the shadcn migration path, causing label clicks to not toggle associated inputs. Replaced getBoxProps spread with direct {...props} so valid HTML attributes (like htmlFor) are no longer silently dropped — Box already handles prop filtering via shouldForwardProp.

---

Ticket: https://jira.uitdatabank.be/browse/III-7291


brampauwelyn (approved):

brampauwelyn:
Thx for noticing and coming up with a fix 👍

PR #1300, opened 2026-06-30
III-7292-Add region prefix to holiday descriptions

vhande:
### Added
- Region prefix to holiday descriptions

---

Ticket: https://jira.uitdatabank.be/browse/III-7292

### DIFF
--- a/src/ui/DatePeriodPicker.tsx
+++ b/src/ui/DatePeriodPicker.tsx
@@ -42,33 +42,20 @@ type HolidayPreset = {
 
 const locales = { nl, fr, de };
 
-const getHolidayLabel = (
-  holiday: ApiHoliday,
-  language: string,
-  t: TFunction,
-): string => {
+const parseHoliday = (holiday: ApiHoliday, language: string, t: TFunction) => {
   const name =
     holiday.name[language as Values<typeof SupportedLanguages>] ?? '';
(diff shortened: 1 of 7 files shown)


alduya (approved):

brampauwelyn:
Main concern — the French forms are grammatically off
French doesn't place origin adjectives before the noun, so the prefix pattern that works for nl/de breaks down here.
Example `Flamandes Vacances de carnaval` is grammatically wrong.

French wants the adjective after: "vacances de carnaval flamandes".

Screenshot 2026-06-30 at 16 26 00

In the quicklinks it is correct 🤔

src/ui/DatePeriodPicker.tsx
```diff
@@ -106,7 +106,19 @@ const filterHolidaysForPreset = (
...
+        const regionPrefix = t(
+          `date_period_picker.region_prefix.${parsed.region}`,
+          { defaultValue: '' },
+        );
+        if (regionPrefix) {
+          return { ...parsed, name: `${regionPrefix} ${parsed.name}` };
+        }
+      }
+      return parsed;
+    })
```
brampauwelyn: You can probably move the logic directly inside the parseHoliday function? 


```suggestion
     return parseHoliday(holiday, language, t);
```

PR #1304, opened 2026-07-01
III-7186 migrate Pagination component to shadecn

danmkt00:
### Added
- shadcn/ui-based `Pagination` component, behind the `SHADCN_MIGRATION` flag (`PaginationLegacy` kept as fallback).

### Changed
- Previous/Next buttons show text labels alongside the arrows instead of icon-only — design agreed with Sarah.
- Reworked page-range calculation: page 1 is now a fixed anchor before the ellipsis, instead of a sliding window that could hide it.

### Removed
- Legacy Bootstrap pagination markup and styled-components overrides from the active component.

### Fixed
- Pagination controls now disable while a page is fetching instead of unmounting/flashing a spinner.
- Label suggestion search no longer keeps stale results while typing — previously a quick click could select a leftover suggestion from before the search term updated.
- Storybook story wires `currentPage`/`onChangePage` to local state, so it's actually clickable.
- Updated the flaky "can paginate labels/roles" e2e specs to assert on actual row content instead of racing `networkidle`.

### Accessibility
- Translated aria-labels (nl/fr/de) on Previous/Next and page-number buttons, so screen readers announce more than a bare icon/number.
- `aria-live="polite"` status region announces "Page X of Y" on page change.

---

Ticket: https://jira.uitdatabank.be/browse/III-7186


src/ui/Pagination.tsx
```diff
@@ -35,149 +36,94 @@ const Pagination = ({
     return pages;
   }, [totalItems, perPage]);
 
-  const currentRange = useMemo(() => {
-    const middle = Math.ceil(limitPages / 2);
+  const renderPage = (page: number) => (
+    <PaginationItem key={page}>
+      <PaginationLink
+        isActive={page === currentPage}
+        onClick={(e) => {
```
alduya: Shadcn normally works with passing href for the links. Without them, the anchor isn't keyboard-focusable. > This is a regression: the legacy react-bootstrap items render focusable anchors with role="button" and key handling. Fix: render a `<button type="button">` instead of `<a>` (and use disabled on Previous/Next rather than only aria-disabled), or give the anchors href values — the e.preventDefault() calls in Pagination.tsx already anticipate that.

src/ui/Pagination.tsx
```diff
@@ -35,149 +36,94 @@ const Pagination = ({
...
-          <BootstrapPagination.Item
-            key={page}
-            active={page === currentPage}
-            onClick={() => {
-              onChangePage(page);
+    <PaginationRoot>
+      <PaginationContent>
+        <PaginationItem>
+          <PaginationPrevious
+            onClick={(e) => {
```
alduya: Same here

src/ui/Pagination.tsx
```diff
@@ -35,149 +36,94 @@ const Pagination = ({
...
+        {startPages.map(renderPage)}
+        {hasEllipsis && (
+          <PaginationItem key="ellipsis">
+            <PaginationEllipsis />
+          </PaginationItem>
+        )}
+        {windowPages.map(renderPage)}
+        <PaginationItem>
+          <PaginationNext
+            onClick={(e) => {
```
alduya: And here

src/ui/Pagination.tsx
danmkt00: Went with `<button type="button">` + real `disabled` instead of href'd anchors — browser handles focus/keyboard blocking natively, no more `aria-disabled` + `preventDefault()` workarounds. Also had to reset button chrome in [button.tsx](vscode-webview://0pje84ealej80m4i54521kdk3m616vc4o5v8e4u4nu408v7fa670/src/ui/shadcn/button.tsx) since we don't load Tailwind preflight. [Commit](https://github.com/cultuurnet/udb3-frontend/pull/1304/changes/f375434c1b00db152a37dda2e49f15f32c20b6a5)

brampauwelyn (approved):

PR #1312, opened 2026-07-02
III-7226 - Enable feature flags via .env vars

brampauwelyn:
### Added
- [example for enabling ff via .env var](https://github.com/cultuurnet/udb3-frontend/commit/9b1ffd7c24abf9690103d34be62edcde579f3f20)
- [possibility to enable ff via .env var](https://github.com/cultuurnet/udb3-frontend/commit/2c1ea564b90bb66448ea1124969afec2dd6d5366)

---
Ticket: https://jira.uitdatabank.be/browse/III-7226


alduya (approved):
Ik zal dit ook toevoegen op UiTPAS Balie.
Wat is het gedrag wanneer je een feature flag probeert uit te zetten vie het console command?

brampauwelyn:
> Ik zal dit ook toevoegen op UiTPAS Balie. Wat is het gedrag wanneer je een feature flag probeert uit te zetten vie het console command?

If the .env var is enabled (true), you cannot disable it via the cookie. 
I think is the desired behavior since otherwise it might be confusing. 
If a product owner would still have a cookie with the value false for example...

vhande (approved):

alduya:
I wasn't clear enough. I meant this:
Screenshot 2026-07-06 at 09 08 34
That feature flag is enabled from `.env`, when you try to disable it from the console, it isn't clear that it is still enabled. Not a problem for local development where you can look at the `.env` file. But confusing for non-dev people on the other environments.

brampauwelyn:
> I wasn't clear enough. I meant this: Screenshot 2026-07-06 at 09 08 34 That feature flag is enabled from `.env`, when you try to disable it from the console, it isn't clear that it is still enabled. Not a problem for local development where you can look at the `.env` file. But confusing for non-dev people on the other environments.

Ah yes, now I added a new `source` field as well, so it's clear why it's enabled.

Screenshot 2026-07-06 at 13 32 11

Screenshot 2026-07-06 at 13 29 22

PR #1321, opened 2026-07-14
[fix] III-7185 wrap custom Tooltip trigger children in a span

danmkt00:
Screenshot 2026-07-14 at 10 45 01

### Fixed
- Tooltip: custom trigger children (e.g. Button) no longer respond to hover/focus. Removing the wrapping span when making the default trigger a real button also stripped it from custom children, so Radix's injected handlers were silently dropped by Button's prop filtering. Children are now wrapped in a span again.

---

Ticket: https://jira.uitdatabank.be/browse/III-7185

### DIFF
--- a/src/ui/Tooltip.tsx
+++ b/src/ui/Tooltip.tsx
@@ -48,7 +48,11 @@ const Tooltip = ({ id, content, side, children }: Props) => {
     <TooltipProvider delayDuration={TOOLTIP_DELAY_MS}>
       <ShadcnTooltip>
         <TooltipTrigger asChild>
-          {children ?? <DefaultTrigger content={content} />}
+          {children ? (
+            <span>{children}</span>
+          ) : (
+            <DefaultTrigger content={content} />
+          )}
         </TooltipTrigger>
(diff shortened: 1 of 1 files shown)


src/ui/Tooltip.tsx
```diff
@@ -48,7 +48,11 @@ const Tooltip = ({ id, content, side, children }: Props) => {
     <TooltipProvider delayDuration={TOOLTIP_DELAY_MS}>
       <ShadcnTooltip>
         <TooltipTrigger asChild>
-          {children ?? <DefaultTrigger content={content} />}
+          {children ? (
+            <span>{children}</span>
+          ) : (
+            <DefaultTrigger content={content} />
+          )}
```
brampauwelyn, line 55: Is it not possible to do it like this?

```suggestion
            <DefaultTrigger content={content}>{children}</DefaultTrigger>
          )}
```

danmkt00: no need since we are not planning to change DefaultTrigger's behavior and look, for that we have children

brampauwelyn (approved):

PR #1324, opened 2026-07-14
III-7335 Add aria-label to icon-only links for screen reader support

danmkt00:
### Fixed
- Icon-only links exposed no accessible name to screen readers (announced only as "link" with no context). Added aria-label to the "Custom Children" example in the Link component's Storybook story and to the sidebar logo link. A codebase-wide search confirmed the sidebar logo was the only production usage missing it.

---

Ticket: https://jira.uitdatabank.be/browse/III-7335

### DIFF
--- a/src/ui/Link.stories.tsx
+++ b/src/ui/Link.stories.tsx
@@ -83,7 +83,12 @@ export const CustomChildren: Story = {
     },
   },
   render: (args) => (
-    <Link customChildren={true} href={args.href} variant={args.variant}>
+    <Link
+      customChildren={true}
+      href={args.href}
+      variant={args.variant}
+      aria-label="View profile"
+    >
       <Icon name={args.iconName} />
     </Link>
   ),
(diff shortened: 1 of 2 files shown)


src/ui/Link.stories.tsx
```diff
@@ -83,7 +83,12 @@ export const CustomChildren: Story = {
     },
   },
   render: (args) => (
-    <Link customChildren={true} href={args.href} variant={args.variant}>
+    <Link
+      customChildren={true}
+      href={args.href}
+      variant={args.variant}
+      aria-label="View profile"
```
brampauwelyn, line 90: We should probably use translatable labels here as well. Can be different in dutch vs french etc...

danmkt00: do we need it for the story?

brampauwelyn: Ah sorry didn't see it was in a story 👍

brampauwelyn (approved):

PR #1330, opened 2026-07-23
III-7339 migrate Text Area component to shadecn

danmkt00:
### Added
- shadcn `Textarea` behind `SHADCN_MIGRATION` flag
- `aria-label`, `placeholder`, `required` props on `TextArea`

### Changed
- `TextArea` no longer accepts `BoxProps`

### Fixed
- `TextArea` had no accessible name outside `FormElement`

---

Ticket: https://jira.uitdatabank.be/browse/III-7339


danmkt00:
910005233 baseline updates

brampauwelyn (commented):
Looks good to me.
Claude has 1 remark:

> One issue found: TextArea.tsx no longer accepts/forwards a maxLength prop. The old version spread arbitrary Box props (including maxLength) through to the native textarea; the new one only accepts a fixed, explicit prop list. FormElement (used to wrap form inputs) always clones its Component with a maxLength prop (FormElement.tsx:66-68), a pattern already used for Input in several places (OrganizerAddModal.tsx, PlaceAddModal.tsx, NameStep.tsx, labelForm.tsx). If TextArea is ever wired up the same way, the value would be silently dropped — no native maxlength HTML attribute would be set, so the browser stops enforcing the limit (only FormElement's visual counter would still work, since that's driven by value.length, not the prop).

brampauwelyn (approved):

PR #1331, opened 2026-07-23
III-7340 migrate TypeAhead Input component to shadecn

danmkt00:
# Tested in FAQ (Additional Information step) — the only current consumer of SuggestionInput

## Legacy
Screenshot 2026-07-23 at 13 00 53

## Shadecn
Screenshot 2026-07-23 at 13 00 46

### Added
- New shadcn-based `SuggestionInput` component (Popover + Command), behind the `SHADCN_MIGRATION` flag ([5dbf1fdd6](https://github.com/cultuurnet/udb3-frontend/commit/5dbf1fdd6)), built on new shadcn popover/command primitives ([e85a9b439](https://github.com/cultuurnet/udb3-frontend/commit/e85a9b439))
- Customizable `icon` prop on `SuggestionInput` (defaults to search icon) ([5dbf1fdd6](https://github.com/cultuurnet/udb3-frontend/commit/5dbf1fdd6))

### Changed
- Renamed `TypeaheadInput` to `SuggestionInput` — too easily confused with the existing, unrelated `Typeahead` component; despite similar names they share no implementation ([5dbf1fdd6](https://github.com/cultuurnet/udb3-frontend/commit/5dbf1fdd6))
- `FaqModal`'s question field now uses a chevron-down icon ([5dbf1fdd6](https://github.com/cultuurnet/udb3-frontend/commit/5dbf1fdd6))

### Removed
- Legacy `TypeaheadInput` implementation renamed to `TypeaheadInputLegacy` (kept, used when the flag is off) ([b90041d8c](https://github.com/cultuurnet/udb3-frontend/commit/b90041d8c))

### Fixed
- Invalid ARIA references (`aria-controls`/`aria-activedescendant`) when no suggestions matched ([5dbf1fdd6](https://github.com/cultuurnet/udb3-frontend/commit/5dbf1fdd6))

---

Ticket: https://jira.uitdatabank.be/browse/III-7340

PR #1332, opened 2026-07-23
Fix Tailwind v4 bracket syntax breaking tooltip transform-origin

danmkt00:
Discovered while migrating TypeaheadInput to shadcn that Tailwind v4's origin-[--var] bracket syntax silently compiles to an invalid transform-origin declaration instead of var(--var), and must be changed to origin-(--var)
Already fixed in [#1331](https://github.com/cultuurnet/udb3-frontend/pull/1331); split into this standalone PR along with CLAUDE.md instructions so it doesn't slip through again in future migrations.

### Added
- Documented Tailwind v4's bracket-vs-parens gotcha for bare CSS vars in CLAUDE.md, so future shadcn migrations catch it instead of shipping silently broken styles

### Fixed
- Tooltip's open/close zoom animation now scales from the correct anchor point instead of always from its own center (Tailwind v4's origin-[--var] bracket syntax silently compiled to an invalid transform-origin declaration; fixed to origin-(--var)) — barely visible, but worth fixing since the previous rule was dead code
---

Ticket: no ticket


CLAUDE.md
```diff
@@ -132,6 +132,8 @@ npx shadcn@latest add <component>
 
 After adding, fix the import order warning by running `yarn lint` or manually reordering imports. The `tw:` prefix is applied automatically via `components.json`.
 
+**Tailwind v4 gotcha**: bare CSS var references must use parens, not brackets — `origin-(--radix-foo)` works, `origin-[--radix-foo]` silently compiles to an invalid declaration. Grep new/migrated components with `grep -n -- '-\[--' <file>` (doesn't apply to `[--foo:value]` or normal arbitrary values like `w-[200px]`).
```
brampauwelyn: I would say `parentheses` but apparently `parens` is also a thing 😅 
```suggestion
**Tailwind v4 gotcha**: bare CSS var references must use parentheses, not brackets — `origin-(--radix-foo)` works, `origin-[--radix-foo]` silently compiles to an invalid declaration. Grep new/migrated components with `grep -n -- '-\[--' <file>` (doesn't apply to `[--foo:value]` or normal arbitrary values like `w-[200px]`).
```

brampauwelyn (approved):

PR #1337, opened 2026-07-30
[Fix] III-7363 button spacing fix

danmkt00:
### Fixed
- Buttons placed next to each other (or next to a conditional message/element) had no visible spacing when the `SHADCN_MIGRATION` flag was off. Root cause: legacy spacing works by copying a margin prop onto each child, but `Button` was rewritten at some point and no longer reads that prop — so the copy silently did nothing. Fixed by replacing those spots with real CSS spacing that works regardless of the flag.
  - Event creation wizard, type/theme step (https://github.com/cultuurnet/udb3-frontend/commit/eacc5494d0547498d6d2ce4ced0b4c6800165d13)
  - Event creation wizard, age step (https://github.com/cultuurnet/udb3-frontend/commit/b93c9bfac485ba36be55c949a49bc35829c21af9)
  - Edit-page footer publish/save buttons (https://github.com/cultuurnet/udb3-frontend/commit/effa2fe05678d056385ba2284b819e8ac0bead02)
  - Sidebar nav links not highlighting full-width on hover/select, and the "Beheer" section label rendering oversized/bold — same "legacy does something fragile the new look doesn't need" pattern (https://github.com/cultuurnet/udb3-frontend/commit/f1223dc80ad75d161a86288ba158ce9649d8b704)
  - Save/cancel, approve/reject, edit/delete, and similar action-button pairs across manage/organizer/production screens (https://github.com/cultuurnet/udb3-frontend/commit/59358729258128491816704b521da3a62c17d92c)
  - Buttons followed by a conditional message (error/max-reached text) or another element, where the gap only disappeared once that message rendered (https://github.com/cultuurnet/udb3-frontend/commit/12b366a44838c85c7c5e215655f6b9526fc23b6a)

**Note:** these fixes intentionally use Tailwind's standard gap scale (`tw:gap-2`, `tw:gap-3`, ...) instead of `getGapClass()`/the exact legacy pixel value, even where that's a few percent off. We'd rather have small, visible, easy-to-spot-and-fix diffs on the standard scale than keep leaning on exact legacy-parity values (or, worse, an unrelated global stylesheet's default like Bootstrap's reboot) to hold things together.

---

Ticket: https://jira.uitdatabank.be/browse/III-7363


danmkt00:
@claude review

danmkt00:
@claude review again please

danmkt00:
@claude final review

